PR #37, opened 2025-09-08
fix: requirements.txt for pytest

rudi-cilibrasi:


gltanaka:
@rudi-cilibrasi I forgot that I put the dev requirements in the toml file that can be installed like this: pip install -e ".[dev]"

PR #39, opened 2025-09-09
docs: add API key + CSV setup instructions for first-time users

sai4464:


examples/hello/.pdd/locks/hello_python.lock
gltanaka, line 1: don't know if you need to check this in

gltanaka (commented):
what is the point of the post_setup dir?

examples/hello/.pdd/locks/hello_python.lock
sai4464: Addressed the feedback regarding post_setup and the .lock file. Let me know if anything else needs adjustment.

Issue #57, opened 2025-09-20
Setup: Auto-seed llm_model.csv on first run (opt-out via PDD_AUTO_SEED=0)


gltanaka:
Summary
- Reduce first-run friction by ensuring a usable model catalog file exists without manual steps.

When It Triggers
- On first use of LLM-backed commands when neither `~/.pdd/llm_model.csv` nor project `.pdd/llm_model.csv` exists.

Behavior
- Create `~/.pdd/llm_model.csv` by copying the packaged default CSV.
- Optional: if API keys are detected, allow filtering to those providers when `PDD_AUTO_SEED_FILTER=1`.
- Never overwrite an existing file.

User Messaging
- Print once (respect `--quiet`): "Created `~/.pdd/llm_model.csv` from defaults. Edit to customize. Set `PDD_AUTO_SEED=0` to disable." If filtered, note that providers were filtered.

Opt-Out / Config
- `PDD_AUTO_SEED=0`: disable auto-seed and keep current fallback to packaged default only.
- `PDD_AUTO_SEED_FILTER=1`: filter to configured providers; otherwise copy as-is.

Error Handling
- If writing fails, continue using packaged default and log a warning (suppressed by `--quiet`).

Security & Privacy
- Do not write secrets; only provider names, model IDs, and env var names.
- Create `~/.pdd/` if missing; do not modify other files.

Testing
- No CSV present: creates file and prints message.
- Env vars present: filtering respected when enabled.
- `PDD_AUTO_SEED=0`: no file created; packaged default still used.
- `--quiet`: no message printed.
- Existing CSV: untouched.

Docs & Comms
- README Getting Started: note auto-seed and `PDD_AUTO_SEED=0` opt-out.
- Changelog entry with credit.

Credit: Rudi Cilibrasi

rudi-cilibrasi:
I would suggest a slightly different method of auto-configuration. On installation, don't do anything.
Wait until the user runs any pdd command except help or setup. If they run help, act as normal. If they run any other non-setup command check if there is already an llm-models.csv. If there is, run the command as normal. If the file does not exist, tell the user to run pdd setup and print the pdd setup command on a line by itself so it's easy to cut and paste.

The pdd setup command should list each API key variable it is probing on a separate line with a space, a colon, and FOUND or UNSET.

For each UNSET environment variable a link should be printed directing the user to the website where they can get it and a space and the word "FREE" or "PAID" depending on if there is a free way to obtain the api key.

[...]

PR #62, opened 2025-09-22
PR for #57: Add llm.csv auto generation

jiaminc-cmu:
Hi, I’ve been working on the first-run CSV generation feature for Issue #57 .

Right now the initial update in cli.py works like this:

pdd config DIRECTORY

Check if an API key is available in the environment.

If yes, ask the user which one to use.

If not, prompt the user to input API keys or quit.

Finally, generate a configuration file under DIRECTORY/.pdd/.

Next step: I’ll update the logic to align with the setup command.


rudi-cilibrasi:
I think we should avoid prompting the user interactively. Instead we should print a message non-interactively giving the user the hint for the next step.

If no API KEYS are discovered we should print a message asking the user to set one of the environment variables that are supported.
If the discovery process has good logging that will help the user understand their options.

PR #76, opened 2025-09-29
CLI: Add generate-tests and generate-examples aliases

keerthivasan2002:
Fix CLI: add generate-tests and generate-examples aliases

Issue

Users currently have to type `pdd test` or `pdd example` explicitly.  
It would be more natural to allow `pdd generate-tests ...` and `pdd generate-examples ...` as shorthand.

Solution

Added convenience aliases to support the following:

- `pdd generate-tests ...` → maps to `pdd test ...`
- `pdd generate-examples ...` → maps to `pdd example ...`

This improves ergonomics without breaking existing commands.

Key Files Changed

Core Code:
- `pdd/cli.py` — added alias registration with `cli.add_command`.

Documentation:
- `README.md` — updated Command Overview section to list the new aliases.

### DIFF
--- a/pdd/cli.py
+++ b/pdd/cli.py
@@ -1572,8 +1572,8 @@ def setup_cmd(ctx: click.Context) -> None:
 #   pdd generate tests ...
 #   pdd generate examples ...
 
-cli.add_command(test, "generate-tests")
-cli.add_command(example, "generate-examples")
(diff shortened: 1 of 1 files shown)


keerthivasan2002:
I went with hyphens since many CLI commands use them for multi-word subcommands, but I’m open to switching to spaces if that improves usability.

PR #119, opened 2025-11-03
Resolves #97:  Include in tests into the context window when generating

### LINKED ISSUE #97: include in tests into the context window when generating
when there a bug -> tests are appended

current state:
generate without tests in context, then we have to fix

Desired state:
generate with test on the context window so that we might not have to pdd fix

allow this to be option

Update Readme

rakshanda-cmu:
### Feat: auto detect and include unit test files during pdd generate 

This pull request introduces a new feature to pdd generate, which auto detects and includes unit test files to for code generation. This helps generate code that passes unit tests without the user manually having to include the test files.

Changes Implemented
New --exclude-tests Flag:

A --exclude-tests flag has been added to the pdd generate command in pdd/cli.py.
When this flag is used, the auto-include unit test feature is disabled.

Integrated Repository Logic in pdd/code_generate_main.py:

The core logic is now consolidated within code_generate_main.py:
- If --exclude-tests flag is set then does nothing (behaviour unmodified)
- The function auto detects unit test case files in the same directory as the .prompt file.   --exclude-tests flag and triggers the new workflow.
- If unit test files are found then a include test file statement is added to the .prompt file with the additional prompt instruction to generate code that passes these unit test
- The new .prompt file is save with prefix "ut_"

## How to Test
Navigate to any Git repository within your project with unit tests.
Run the command:
`pdd generate`

Observe the output:
The tool will first scan for unit test files and update .prompt file to include the unit test file and add prompt instructions to generate code that passes these tests. The updated .prompt file is saved as "ut_<original_filename>.prompt"

To disable the feature:
`pdd --exclude-tests generate`

The method now does not auto include test files during generation and runs as usuall

### DIFF
--- a/pdd/code_generator_main.py
+++ b/pdd/code_generator_main.py
@@ -305,7 +305,9 @@ def code_generator_main(
             test_filename = f"test_{basename}.py"
 
         # Create a semantic grouping tag name derived from the test filename
-        # Example: test_calculator.py -> tag 'test_calculator'
+        # and make the <include> reference a path that the preprocessor can
+        # resolve relative to the prompt file. Example: test_calculator.py ->
+        # tag 'test_calculator' and include 'tests/test_calculator.py'.
         try:
             raw_tag = pathlib.Path(test_filename).stem
             # Sanitize to XML-like tag name: allow letters, digits and underscore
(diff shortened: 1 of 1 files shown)


gltanaka:
0. Carefully follow Contributing guide
1. make sure all unit tests pass
2. append new tests to tests/test_*

target 11/5

PR #128, opened 2025-11-13
Resolves #109 Add CLI core-dump support for reproducible bug reports

### LINKED ISSUE #109: Telemetry and feedback hooks
- Problem: Hard to pinpoint user pain points.

  - Proposed fix:

    - Opt-in telemetry for command outcomes, error types

    - pdd feedback command to submit issues

- Acceptance:

    - Dashboard shows update success/failure rates and top errors

jiaminc-cmu:
# Add core-dump pipeline and report command to `pdd` CLI

## Summary

This PR adds first-class support for collecting and reporting “core dumps” from failed `pdd` runs. It introduces:

- A global `--core-dump` flag on the CLI to enable structured crash capture.
- Wiring so commands executed with `--core-dump` feed failure context into the core-dump machinery instead of crashing the CLI.
- A new `pdd report-core` command to turn a saved core dump into a human-readable issue report.

The default behavior for users who do **not** use `--core-dump` is unchanged.

---

### New global flag

- `--core-dump` (top-level option)
  - When set, `ctx.obj["core_dump"]` is `True`.
  - If a subcommand raises, the CLI handles the error via the existing error handler and exits cleanly (no traceback propagated to Click), allowing the core-dump machinery to run instead of crashing the CLI.
  - When the flag is **not** set, behavior remains the same as today (friendly error message, no traceback).

### New command: `report-core`

- `pdd report-core [CORE_FILE]`
  - Invokes the core-dump reporting pipeline to generate a human-readable issue-style report.
  - `CORE_FILE` is an optional path to a core-dump artifact produced by a previous `pdd` run.
  - If `CORE_FILE` is omitted, the underlying `report_core_main` implementation can decide how to locate a default core dump (e.g., “latest” in a standard directory).
  - The report is intended to be copy-pasteable into a GitHub issue or internal bug tracker.

---

## Implementation details

- **Global context**
  - The top-level Click entrypoint now accepts a `--core-dump` option.
  - The value is stored in `ctx.obj["core_dump"]` and made available to subcommands and error-handling logic.

- **Error handling with core dumps**
  - Existing error-handling behavior is preserved for normal runs (no `--core-dump`).
  - When `--core-dump` is enabled:
    - Exceptions raised by a subcommand are handled by the CLI error handler.
    - The CLI does **not** propagate the exception to Click (no non-zero exit due to the Python exception itself).
    - This allows the core-dump pipeline to capture the failure context and write a structured artifact without crashing the CLI.

- **Report command**
  - `report-core` is implemented as a standard Click subcommand.
  - It delegates to `report_core_main`, passing through the optional `core_file` path.
  - The reporting pipeline is responsible for:
    - Reading the core-dump artifact.
    - Extracting metadata (command, options, exception info, model/cost if available, timestamps, etc.).
    - Emitting a formatted issue-style summary.

---

## Tests

New tests added to `tests/test_cli.py`:

- **Core-dump global option:**
  - `test_cli_core_dump_default_flag_false`
    - Verifies that, by default, `ctx.obj.get("core_dump", False)` is `False` when running a normal command.
  - `test_cli_core_dump_flag_sets_ctx_true`
    - Ensures that `--core-dump` sets `ctx.obj["core_dump"]` to `True` and that the flag is threaded into the subcommand call.
  - `test_cli_core_dump_does_not_propagate_exception`
    - When `--core-dump` is enabled and the underlying command raises:
      - The exception is handled by the CLI error handler.
      - No exception is propagated back to `CliRunner`.
      - Exit code remains `0`, so the core-dump machinery can run without the CLI itself failing.

(Additional behavior is covered by existing CLI tests; non-core-dump flows remain unchanged.)

---

## Notes

- Existing users who do not opt into `--core-dump` see no behavior changes.
- This PR lays the groundwork for richer debugging workflows, such as:
  - Running `pdd <command> --core-dump` to capture all crash context.
  - Using `pdd report-core` on that artifact to quickly prepare and submit a GitHub issue or internal bug report.

### DIFF
--- a/pdd/cli.py
+++ b/pdd/cli.py
@@ -2305,7 +2305,7 @@ def setup_cmd(ctx: click.Context) -> None:
         install_completion(quiet=quiet_mode)
         _run_setup_utility()
         if not quiet_mode:
-            console.print("[success]Setup completed. Restart your shell or source your RC file to apply changes.[\success]")
+            console.print("[success]Setup completed. Restart your shell or source your RC file to apply changes.[/success]")
     except Exception as exc:
         handle_error(exc, command_name, quiet_mode)
 


gltanaka:
@jiaminc-cmu This is a big step forward!  Thanks for your work on this.

I wonder if we should also do the following:
1) Give the user the option to autopost this issue to create a github issue on the pdd repo
2) Allow the users to auto upload the input and output files so we can reproduce the issue (we asked for permission)
3) Allow to enter a prompt to describe the issue so we can auto-write a good github issue
4) Have some sort of mechanism were we can auto-play the issue back

@rudi-cilibrasi what are your thoughts?

Issue #134, opened 2025-11-17
Error saving Gemini API key in setup


paul-alumni-caltech:
Error saving Gemini key in setup.

On running setup and pasting my Gemini key into the command line I got an error similar to  "could not parse characters in locations 12 to 90" ( I didn't record the exact language) .  After following the instructions for manual configuration, it worked fine.

I had no problem using setup with the Open AI API key.

Frostday:
Hi Paul,

I want to confirm: does your Gemini API key contain shell metacharacters (like quotes, dollar signs, or backslashes)? 

If so, the setup script writes the key directly into a shell script file without escaping, which causes the shell to fail parsing with "could not parse characters" at those positions. This would explain why OpenAI keys (which typically don't have these characters) work, while Gemini keys fail. Can you confirm if your key has any special characters? If so, I'll fix it by properly escaping the key values before writing them to the script files.

PR #206, opened 2025-12-18
Fix for issue #134

Frostday:
Working on a fix for issue - https://github.com/promptdriven/pdd/issues/134

I believe the error occurs because API key values aren't properly escaped when written to shell script files. Gemini keys likely contain special characters (quotes, backslashes, dollar signs) that break shell parsing, causing the "could not parse characters" error.


Frostday:
Made the bug fix using the pdd fix command

Issue #403, opened 2026-01-26
Bug: File Handle Resource Leak in SyncLock.acquire()


Serhan-Asad:
## Bug Description
**File**: `pdd/sync_determine_operation.py`  
**Line**: 183  
**Type**: Resource Leak

The `SyncLock.acquire()` method opens a file without guaranteeing it will be closed if certain exceptions occur, leading to file descriptor leaks in long-running processes.

## Location

```python
# pdd/sync_determine_operation.py:183-200

def acquire(self):
    try:
        # ... setup code ...
        
        # Line 183: File opened, FD assigned
        self.fd = open(self.lock_file, 'w')  # ← PROBLEM: No guaranteed cleanup
        
        # Lines 185-194: Operations that can raise various exceptions
        if HAS_FCNTL:
            fcntl.flock(self.fd.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
        elif HAS_MSVCRT:
            msvcrt.locking(self.fd.fileno(), msvcrt.LK_NBLCK, 1)
        
        self.fd.write(str(self.current_pid))
        self.fd.flush()
        
    except (IOError, OSError) as e:  # ← Only catches IOError and OSError!
        if self.fd:
            self.fd.close()
            self.fd = None
        raise TimeoutError(f"Failed to acquire lock: {e}")
```

## Root Cause

The file is opened at line 183, but the exception handler only catches `(IOError, OSError)`. If any other exception occurs (e.g., `KeyboardInterrupt`, `RuntimeError`, `MemoryError`), the file descriptor is never closed.

### When Does This Leak Occur?

1. **User presses Ctrl+C** during lock acquisition
2. **Unexpected exceptions** (not IOError/OSError) during fcntl.flock()
3. **System interrupts** during the lock acquisition process
4. **Out of memory** or other system errors

### Consequences

- **File descriptors accumulate** over time in long-running processes
- **Eventually causes**: `OSError: [Errno 24] Too many open files`
- **Process crashes** requiring manual restart
- **Affects**:
  - PDD server running continuously
  - Automated scripts with repeated sync operations
  - CI/CD pipelines
  - Users frequently interrupting operations

### Automated Test Results

Created isolated test in `bug_verification_workspace/simple_leak_test.py`:

```
$ python3 simple_leak_test.py

======================================================================
SIMPLE FILE HANDLE LEAK TEST
======================================================================

Process PID: 9908

1. Initial state (no locks):
   Open .lock files: 0

[...]

PR #410, opened 2026-01-27
Revert "Add failing tests for #403: File Handle Resource Leak in SyncLock.acquire()"

gltanaka:
Reverts promptdriven/pdd#405

This PR made pdd sync super slow


gltanaka:
Closing as superseded. This revert was opened to remove the slow #403 tests, but current upstream main has already removed tests/test_e2e_issue_403_file_handle_leak.py while retaining the current sync_determine_operation test suite. Applying this PR now would delete hundreds of lines from current tests/code, so it no longer serves its original purpose.

Issue #375, opened 2026-01-22
Race Condition in LLM Cost Tracking Causes Data Corruption


Serhan-Asad:
## Summary

The global `_LAST_CALLBACK_DATA` dictionary in `pdd/llm_invoke.py` is shared across concurrent jobs in server mode, causing cost data to be corrupted when multiple jobs run simultaneously. This breaks budget enforcement and cost reporting features.

## Affected Components

- `pdd/llm_invoke.py:678-684` - Global `_LAST_CALLBACK_DATA` dictionary
- `pdd/llm_invoke.py:686-759` - `_litellm_success_callback()` function
- `pdd/sync_orchestration.py:1020-1026` - Budget enforcement logic
- `pdd/sync_orchestration.py:1532,1555` - Cost accumulation and reporting
- `pdd/server/app.py:54` - JobManager with `max_concurrent=3`

## Environment

- **PDD Version:** v0.0.122
- **Operating System:** macOS (tested), likely affects all platforms

### Root Cause

The LiteLLM callback stores cost/token data in a module-level global dictionary:

```python
# pdd/llm_invoke.py:678-684
_LAST_CALLBACK_DATA = {
    "input_tokens": 0,
    "output_tokens": 0,
    "finish_reason": None,
    "cost": 0.0,
}
```

When the PDD server runs multiple jobs concurrently (up to 3 by design), all threads share this same dictionary:

1. **Thread 1** writes its cost ($0.10) to `_LAST_CALLBACK_DATA`
2. **Thread 2** immediately overwrites with its cost ($0.30)
3. **Thread 3** overwrites again with its cost ($0.05)
4. **Thread 1** reads back the cost → gets $0.05 (Thread 3's value!) ❌
5. **Thread 2** reads back the cost → gets $0.05 (Thread 3's value!) ❌

The code even has a warning comment acknowledging this issue:
```python
# Line 678: Module-level storage for last callback data
# (Use with caution in concurrent environments)
```

But no synchronization mechanism is implemented.

### When the Bug Occurs

The bug **only** affects **server mode** (`pdd connect`), not CLI mode:

- ✅ **Safe:** Running `pdd sync` in separate terminals (separate processes)
- ❌ **Affected:** Using `pdd connect` web UI with multiple operations
- ❌ **Affected:** Multiple users on a shared PDD server
- ❌ **Affected:** Automation/CI scripts submitting concurrent jobs via API

#### 1. Budget Limits Don't Work

```python
[...]

PR #478, opened 2026-02-07
fix: thread-safe cost extraction from response objects (#375)

Serhan-Asad:
## Summary

- **Fixes race condition** in concurrent LLM cost tracking (issue #375) by reading cost directly from each response object instead of the shared `_LAST_CALLBACK_DATA` global dict
- **Extracts `_get_response_cost()`** — reads `response._hidden_params['response_cost']` (pre-computed by LiteLLM), falls back to `_calculate_completion_cost()` (public API)
- **Extracts `_calculate_completion_cost()`** — 3-tier fallback: `litellm.completion_cost(response)` → `litellm.completion_cost(model, tokens)` → manual CSV rates
- **Adds 7 new tests** covering thread safety (10 concurrent threads), batch mode, empty response list, exception handling, and `_hidden_params` fallback behavior

## Test plan

- [x] All 92 existing + new tests pass (`pytest tests/test_llm_invoke.py`)
- [x] Thread safety test: 10 concurrent threads each get correct per-response cost (no cross-thread contamination)
- [x] Validated with real `pdd sync` runs (5 rounds, sequential + 3 parallel syncs) — all report correct non-zero costs
- [x] Backward compatible: `_LAST_CALLBACK_DATA` callback still populated for single-threaded CLI usage

### DIFF
--- a/pdd/llm_invoke.py
+++ b/pdd/llm_invoke.py
@@ -2808,8 +2808,8 @@ def llm_invoke(
                 if verbose:
                     # Extract tokens directly from last response (thread-safe)
                     last_usage = getattr(response_list[-1], 'usage', None) if response_list else None
-                    input_tokens = getattr(last_usage, 'prompt_tokens', 0) or 0
-                    output_tokens = getattr(last_usage, 'completion_tokens', 0) or 0
+                    input_tokens = getattr(last_usage, 'prompt_tokens', None) or getattr(last_usage, 'input_tokens', 0) or 0
+                    output_tokens = getattr(last_usage, 'completion_tokens', None) or getattr(last_usage, 'output_tokens', 0) or 0
 
                     cost_input_pm = model_info.get('input', 0.0) if pd.notna(model_info.get('input')) else 0.0
                     cost_output_pm = model_info.get('output', 0.0) if pd.notna(model_info.get('output')) else 0.0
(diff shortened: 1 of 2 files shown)


gltanaka:
Hey @Serhan-Asad, thanks for the detailed PR and for taking the time to dig into the cost tracking code! The analysis of `_LAST_CALLBACK_DATA` being unsafe for concurrent use is spot-on in principle — that global dict pattern is definitely a code smell worth flagging.

After tracing the execution path, I think the race condition can't actually manifest in the current architecture though. The server (`pdd connect`) runs each job as a **separate subprocess** via `subprocess.Popen` (`server/jobs.py:462`), so each job gets its own Python process, its own memory space, and its own `_LAST_CALLBACK_DATA`. The `ThreadPoolExecutor` in `JobManager` only manages the subprocess lifecycle (reading stdout/stderr), not in-process LLM calls.

Here's the evidence chain:
- `server/app.py:54` → `JobManager(max_concurrent=3)` (no custom executor)
- `server/jobs.py:392-395` → always takes `_run_click_command()` path
- `server/jobs.py:462` → `subprocess.Popen(cmd_args, ...)` — fully isolated process

So there's no shared state between concurrent jobs in practice.

That said, the code quality improvements here are real:
- Reading cost from the response object rather than a global is cleaner design
- The `_calculate_completion_cost()` extraction is a nice refactor
- The batch mode cost summing is more correct than reading the last callback value (though batch mode isn't used in production today)

I'm going to close this one since the core issue doesn't apply, but I appreciate you looking into this area of the codebase — and the thoroughness of the tests you wrote. If you're interested in contributing, I'd love to have you look at some of the open issues. Keep it up! 🙏

Issue #480, opened 2026-02-07
pdd setup: Expand to all LLM providers and improve API key management


niti-go:
This issue consolidates `pdd setup` improvements from Issue #100  and PR #123  plus additional related improvements.

`pdd setup` is currently limited to only 3 LLM providers (OpenAI, Google, Anthropic) and uses hardcoded API testing. This prevents users with other providers (Fireworks, Groq, Vertex AI) from using their API keys even if they have them configured. Also, there's a mismatch between how setup discovers API keys and how they're loaded at runtime.

### Improvements

**1. Use llm_invoke for API Key Testing**
Currently, setup uses hardcoded HTTP tests for only 3 providers. This prevents testing providers with complex authentication like Vertex AI. The solution is to use the existing `llm_invoke()` function, which handles all LiteLLM providers automatically.

**2. Support All LLM Providers Dynamically**
Setup filters models to only include `gpt-*`, `gemini/*`, and `anthropic/*` models, excluding Fireworks, Groq, Vertex AI, and local models even when API keys are available. Instead, read all providers from `llm_model.csv` and include any model where the API key exists.

**3. API Key Management**
Users should be able to delete API keys through 'pdd setup' and see where they're loaded from (shell env vars vs `.env` file). Currently there's no delete option and no transparency about key sources.

**4. Align API Key Loading Hierarchy**
Runtime (`llm_invoke.py`) loads keys from `.env` first, then shell vars. Setup only checks shell vars and ignores `.env`. This can cause users to set keys in `.env`, have `pdd setup` ignore them, and end up with duplicates in multiple places.

**5. Model Tier Guidance**
Setup loads all available models (Opus, Sonnet, Haiku) into `llm_model.csv` without explanation. Users don't understand which model will be used or the 10x cost difference between tiers. We should add interactive model selection during setup.

niti-go:
## Step 4: Requirements Clarity Check

**Status:** Clarification Needed

### Clarity Assessment
| Aspect | Status |
|--------|--------|
| Problem statement | ✅ Clear |
| Expected outcomes | ⚠️ Partially defined |
| Scope | ✅ Bounded |
| Edge cases | ❌ Not addressed |

### What's Clear

The issue clearly identifies 5 improvements to `pdd setup`:
1. Use `llm_invoke()` for API testing instead of hardcoded HTTP tests
2. Support all LLM providers dynamically from `llm_model.csv`
3. Add API key deletion and source transparency
[...]

PR #544, opened 2026-02-20
feat: add comprehensive pdd setup prompts for all LLM providers (#480)

niti-go:
## Restructure `pdd setup` into an auto-configuring flow with expanded provider support and functionality. ##

1. Auto-configures Gemini, Claude, and Codex (required for agentic tools)
2. Adds support for many more LiteLLM-supported providers, even those with unique authorization flows (Vertex AI, AWS Bedrock, Azure, etc.)
3. Model testing uses real LLM calls so errors surface during setup, not at runtime
4. `.pddrc` is automatically configured based on detected project language

**Important Notes** 

- **I could only test Gemini models (I had an API key). I recommend more testing for providers with multi-credential or complex auth, especially GitHub Copilot, Vertex AI, and AWS Bedrock.**
- **This completely rewrites data/llm_model.csv to support more providers. It also has the script** `generate_model_catalog.py`**that generates this file, but PDD maintainers need to re-run this script whenever LiteLLM's registry updates, and periodically update the hardcoded ELO scores in the file. Please take a look if the csv has the desired models, and if we want to set up a different way to update the csv file.**

**Design choices**

- Deterministic over agentic. After initial CLI setup, the rest of pdd setup could theoretically be agentic, but in practice this was slower and didn’t offer any improvements that deterministic python code couldn’t do, where it just needs to scan env vars, create files, and save api keys, etc.
- ELO threshold of 1400. `data/llm_model.csv` only includes models with ELO scores above 1400, which are really the only ones useful for coding. Local LLMs (Ollama or LMStudio) are not included because they don’t score that high.
- Users have to press "enter" multiple times to progress through pdd setup. I want users to see what's being configured and where files live. This keeps things interpretable for users who want to customize later.

### **New Dev Units** ###

Each of these comes with examples and tests (and prompt files in `pdd_cap`).

`setup_tool.py`

- The main orchestrator. First, it bootstraps agentic CLIs via `cli_detector`. Then, it auto-discovers API keys from env/dotenv/api-env files, matches them to models from the reference CSV, writes the user's `~/.pdd/llm_model.csv`, optionally creates `.pddrc`, and smoke-tests the first available model.
- Prints a structured summary (CLIs, keys, models, test result), writes `PDD-SETUP-SUMMARY.txt` and creates a sample `success_python.prompt` for immediate first use. Also offers an optional menu to add providers or test models.

`cli_detector.py`

- Detects Claude, Codex, and Gemini CLIs via PATH/fallback lookup. Presents a numbered selection table with install/key status, and walks through installation (npm) and API key configuration (saved to `~/.pdd/api-env.{shell}`) for each selected CLI.

`api_key_scanner.py`

- Reads the local model catalog to determines which API keys are needed (including multi-key providers like AWS Bedrock), and checks `.env`, shell environment, and PDD config files. Reports what's found and what's missing without exposing key values.

`provider_manager.py`

- Add and remove LLM providers and models from the local PDD configuration. Supports multi-credential providers like Vertex AI and AWS Bedrock (referenced LiteLLM documentation for the credentials needed). Saves API keys to a shell-sourced env file so they persist across terminal sessions.

`model_tester.py`

- Lets users pick a configured model, checks that required credentials are set, then sends a real test call. Shows cost, speed, and a clear error message if something fails. Results persist within a session so users can test multiple models and compare.

`pddrc_initializer.py`

- Detects project language from marker files and generates a `.pddrc` with sensible defaults. Asks for confirmation before writing; skips if one already exists.

**New Script (not a full dev unit)**

`generate_model_catalog.py`

- Script that generates `data/llm_model.csv` by looking up all models/providers from LiteLLM and filtering to ELO scores above 1400. **PDD maintainers should re-run this script whenever LiteLLM's registry updates, and periodically update the hardcoded ELO scores in the file.**
- I decided not to make this a full pdd dev unit because it’s short and relies on a lot of hardcoded values.

### **Updated files** ###

`data/llm_model.csv`

- Now includes all top models with ELO ≥ 1400 across all LiteLLM-supported providers.
- The `api_key` column now supports pipe-delimited fields (e.g. `VERTEXAI_PROJECT|VERTEXAI_LOCATION|GOOGLE_APPLICATION_CREDENTIALS`) for providers whose auth requires multiple credentials.

`llm_invoke.py`

The `api_key` handling was rewritten to support the new pipe-delimited credentials format and generalized to remove provider-specific logic.

- `*_ensure_api_key` (pre-call credential check):*
    - **Single var** (e.g. `OPENAI_API_KEY`) — same as before: check env, prompt if missing, save to `.env`.
    - **Multiple vars** (e.g. `VAR1|VAR2|VAR3`) — checks all are set. If any are missing, directs the user to `pdd setup` instead of prompting interactively. Includes an ADC fallback for Vertex AI when `GOOGLE_APPLICATION_CREDENTIALS` isn't set but `gcloud auth` was used.
    - **Empty** — no key needed (device flow or local model).
- *LiteLLM call setup (used to have ~80 lines of Vertex-specific logic, now ~20 lines of generic logic):*
    - **If the model has a single var** — read env var, pass as `api_key=`.
    - **If the model has multiple vars** — let LiteLLM read them from the environment directly.
    - **If the model has no api_key value** — pass nothing (authentication may be through device flow, e.g. GitHub Copilot).
- Retry kwargs were also simplified — removed `vertex_credentials`, `vertex_project`, `vertex_location` since those are now handled through environment variables.

 `test_llm_invoke.py` 

- Updated to support new model api_key column. The old tests were setting `VERTEX_CREDENTIALS` as the mock env var, but the new CSV `api_key` column for Vertex AI models specifies `GOOGLE_APPLICATION_CREDENTIALS|VERTEXAI_PROJECT|VERTEXAI_LOCATION`. The `_ensure_api_key` function checked those exact env vars (because it previosuly only needed to look for one var), finds them missing , returns `False`, and skips the model — causing "All candidate models failed."
- **Fix**: Changed the tests to set the correct env vars (`GOOGLE_APPLICATION_CREDENTIALS`, `VERTEXAI_PROJECT`, `VERTEXAI_LOCATION`) that match what the CSV now declares.

 `test_llm_invoke_vertex_retry.py` 

- Updated to support new model api_key column. The old tests verified that `llm_invoke` **explicitly passed** Vertex AI credentials on retry. The new tests verify that `llm_invoke` **does NOT** pass them, because under the multi-credential convention, LiteLLM reads from `os.environ` directly — both on initial calls and retries.

These tests all pass now.

`README.md` and `SETUP_WITH_GEMINI.md`

- Updated `pdd setup` documentation

### DIFF
--- a/pdd/api_key_scanner.py
+++ b/pdd/api_key_scanner.py
@@ -76,13 +76,15 @@ def _load_dotenv_values() -> Dict[str, str]:
     """
     Load values from a .env file using python-dotenv's dotenv_values (read-only).
     Returns an empty dict on any failure.
+    Only includes keys with non-empty values.
     """
     try:
         from dotenv import dotenv_values  # type: ignore
 
         values = dotenv_values()
         # dotenv_values returns an OrderedDict; values can be None for keys without values
-        return {k: v for k, v in values.items() if v is not None}
+        # Filter out None and empty string values
+        return {k: v for k, v in values.items() if v and v.strip()}
     except ImportError:
         logger.debug("python-dotenv not installed; skipping .env file check.")
         return {}
@@ -105,7 +107,7 @@ def _detect_shell() -> Optional[str]:
 def _parse_api_env_file(file_path: Path) -> Dict[str, str]:
     """
     Parse a ~/.pdd/api-env.{shell} file for uncommented `export KEY=value` lines.
-    Returns a dict of key names to values found.
+    Returns a dict of key names to non-empty values found.
     """
     result: Dict[str, str] = {}
 
@@ -137,7 +139,8 @@ def _parse_api_env_file(file_path: Path) -> Dict[str, str]:
                         ):
                             value = value[1:-1]
 
-                        if key and value:
+                        # Only include keys with non-empty values
+                        if key and value and value.strip():
                             result[key] = value
 
     except Exception as e:
@@ -183,17 +186,17 @@ def scan_environment() -> Dict[str, KeyInfo]:
             api_env_source_label = f"~/.pdd/api-env.{shell_name}"
 
         for key_name in key_names:
-            # Check in priority order
+            # Check in priority order, ensuring values are non-empty
             if key_name in dotenv_vals:
                 result[key_name] = KeyInfo(source=".env file", is_set=True)
-            elif key_name in os.environ:
+            elif key_name in os.environ and os.environ[key_name].strip():
                 result[key_name] = KeyInfo(source="shell environment", is_set=True)
             elif key_name in api_env_vals:
                 result[key_name] = KeyInfo(
                     source=api_env_source_label, is_set=True
                 )
             else:
-                # Key not found in any source
+                # Key not found in any source or has empty value
                 result[key_name] = KeyInfo(source="", is_set=False)
 
     except Exception as e:
--- a/tests/test_api_key_scanner.py
+++ b/tests/test_api_key_scanner.py
@@ -46,6 +46,9 @@
 #
 # VIII. scan_environment — Edge Cases
 #   26. test_scan_special_chars_in_key_value: Key value with special chars → no crash.
+#   27. test_scan_empty_string_values_not_set: Keys with empty values → is_set=False.
+#   28. test_scan_empty_dotenv_falls_through_to_shell: Empty .env value → shell environment wins.
+#   29. test_scan_empty_shell_falls_through_to_dotenv: Empty shell value → .env file wins.
 
 import csv
 from pathlib import Path
@@ -327,7 +330,8 @@ def test_scan_detects_shell_env_key(tmp_path, monkeypatch):
     monkeypatch.setenv("OPENAI_API_KEY", "sk-test123")
     monkeypatch.delenv("ANTHROPIC_API_KEY", raising=False)
 
-    result = scan_environment()
+    with mock.patch("pdd.api_key_scanner._load_dotenv_values", return_value={}):
+        result = scan_environment()
 
     assert result["OPENAI_API_KEY"].is_set is True
     assert result["OPENAI_API_KEY"].source == "shell environment"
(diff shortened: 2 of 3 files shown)


gltanaka (changes requested):
## Review: Cherry-pick + isolation testing on `test/pr-544-isolated`

Hey @niti-go — really impressive work here\! I cherry-picked all 4 commits onto an isolated test branch and put them through a full manual + automated test suite. Here's what I found:

---

### What's working great

The core of this PR is solid. Every critical path I tested passed:

- **All 5 new modules import cleanly** — `provider_manager`, `api_key_scanner`, `model_tester`, `cli_detector`, and `pddrc_initializer` all load without errors.
- **Pipe-delimited API key format works end-to-end** — `parse_api_key_vars("AWS_ACCESS_KEY_ID|AWS_SECRET_ACCESS_KEY|AWS_REGION_NAME")` correctly returns a list, and `_ensure_api_key()` correctly returns `True` when all vars are set and `False` when any one is missing. This is the most important new behavior and it's correct.
- **Backward compatibility is preserved** — single-credential providers like Anthropic and OpenAI continue to work exactly as before.
- **Vertex AI retry tests: 4/4 pass** — the new env-var-based credential convention doesn't break the retry logic.
- **217/218 tests in `test_llm_invoke.py` pass** — the LLM calling path is in great shape.
- **The curated 92-model CSV loads correctly** and the `avg_cost` column is properly parsed.

The architectural idea here — generalizing credential handling to support multi-var providers like AWS Bedrock and Vertex AI using a pipe-delimited format — is elegant and the right call. Well done.

---

### Two things to fix before merging

**1. `api_key_scanner` source attribution bug** (`tests/test_api_key_scanner.py`, 2 failures)

`scan_environment()` always reports `".env file"` as the source even when the key was set via the shell environment or `~/.pdd/api-env.bash`. The two failing tests catch this precisely:

- `test_scan_detects_shell_env_key` — expects `source == "shell environment"`, gets `".env file"`
- `test_scan_detects_api_env_file_key` — expects `source == "~/.pdd/api-env.bash"`, gets `".env file"`

The priority-order check in `scan_environment()` (`.env` → shell → api-env file) seems to be matching `.env` first even when the key isn't actually in `.env`. Worth double-checking `_load_dotenv_values()` — if the test fixture's home directory happens to have a `.env` file that picks up the key, it would shadow the shell/api-env sources.

**2. CSV / test consistency** (`test_llm_invoke.py::test_deepseek_maas_passes_response_format_for_structured_output`, 1 failure)

The CSV curation commit (`78cdd40`) removed `vertex_ai/deepseek-ai/deepseek-v3.2-maas`, but this test checks for that exact model. Either add the model back to the CSV or update the test to reflect the curated set. Since the test was written to catch a specific `structured_output` bug for that model, it's worth keeping — just needs the CSV entry restored.

---

### Recommendation

Fix those two issues and this is ready to merge. The hard part (the multi-credential architecture) is already correct. These are small, targeted fixes.

Great contribution — this significantly improves the setup experience for multi-var providers and the Pareto-filtered model list is a nice touch\!

Issue #69, opened 2025-09-25
install via vscode extension


gltanaka:

qanagattandyr:
Hi, I will be taking this issue and I will finish it in a few days

PR #694, opened 2026-03-16
Fix/issue 691

### LINKED ISSUE #691: Optional context/example.prompt include injects "[File not found: ...]" into LLM prompt
## Problem

The example-generator template (pdd/prompts/example_generator_LLM.prompt) has:
<include>./context/example.prompt</include>
context/example.prompt is optional project context (README: “If found, the content … is included”; many projects don’t have it). When the file is missing, pdd/preprocess.py in process_include_tags() includes it in the LLM prompt. So the include is replaced by the literal string [File not found: ./context/example.prompt]. That preprocessed text is then used as the LLM prompt for example generation (pdd example and the example step of pdd sync). So the model receives an internal “file not found” message as part of the task prompt.

## Impact

- Wrong audience: The placeholder is for developer visibility, but it’s sent to the LLM as if it were an instruction.
- Wrong semantics: For an optional include, “missing” should mean “no extra content,” not a line about the filesystem.
- Prompt pollution: The prompt should only contain task instructions; this can confuse the model or leak into output (e.g. comments about the missing file).

## Suggested Fix
When an included file is not found and we’re building a prompt that goes to the LLM, do not inject [File not found: ...] into that prompt. For optional/conventional project includes (e.g. context/example.prompt, context/test.prompt), replace the include with empty string when the file is missing. Keep the existing console warning (Warning: File not found: ...) so developers still see the issue. 

Options:
In preprocess.py: For FileNotFoundError when recursive=False, return "" instead of f"[File not found: {file_path}]" (may affect other callers that rely on the placeholder).
Convention-based: In process_include_tags(), treat paths like ./context/example.prompt and ./context/test.prompt as optional: on FileNotFoundError, return ""; for other paths, keep current placeholder behavior.
New tag: Add something like <include-optional>...</include-optional> that expands to empty when the file is missing, and use it in example_generator_LLM.prompt (and similar templates) for optional project context.

vishalramvelu:
## Summary

Treats context/example.prompt and context/test.prompt as truly optional project context for LLM prompts. When these conventional files are missing, <include>./context/example.prompt</include> (and ./context/test.prompt if used similarly) now expands to an empty string instead of [File not found: …], while still logging a console warning. This prevents internal “file not found” markers from being sent to the LLM as instructions or leaking into generated examples/tests, while keeping developers informed via the console.

## Test Results

- Unit tests:
PASS — tests/test_preprocess.py::test_file_not_found
PASS — tests/test_preprocess.py::test_optional_context_example_missing_is_silent
- Preprocess suite:
PASS — uv run pytest tests/test_preprocess.py -v --tb=short

## Manual Testing

- Ran uv run pdd-cli example <prompt-file> in a repo without context/example.prompt:
Before fix: the preprocessed prompt included [File not found: ./context/example.prompt] where the include tag was.
After fix: the include expands to empty in the LLM-facing prompt, while the console still shows Warning: File not found: ./context/example.prompt.
- Verified that non-optional includes (e.g. <include>missing.txt</include>) still produce [File not found: missing.txt] in the prompt, preserving existing behavior for genuinely missing required files.

## Checklist

- [x] All relevant preprocess tests pass (tests/test_preprocess.py)
- [x] Confirmed optional context includes no longer inject [File not found: …] into LLM prompts
- [x] Kept console warnings for missing optional context files for developer visibility

Fixes #691

### DIFF
--- a/README.md
+++ b/README.md
@@ -1793,6 +1793,7 @@ PDD supports the following XML-like tags in prompt files. Note: XML-like tags (`
    - `select=` — deterministic structural extraction (functions, classes, line ranges, headings, regex, JSON/YAML paths). Composable via comma-separation.
    - `mode="interface"` — Python-only. Extracts signatures and docstrings with bodies replaced by `...`.
    - `query=` — LLM-powered semantic extraction, cached in `.pdd/extracts/`.
+   - `optional` — when present on an `<include ...>` tag, a missing file resolves to an empty string (`""`) during non-recursive preprocessing (while still logging a warning).
    - When both `select=` and `query=` are present, `select=` wins (no LLM cost).
 
    This mechanism is also used internally by some commands (like `test` and `example`) to automatically incorporate project-specific context files if they exist in conventional locations (e.g., `context/test.prompt`). See 'Providing Command-Specific Context' for details. For the full selector reference, see the [Prompting Guide](docs/prompting_guide.md#selective-includes).
--- a/pdd/docs/prompting_guide.md
+++ b/pdd/docs/prompting_guide.md
@@ -174,7 +174,7 @@ These patterns are used across prompts in this repo:
 
 - Preamble and role: start with a concise, authoritative description of the task and audience (e.g., “You are an expert Python engineer…”).
 - Includes for context: bring only what the model needs.
-  - Single include: `<include>path/to/file</include>`. **Note:** This is a PDD directive, not standard XML. The PDD tool replaces this tag with the actual file content *before* the LLM sees it. (Handles both text and images).
+  - Single include: `<include>path/to/file</include>`. **Note:** This is a PDD directive, not standard XML. The PDD tool replaces this tag with the actual file content *before* the LLM sees it. (Handles both text and images). Use `<include optional>path/to/file</include>` to treat missing files as empty string (while still logging a warning).
   - Multiple: `<include-many>path1, path2, …</include-many>`
   - Grouping: wrap includes in a semantic tag to name the dependency or file they represent, for example:
     ```xml
--- a/pdd/preprocess.py
+++ b/pdd/preprocess.py
@@ -242,6 +242,10 @@ def _parse_attrs(attr_str: str) -> dict:
     # Simple attribute parser: key="value" or key='value'
     for match in re.finditer(r'(\w+)\s*=\s*["\']([^"\']*)["\']', attr_str):
         attrs[match.group(1)] = match.group(2)
+    # Boolean attributes (e.g., <include optional>path</include>)
+    # This keeps the include tag syntax ergonomic without requiring key="true".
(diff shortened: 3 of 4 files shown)


gltanaka:
target 3/17

vishalramvelu:
> Thanks for tackling this — the bug is real and the fix direction is right. The `[File not found: ...]` marker leaking into LLM prompts is definitely something we need to address.
> 
> A few things I'd like to see before merging:
> 
> **1. Move optionality to the include site instead of a hardcoded path list**
> 
> The hardcoded check for `context/example.prompt` and `context/test.prompt` in `preprocess.py` means every future optional include requires a code change to the preprocessor. Consider adding an `optional` attribute to the `<include>` tag instead:
> 
> ```
> <include optional>./context/example.prompt</include>
> ```
> 
> Then the `replace_include` handler just checks for that attribute and returns `""` on `FileNotFoundError`. This way the prompt templates themselves declare optionality, no list to maintain, and any new optional include just works.
> 
> **2. Update the prompt files that use these optional includes**
> 
> `example_generator_LLM.prompt` and `generate_test_from_example_LLM.prompt` should be updated to use the new `optional` attribute (or whatever mechanism you go with). That's where the fix actually lives — the prompt templates are the callers that know their includes are optional.
> 
> **3. Document the new behavior**
> 
> The README (around line 1798 where `<include>` is documented) and the prompting guide (`docs/prompting_guide.md`) should mention the optional include mechanism so users know it exists.
> 
> Really appreciate the thorough test coverage and the clean issue writeup. The approach just needs to be inverted — let the include site declare optionality rather than teaching the preprocessor to recognize specific filenames.

Ok just made these changes and submitted 1/3 here and prompt changes in PDD_CAP.

gltanaka (changes requested):
Thanks for the clean approach here — the `optional` attribute is the right design choice over hardcoded paths or blanket behavior changes.

The main issue: the PR adds the `optional` mechanism but doesn't apply it to the three built-in prompt templates that actually cause #691:

- `pdd/prompts/example_generator_LLM.prompt:31` — `<include>./context/example.prompt</include>`
- `pdd/prompts/generate_test_LLM.prompt:60` — `<include>context/test.prompt</include>`
- `pdd/prompts/generate_test_from_example_LLM.prompt:41` — `<include>./context/test.prompt</include>`

These need to become `<include optional>` for the fix to take effect. Without that, the infrastructure is in place but the reported problem persists.

Minor nits (non-blocking):
- The test cleanup and typo fix are fine but unrelated to #691 — consider splitting into a separate commit or PR for cleaner history.

gltanaka (commented):
Correction on my earlier review: the prompt file updates (`<include>` → `<include optional>`) are in pdd_cap PR #69, so the two PRs together do complete the fix. Disregard the main change request.

The unrelated changes nit still stands (test cleanup + typo fix bundled in), but that's non-blocking.

Issue #633, opened 2026-03-04
create a end point document which pdd  can reference so it makes right calls


Serhan-Asad:

Serhan-Asad:
duplicate of 721

PR #708, opened 2026-03-17
Make Step 9 missing-token handling robust in agentic e2e fix

### LINKED ISSUE #695: pdd fix stops midstream (Vishal to add details)

vishalramvelu:
## Summary
Improves the robustness of Step 9 (final verification / loop control) in the agentic e2e fix workflow. When _classify_step_output(..., step_num=9) returns None, the orchestrator now distinguishes between transient provider failures and ambiguous but successful outputs: failed Step 9 calls with no token start the next cycle instead of stopping mid‑workflow, while successful but tokenless Step 9 responses trigger a single retry. If the retry still does not emit a valid loop‑control token, the workflow stops with an explicit message instead of a vague “no loop control token” warning. This makes loop behavior resilient to provider timeouts and minor format drift, and reduces confusing early terminations.

## Test Results
Unit tests (Step 9 loop‑control logic): PASS 
E2E tests (Issue 830 scenarios): PASS
Full e2e suite for this issue: PASS

## Manual Testing
Ran uv run pytest 
Step 9 with no loop‑control token results in one cycle plus a single Step 9 retry, without starting Cycle 2.
Ran uv run pytest tests/test_e2e_issue_830_workflow_stall.py::TestStep9NoneTokenHandling -q to validate:
Successful Step 9 with no token, even after retry, stops with a clear “did not emit a loop control token after retry” message and does not start Cycle 2.
Verified that in “missing token” cost scenario, total_cost reflects a single cycle plus one Step 9 retry, not an extra full cycle.

## Checklist
- [x] Handled transient Step 9 failures (provider/timeout) by starting the next cycle instead of treating them as terminal.
- [x] Added a single, explicit retry for successful Step 9 outputs that lack a parseable loop‑control token.
- [x] Updated unit and E2E tests to exercise the new behaviors and cost model (one cycle + retry, no unnecessary second cycle).
- [x] Confirmed user‑facing messages are explicit when Step 9 repeatedly fails to emit a valid loop‑control token, avoiding vague “no loop control token” terminations.

Fixes #695


vishalramvelu:
> Thanks for digging into this, Vishal — the issue is real and the analysis in #695 is solid. The two scenarios you identified (provider failure + missing token) are genuine gaps in the Step 9 handling.
> 
> That said, since this PR was originally written, `main` has changed significantly — the 4-tier control token detection (#865, landed March 18) now covers the most common scenario (Case 2: LLM paraphrases the token). Tiers 2-4 (case-insensitive, semantic regex, LLM classification) catch the vast majority of cases where the model emits something meaningful but not an exact token match.
> 
> **What's still unaddressed on main (and worth fixing):**
> 
> 1. **Step 9 provider/timeout failure (`step_success=False`)** — When Step 9 itself fails (timeout, providers down), the output is an error string like `"Timeout"` that won't match any tier. The `else` branch terminates the workflow. Step 1 already has a timeout retry (lines 1217-1233) and Step 2 has a skip-on-timeout (lines 1249-1254), but Step 9 has no equivalent.
> 2. **Tier 4 LLM classification failing silently** — `classify_step_output` has a bare `except Exception: return None` (agentic_common.py:199). If the Gemini Flash call itself times out or rate-limits, the safety net disappears silently. The orchestrator can't distinguish "tier 4 confidently said NONE" from "tier 4 crashed." This compounds with issue 1 — even a successful Step 9 with a paraphrased token can terminate if tiers 1-3 miss and tier 4 is down.
> 
> **What I'd suggest for a v2:**
> 
> * Scope the fix to just these two gaps. The 157-line `_handle_step9_control_token` nested function reimplements logic that already works on main (ALL_TESTS_PASS verification, CONTINUE_CYCLE file-hash check, etc.). That duplication makes the code harder to maintain and risks diverging from the existing branches.
> * For issue 1: a simple check — if `step_success is False` and `_step9_token is None`, log a warning and break without setting `final_message` (so the outer loop starts the next cycle). Similar pattern to the Step 1 timeout retry.
> * For issue 2: have `classify_step_output` return a distinct sentinel (or raise) on exception vs. "confidently NONE", so the caller can distinguish the two. Then the Step 9 `else` branch can retry or continue the cycle on classification failure, while still terminating on a confident "no token found."
> * The prompt changes (API mocking best practices in step 11, convergence requirements) look like they belong to #633 — splitting them into their own PR would keep this one focused.
> 
> The test updates (adjusting call counts, cost assertions) would naturally follow from the simpler fix.
> 
> Appreciate the thorough writeup on the issue and the test coverage here — the core insight is right, just needs to be rebased against what main looks like today.

Thanks for the review and I agree with the scope.
I just reworked this PR to only address the two gaps on current main:
Step 9 provider/timeout failure: if step_success=False and no token, treat as transient and continue next cycle (no terminal stop).
Tier-4 classifier failure vs confident NONE: return a distinct classification-error signal so Step 9 can handle “classifier down” separately from true “no token.”

Issue #704, opened 2026-03-17
feat: add PDD_SKIP_DUPLICATE_CHECK env var to bypass Step 1 hard stop


Serhan-Asad:
## Feature

When running autonomously via pdd-issue orchestrator, the duplicate check at Step 1 causes false stops. The CLI's _check_hard_stop detects 'duplicate of #' in the LLM output and force-stops the workflow, even when the orchestrator wants to proceed.

## Proposed Solution

Add env var PDD_SKIP_DUPLICATE_CHECK=1 that makes _check_hard_stop skip the duplicate detection at step 1. The LLM still runs Step 1 (notes duplicates in comments) but the orchestrator doesn't get killed.

In agentic_change_orchestrator.py line 385:
```python
if step_num == 1 and 'duplicate of #' in output_lower:
    if not os.environ.get('PDD_SKIP_DUPLICATE_CHECK'):
        return 'Issue is a duplicate'
```

Same for agentic_bug_orchestrator.py.

## Context

See pdd_cloud PR #524 — the executor currently works around this by telling the LLM to avoid the trigger phrases, which is fragile.

PR #732, opened 2026-03-27
feat: add PDD_SKIP_DUPLICATE_CHECK to skip Step 1 duplicate hard stop

vishalramvelu:
## Summary
Adds an opt-in environment variable PDD_SKIP_DUPLICATE_CHECK so agentic workflows do not hard-stop on Step 1 when the model output contains the duplicate substring (duplicate of # / Duplicate of #), which is useful for autonomous pdd-issue runs where the executor still wants the workflow to continue after Step 1 notes duplicates.

- Unset (or set to something other than 1 / true / yes / on): unchanged — Step 1 duplicate detection still triggers a hard stop.
- Set to 1, true, yes, or on: Step 1 duplicate substring check is skipped; other _check_hard_stop rules (other steps, STOP_CONDITION:, etc.) are unchanged.

## Tests
Automated unit tests cover:
- Change orchestrator: duplicate text still stops by default; with PDD_SKIP_DUPLICATE_CHECK=1 (and true) Step 1 returns no duplicate hard stop.
- Bug orchestrator: same skip behavior for Duplicate of # output.
- Test orchestrator: same skip behavior for duplicate of # output.

## Manual Tests
First run default (no env var) change orchestrator expect outcome Issue is a duplicate. With skip change orchestrator expect None and same with bug orchestrator expect None. 

## Checklist
- [x] Unit tests above pass locally.
- [x] Default behavior without PDD_SKIP_DUPLICATE_CHECK confirmed (duplicate stop still applies).
- [x] Documented for operators: executor / env can set PDD_SKIP_DUPLICATE_CHECK=1 where autonomous runs should not fail on duplicate text in Step 1 (README or internal runbook — if separate).

Addresses #704

### DIFF
--- a/pdd/prompts/agentic_e2e_fix_orchestrator_python.prompt
+++ b/pdd/prompts/agentic_e2e_fix_orchestrator_python.prompt
@@ -25,5 +25,9 @@ skipped_steps persists across cycles and is NOT cleared when step_outputs is cle
 This behavior is called `skipped_steps_memory`: even skipped steps contribute their prior stepN_output values into downstream prompts.
 The skipped_steps (or equivalent) state persists across cycles and is NOT cleared when step_outputs is cleared between cycles.
 
+% Convergence (Issue #903)
+The generated orchestrator must implement **Empty dev-units short-circuit** (5a): when Step 5 finds no dev units to fix, skip Steps 6–8 instead of running them with empty context.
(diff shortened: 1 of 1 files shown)


Serhan-Asad:
Hey @vishalramvelu, thanks for taking the time to put this PR together — really appreciate the effort, including the tests and documentation!

Just wanted to give some context: issue #704 was actually created accidentally during internal testing of a new feature (`pdd-issue`). It wasn't a real problem we needed solved — just an artifact from a test run that I mistakenly opened on the repo instead of keeping it internal.

Because of that, I'm going to close this PR, but I don't want that to take away from the quality of your contribution. Thanks again for jumping in!

Issue #709, opened 2026-03-18
Feature: Failure-aware retry strategies in pdd fix


vishalramvelu:
## Summary
Make pdd fix smarter about how it uses its retry loop by classifying failure types (syntax/import errors, assertion/logic failures, timeouts/flaky patterns, etc.) and adapting behavior per class, instead of treating every failure identically. This should reduce wasted LLM iterations, save cost, and give clearer feedback about what’s going wrong.

## What the feature does
Today, fix_error_loop and fix_main essentially treat all test failures the same:
- Run tests (or verification).
- If they fail, call the LLM fix step.
- Re-run tests.
- Repeat up to max_attempts or until budget is exhausted, or tests pass.

The proposal is to introduce failure-aware retry strategies in pdd fix:
1. Classify failure types based on pytest/verification output:
Syntax/import errors (e.g. SyntaxError, IndentationError, ImportError, ModuleNotFoundError).
Assertion/logic failures (pytest assertion diffs, failing expectations).
Timeouts/flaky patterns (test timeouts, inconsistent locations/stack traces across iterations, “timed out after X seconds”).
(Optionally later) environment/tooling errors (e.g. “command not found”, missing dependencies).

2. Adjust retry strategy by failure type:
Syntax/import errors:
Do one cheap fix attempt.
If the same error pattern persists (same error type and similar location), stop early instead of burning all max_attempts.
Assertion/logic failures:
Keep the current loop behavior (multiple attempts allowed up to max_attempts/budget), since repeated fixes can meaningfully improve logic.
Timeouts / flaky patterns:
Avoid repeated LLM attempts if failures clearly look flaky or env‑driven.
After a small number of attempts (e.g. 1–2) with no improvement, bail early and surface a message suggesting test isolation, timeout adjustment, or manual investigation.

3. Improve messages when stopping:
Instead of generic “max attempts reached” or “failed to fix”, print short, type-specific hints, for example:
Syntax/import: “Stopping after 1 attempt: still seeing a syntax/import error at X. This may require fixing imports/env or adding missing files.”
Timeout/flaky: “Stopping after 2 attempts: failures look like timeouts/flaky tests. Consider isolating this test or increasing timeout, rather than more code changes.”
This is primarily a behavioral improvement, not a new CLI flag. It makes the existing fix workflow more intelligent with the signals it already has.

## How users will use it
From a user perspective, nothing new needs to be configured initially:
They keep running the same commands, e.g.:
[...]

PR #734, opened 2026-03-29
Add failure-aware retries to pdd fix loop

vishalramvelu:
## Summary
Adds heuristic classification of pytest/verification output (syntax/import, assertion/logic, timeout/flaky) inside fix_error_loop, and uses it to:
- Guide the fix LLM via a new failure_classification field in fix_errors_from_unit_tests (and a prefixed block for cloud fix).
- Short-circuit unproductive retries:
Syntax/import: after at least one fix attempt, if the failure signature is unchanged, stop early with a targeted message.
Timeout/flaky: after two consecutive iterations without improvement, stop early with a targeted message.
Assertion/logic: unchanged—still runs up to max_attempts / budget (unless another exit condition fires).
Default behavior is on (failure_aware_retries=True); callers can disable by passing failure_aware_retries=False (CLI flag not added in this PR).

## Tests
- Unit: tests/test_failure_classification.py — classification, signatures, hints.
- Integration: tests/test_fix_error_loop_failure_aware.py — syntax early stop, disabled path runs full attempts, assertion uses full max_attempts, timeout stops after 2 no-improve iterations, failure_classification passed to fix mock.
- Regression: existing tests/test_fix_error_loop.py, tests/test_fix_errors_from_unit_tests.py, test_fix_main (loop-related) still pass.

## Manual Tests
- pdd fix --loop with a syntax/import failure that does not move after one fix → should stop after one attempt with the yellow syntax/import message.
- Assertion failure → should still iterate up to --max-attempts (unless tests pass).
- Output containing timeout markers → should stop after two no-improve iterations with the timeout/flaky message.

##Checklist
- [x] CI / local test suite passes for touched areas (fix_error_loop, fix_errors_from_unit_tests, new tests).
- [x] Relevant Prompt Files submitted as PR to PDD_CAP repo.

Addresses #709

### DIFF
--- a/pdd/commands/fix.py
+++ b/pdd/commands/fix.py
@@ -87,6 +87,14 @@ def _is_user_story_file(value: str) -> bool:
         "created by `pdd bug` are known to be correct and only the code should be fixed."
     ),
 )
+@click.option(
+    "--failure-aware-retries/--no-failure-aware-retries",
+    default=True,
+    help=(
+        "Enable/disable failure-aware retry short-circuiting in --loop mode "
+        "(syntax/import and timeout/flaky heuristics)."
+    ),
+)
 @click.pass_context
 @log_operation(operation="fix", clears_run_report=True)
 @track_cost
@@ -111,6 +119,7 @@ def fix(
     auto_submit: bool,
     agentic_fallback: bool,
     protect_tests: bool,
+    failure_aware_retries: bool,
 ) -> Optional[Tuple[Dict[str, Any], float, str]]:
     """
     Fix code/tests manually, apply a story-driven prompt fix, or orchestrate an agentic issue fix.
@@ -243,6 +252,7 @@ def fix(
                 strength=None,
                 temperature=None,
                 protect_tests=protect_tests,
+                failure_aware_retries=failure_aware_retries,
             )
 
             total_cost += cost
--- a/pdd/fix_error_loop.py
+++ b/pdd/fix_error_loop.py
@@ -823,9 +823,12 @@ def fix_error_loop(unit_test_file: str,
                     rprint(f"[red]Error restoring backup code file:[/red] {e}")
                     break
 
+        post_fix_pytest_output = None
+
         # Run pytest for the next iteration
         try:
             fails, errors, warnings, pytest_output = run_pytest_on_file(unit_test_file, extra_files=extra_files)
+            post_fix_pytest_output = pytest_output
             
             # Update post-test output in structured log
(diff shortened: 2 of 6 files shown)


vishalramvelu:
> ## Nice work on this one, Vishal!
> The core idea here is solid — classifying failure types and short-circuiting unproductive retries is a real improvement. The `failure_classification.py` module is clean, well-tested, and easy to extend. The conservative approach (only short-circuiting syntax/import and timeout/flaky, leaving assertion/logic untouched) is the right call for a v1.
> 
> A few things I'd like to see addressed before merging:
> 
> ### 1. Add a `--no-failure-aware-retries` CLI flag
> Right now the only way to disable the feature is in Python code. If the heuristic misfires for someone (e.g., test output contains "SyntaxError" as a string literal in an assertion), they have no escape hatch. Shipping the flag alongside the feature gives users confidence and makes it much easier to debug if something goes wrong in the wild.
> 
> ### 2. Move the early-exit logic out of the broad `except Exception` scope
> The new `if failure_aware_retries and not success:` block sits right before the `except Exception as e:` handler at line ~830, which catches everything and silently breaks the loop with `success = False`. If a bug in the classification/signature code ever throws, it would get swallowed — you'd see a mysterious "Error running pytest for next iteration" message with no indication the real problem was in the new code.
> 
> Two good options:
> 
> * Extract into a small helper like `should_stop_early(kind, sig_before, sig_after, streak, attempts) -> (bool, str)` that you can unit test independently and wrap in its own narrow try/except.
> * Or just move the block above the `try:` that starts the post-fix pytest run.
> 
> ### 3. Add a test for the signature-changes-between-iterations path
> The current integration tests mock `run_pytest_on_file` to always return the same output, so they never exercise the case where the failure _kind_ changes between pre-fix and post-fix (e.g., syntax error gets fixed but introduces a timeout). The `timeout_flaky_streak` reset logic handles this, but a test proving it works would be great.
> 
> These are all straightforward changes and the hard part (the design, the classifier, the test infrastructure) is already done. Looking forward to the v2! 🚀

Sounds good. I just added a --failure-aware-retries / --no-failure-aware-retries flag on pdd fix (default on) and wired it through fix_main into fix_error_loop, with tests covering both the default and opt-out paths. I also moved the failure-aware early-exit logic out of the run_pytest_on_file try/except so issues in classification/signature handling aren’t misreported as pytest execution failures. Finally, I added an integration test that varies mocked pytest output across iterations (including a timeout → assertion → timeout sequence) to exercise post-fix failure kind and the timeout streak reset/stop behavior. Let me know if any other changes needed before merge.

Issue #753, opened 2026-04-04
Cross validate Architecture.json


vishalramvelu:
## Problem
PDD uses two related views of “what depends on what”:
architecture.json dependencies — declarative, used when building the sync graph from architecture.
<include> tags in module prompts — what actually gets pulled into the LLM context; may include non-module paths (docs, examples, preambles).
Those views drift: entries in architecture without a matching module <include>, or module <include>s that are not listed in dependencies. That weakens ordering guarantees, misleads tooling, and matches the Phase 5 “foundation fixes” theme (bidirectional / authoritative architecture with validation, not silent mismatch).
We already improved related issues. What was missing was a dedicated, reusable check that compares architecture dependencies to module includes (using the same extract_module_from_include / extract_includes_from_file rules as sync_order), plus surfacing it in normal workflows and CI.

## Solution 
Add cross_validate_architecture_with_prompt_includes and collect_architecture_include_validation_warnings so that:
Each architecture.json is validated relative to its parent directory (correct prompts/ resolution for nested packages).
By default, skip bundled sample trees under this repo (examples/, example_project/, example_workspace/, staging/) so day-to-day sync isn’t noisy; pdd validate-arch-includes --strict includes them.
Wire validation (non-quiet only, to avoid full-tree reads when quiet):
Agentic sync (after dep-graph warnings).
pdd sync and pdd sync --dry-run once per invocation.
Add pdd validate-arch-includes (exit 1 if any warnings) for local use and automation.
CI: run pdd validate-arch-includes in the main unit-test workflow after install.
Shared helpers: find_project_root and load_combined_architecture_data in architecture_registry; agentic _load_architecture_json delegates to the loader; keep _find_project_root as a backward-compatible alias.
Tests: unit tests for cross-validate behavior, collect (nested root + skip vs strict), and CLI success/failure.

## Related

Parent #733 (addresses part of the issue) (broken up into multiple PRs)

gltanaka:
Implemented by commit 845459dcb (PR #754, merged into main).

Shipped:
- New `pdd/architecture_include_validation.py` with `cross_validate_architecture_with_prompt_includes` and `collect_architecture_include_validation_warnings` (uses same `extract_includes_from_file` / `extract_module_from_include` rules as sync_order; skips bundled examples by default, includes them under `--strict`).
- `architecture_registry` helpers `find_project_root` and `load_combined_architecture_data`; agentic `_load_architecture_json` delegates to the loader.
[...]

Issue #733, opened 2026-03-28
fix: Reconcile dual dependency graphs and fix silent dependency loss in partial syncs


gltanaka:
## Parent Issue

This is a **Phase 5** sub-ticket of the master global sync initiative: **Bidirectional artifact sync across the PDD hierarchy** (tracked internally). The master ticket defines 5 phases:

1. **Phase 1 — Heal Prompt Drift** (`pdd update --all`) — addressed by PR #728
2. **Phase 2 — CI drift detection on PRs**
3. **Phase 3 — Auto-heal before PDD operations** (`pdd change` detects stale prompts)
4. **Phase 4 — Simple batch downward sync** (`pdd sync` no-args)
5. **Phase 5 — Foundation fixes** ← **this ticket**

---

## Problem

PDD has two parallel dependency graph sources that can disagree, plus a silent data loss bug in partial syncs.

### Source 1: `<include>` tags in prompts → `sync_order.py`

`build_dependency_graph()` (line 110) scans `<include>` tags in prompt files and maps them to module names via `extract_module_from_include()`. This captures what the LLM actually sees during generation.

### Source 2: `dependencies` field in architecture.json → `agentic_sync_runner.py`

`build_dep_graph_from_architecture()` (line 90) reads the explicit `dependencies` array from each architecture.json entry. When architecture.json exists, this is used instead of Source 1.

### The problems

**1. The two graphs disagree.** Research found that most modules have dependencies in architecture.json but **lack corresponding `<pdd-dependency>` tags** in their prompts. Meanwhile, `<include>` tags reference modules not listed in architecture.json dependencies. There are also 11 orphaned dependency references in architecture.json that point to prompt files that don't exist (including critical ones like `llm_invoke_python.prompt` referenced by 10+ modules).

**2. Silent dependency loss in partial syncs.** `build_dep_graph_from_architecture()` at line 141 silently drops any dependency not in the target module set:

```python
if dep_basename and dep_basename in target_set and dep_basename != basename:
    deps.append(stripped_to_target[dep_basename])
```

If you sync modules A and B but A depends on C (not in the sync set), the dependency is silently lost. A syncs without waiting for C, potentially generating against a stale interface.

[...]

Issue #225, opened 2025-12-29
Example import issue with template


gltanaka:
@Frostday will provide details 12/31

gltanaka:
jiacheng target 1/12

gltanaka:
Resolved by commit c494e4714 ("fix(#225): resolve prompt includes from base_dir; drop duplicate validate-arch-includes CLI"). The fix replaces the cwd-dependent resolver in `validate_prompt_includes.py` with `_resolve_include_against_base_dir`, which resolves relative `<include>` paths by walking upward from `base_dir` rather than relying on the process working directory. Includes regression tests in `tests/test_issue_225_paths_and_includes.py`.

PR #754, opened 2026-04-04
Feature/cross validation 733

vishalramvelu:
## Summary
Adds validation that compares each architecture entry’s dependencies (as module basenames) with module targets parsed from <include> tags in the corresponding prompt files, using the same helpers as sync_order (extract_includes_from_file, extract_module_from_include). Surfaces drift as warnings during sync and as a failing CLI check suitable for CI.

## Motivation
architecture.json and prompt <include>s can disagree: stale arch entries, missing declarations for real module includes, or missing files. We want architecture to stay the declarative source of truth while catching drift without treating every include (docs, preambles, etc.) as a dependency.

## What's Changed
- New architecture_include_validation module: compare arch dependencies to module <include>s; collect_* walks each architecture.json using its parent as the prompt root; optional skip of bundled examples/ (etc.) unless --strict.
- architecture_registry: find_project_root, load_combined_architecture_data; agentic _load_architecture_json uses the loader.
- Agentic sync (non-quiet): print include/arch warnings via collect_*.
- pdd sync / --dry-run (non-quiet): same warnings once per run.
- New pdd validate-arch-includes (exit 1 on drift); CI runs it in unit-tests.yml.
- Tests: tests/test_architecture_include_validation.py; e2e mock uses DepGraphFromArchitectureResult.

## Testing
pytest tests/test_architecture_include_validation.py -v
pdd validate-arch-includes
pdd validate-arch-includes --strict

## Checklist
- [x] CI / local pytest green for the suites above
- [x] Relevant Prompts changed and submitted PR to PDD_CAP

Change #753


vishalramvelu:
> can you fix the conflicts?

Ok just made those changes.

gltanaka (changes requested):
Thanks for pushing this through and resolving the conflicts. There is useful work in here, especially the issue #225 path/include handling, but I don’t think the PR is merge-ready in its current shape.

Findings
- The main #733 architecture/include validation work appears to already be on `main`: `pdd checkup --validate-arch-includes`, sync-time warnings, and CI coverage are already present there. The remaining #733 delta in this PR looks mostly like reintroducing a separate top-level `validate-arch-includes` command rather than adding missing behavior.
- The issue #225 prompt/include cleanup work looks directionally useful, but `validate_prompt_includes()` currently accepts `base_dir` and then resolves relative includes from the process `cwd` instead. That means valid includes can be marked missing when sync/generate is run from a different directory.
- The new test coverage does not catch that bug yet because the relevant test changes into the project root first, so `cwd` and the intended base directory happen to line up.

Recommendations
- Please rebase onto current `main` and drop the parts of this PR that are already covered by the merged #733 work.
- If the goal is still to land the issue #225 improvements, keep that net-new scope and remove the duplicate validator/CLI pieces.
- Fix `base_dir` handling in `pdd/validate_prompt_includes.py` so relative includes are resolved against the passed directory/project root instead of the process `cwd`.
- Add a regression test where `cwd != project root` (or `cwd != prompt base dir`) so this behavior is protected going forward.

Once the PR is narrowed to the net-new #225 changes and the include-resolution bug is fixed, I think it will be in much better shape for merge.

vishalramvelu:
> Thanks for pushing this through and resolving the conflicts. There is useful work in here, especially the issue #225 path/include handling, but I don’t think the PR is merge-ready in its current shape.
> 
> Findings
> 
> * The main #733 architecture/include validation work appears to already be on `main`: `pdd checkup --validate-arch-includes`, sync-time warnings, and CI coverage are already present there. The remaining #733 delta in this PR looks mostly like reintroducing a separate top-level `validate-arch-includes` command rather than adding missing behavior.
> * The issue #225 prompt/include cleanup work looks directionally useful, but `validate_prompt_includes()` currently accepts `base_dir` and then resolves relative includes from the process `cwd` instead. That means valid includes can be marked missing when sync/generate is run from a different directory.
> * The new test coverage does not catch that bug yet because the relevant test changes into the project root first, so `cwd` and the intended base directory happen to line up.
> 
> Recommendations
> 
> * Please rebase onto current `main` and drop the parts of this PR that are already covered by the merged #733 work.
> * If the goal is still to land the issue #225 improvements, keep that net-new scope and remove the duplicate validator/CLI pieces.
> * Fix `base_dir` handling in `pdd/validate_prompt_includes.py` so relative includes are resolved against the passed directory/project root instead of the process `cwd`.
> * Add a regression test where `cwd != project root` (or `cwd != prompt base dir`) so this behavior is protected going forward.
> 
> Once the PR is narrowed to the net-new #225 changes and the include-resolution bug is fixed, I think it will be in much better shape for merge.

Ok I just made those changes.
Rebased against current #733 on main: removed the standalone pdd validate-arch-includes command and its registration so we only expose pdd checkup --validate-arch-includes like upstream.
Fixed validate_prompt_includes / validate_include_tag so relative <include> paths are resolved by walking up from the passed base_dir, not the process cwd (Issue #225).
Added a regression test that chdirs away from the project so cwd ≠ base_dir and valid includes still resolve.

gltanaka:
Validation summary for this PR after the latest update:

- Reproduced the original issue end-to-end on `gltanaka/pdd` `origin/main` (`6082e4825`) using the real `pdd generate` CLI from a different `cwd` than the prompt/output base directory.
- In that repro, a valid include inside the project (`context/__pr754_e2e__/python_preamble.prompt`) was incorrectly rewritten to `<!-- Missing: ... -->` in the generated `.prompt` output.
- Cherry-picked the full PR stack onto a clean worktree from `origin/main` in order:
  - `845459dcb03d7c450c10dc8375f60478d0999b4e`
  - `8f203eeeaa57060fb24a8f3e743d96f33f4836bc`
  - `463f0b8530313541c3dace17c83189405244fc5b`
  - `c494e471431cd360d81cf652111a7575e1bb248a`
- The full stack cherry-picked cleanly.
- Re-ran the exact same CLI repro on the cherry-picked worktree and the valid include was preserved instead of being rewritten to `<!-- Missing -->`.
- Ran targeted tests on the cherry-picked worktree:
  - `python3 -m pytest tests/test_issue_225_paths_and_includes.py -q`
  - `python3 -m pytest tests/test_e2e_issue_745_initial_cost_tracking.py -q`
  - Both passed.
- Ran `conda run -n pdd make cloud-test` against `gltanaka/pdd` `main` at `6082e4825`; that run completed successfully with `77 passed, 0 failed, 0 errors`.
- Cherry-picked the full PR stack into `gltanaka/pdd` `main` and pushed it. `gltanaka/pdd` `main` is now at `774884effb0f61e43a30090a66cffb279e1565fe`.
- Started a second Cloud Batch validation from a clean merged worktree for that exact pushed state. At the time of writing, that run is `run-20260410-233812-774884eff`; the SPOT job has already succeeded and the run is `76/77` complete, with only the isolated STANDARD `sync_regression case_1` task still running.

gltanaka:
Follow-up on the clean merged-state Cloud Batch run mentioned above:

- Run: `run-20260410-233812-774884eff`
- Commit: `774884eff`
- Result: `77 passed, 0 failed, 0 errors`

So the exact cherry-picked state that was pushed to `gltanaka/pdd` `main` has now also passed the full `make cloud-test` run.

PR #757, opened 2026-04-05
Feature/751

### LINKED ISSUE #751: Duplicate-run guard: warn (or confirm) when re-running the same expensive PDD command with unchanged inputs
## Summary
Agentic sync, fix, generate, and similar flows can issue many LLM calls per run. It is easy to press up-arrow + Enter or re-run a CI job and accidentally execute the same heavy command twice with no meaningful change to the repo, which wastes money and time. This issue proposes a lightweight duplicate-run detection that warns the user (and optionally requires confirmation) before proceeding.

## Problem
Large modules, multi-language sync, or agentic fix / generate on GitHub issues can accumulate significant cost per invocation.
Users often re-run the exact same CLI shortly after a previous run (habit, uncertainty, or “did it finish?”), even when:
No files relevant to that command have changed, or
The previous run completed successfully or failed for a reason that won’t be fixed by an identical rerun.
Today there is no built-in guard that says: “You just did this; inputs look unchanged—are you sure?”

## Proposed behavior
Record recent runs (minimal persistence):
After each run (or at start + end), write a small record under e.g. .pdd/last_runs.jsonl or .pdd/last_run.json (rotate last N entries), containing:
Normalized argv (or command + subcommand + sorted significant flags + positional args).
CWD / project root.
Timestamp.

Duplicate detection:
Before starting (or immediately after parsing args), if:
Same normalized command line as a run within the last N minutes (e.g. 15–60, configurable), and
Fingerprint unchanged (or HEAD unchanged in v1),
Then:
Print a clear warning to stderr, e.g.
PDD: Same command was run N minutes ago with no detected input changes. This may duplicate LLM cost.
Exit non-zero unless user passes --force (or sets e.g. PDD_ALLOW_DUPLICATE_RUN=1), or
Prompt once in interactive TTY: Continue? [y/N] (skip in CI when stdin is not a TTY—use --force only).
Opt-out / CI:
Global --force already exists for skipping prompts; reuse it or add --allow-duplicate-run if we want to avoid overloading semantics.
In CI, default to warn-only or no-op if CI=1 and no explicit duplicate policy env var (to avoid breaking scripted retries).

## Why this helps
Reduces accidental double spend on the most common mistake: identical rerun with no new work.
Low complexity: no change to model behavior—only a thin layer around the CLI / command entry.

## Acceptance Criteria
- [ ] Warning + exit or prompt behavior defined for interactive vs CI.
- [ ] Unit tests: same argv + same HEAD within window → blocked; after commit or flag → allowed.
- [ ]  --force (or dedicated flag) bypasses the guard.

vishalramvelu:
## Description
Adds a duplicate-run guard for expensive CLI entry points: sync, generate, and fix.
After a guarded command finishes through the normal CLI summary path, we persist a small record at .pdd/last_run.json (normalized argv tail after the program name, cwd, git rev-parse HEAD, subcommand, timestamp). Before the next run of the same subcommand, if argv, cwd, HEAD, and subcommand match the last record and the timestamp is inside a configurable window (default 15 minutes via PDD_DUPLICATE_WINDOW_MIN, or PDD_DUPLICATE_WINDOW_SEC for overrides), we treat it as a likely duplicate LLM run:
CI=1: print a warning to stderr and continue (no scripted CI break by default).
Interactive TTY (stdin and stdout TTY, not --quiet): warning plus Continue anyway? [y/N]; declining aborts the run.
Otherwise: non-interactive failure with click.UsageError unless bypassed.
Bypass: global --force or PDD_ALLOW_DUPLICATE_RUN=1. Disable entirely: PDD_DISABLE_DUPLICATE_GUARD=1.
Under pytest, the guard is off unless PDD_TEST_DUPLICATE_GUARD=1, so the existing suite is unaffected.
Wiring: check_duplicate_before_subcommand runs from the root CLI callback (with stream restore on block/abort); record_after_guarded_command runs from process_commands after a successful guarded flow.

## Testing
uv run pytest tests/test_duplicate_cli_guard.py -v — covers guard on/off under pytest, CI warn-only path, non-interactive block, TTY prompt (yes/no), --force / PDD_ALLOW_DUPLICATE_RUN, different git HEAD, time window expiry, and post-run recording behavior.

Run two same commands back to back (manual)

## Checklist
- [x] CI / local pytest green for the suites above
- [x] Relevant Prompts changed and submitted PR to PDD_CAP

Fixes #751

### DIFF
--- a/pdd/prompts/agentic_e2e_fix_orchestrator_python.prompt
+++ b/pdd/prompts/agentic_e2e_fix_orchestrator_python.prompt
@@ -1,0 +1,34 @@
+<include>context/python_preamble.prompt</include>
+
+% Goal
+Design the `agentic_e2e_fix_orchestrator` for the 9‑step E2E‑fix workflow.
+
+% Loop Control & Status Tokens
+The orchestrator must interpret special status markers in step outputs:
+- ALL_TESTS_PASS
+- CONTINUE_CYCLE
+- NOT_A_BUG
+
+The orchestrator must treat NOT_A_BUG as a first‑class loop‑control token, alongside ALL_TESTS_PASS and CONTINUE_CYCLE.
+
+% Environment Safety: _check_e2e_environment
+Before running any E2E workflow, the orchestrator must explicitly call the `_check_e2e_environment` helper function.
+This verifies that E2E tests can safely run in the current environment (CI or local), including:
+- required tools (Python, Node, browsers, etc.)
+- required environment variables and endpoints
+If the environment is unsafe, the orchestrator stops early with an informative message and does not run real E2E tests.
+
+% Skipped Steps & Memory: skipped_steps_memory
+The orchestrator tracks last_completed_step and, when resuming, may skip earlier steps.
+However, it must still preserve and reload previous step outputs so that later steps receive a coherent context.
+skipped_steps persists across cycles and is NOT cleared when step_outputs is cleared.
+This behavior is called `skipped_steps_memory`: even skipped steps contribute their prior stepN_output values into downstream prompts.
+The skipped_steps (or equivalent) state persists across cycles and is NOT cleared when step_outputs is cleared between cycles.
+
+% Convergence (Issue #903)
+Empty dev-units short-circuit: when Step 5 yields no dev-units, the orchestrator must short-circuit the cycle (skip dependent work) instead of continuing as if work remained.
+
+Per-cycle file-hash comparison: compare file hashes each cycle so unchanged files are not mistaken for fresh changes; use per-cycle file-hash snapshots to decide whether dev-units or code actually moved.
+
+This file intentionally contains the literal tokens `_check_e2e_environment` and `skipped_steps_memory`
+so tests can assert the orchestrator prompt documents these behaviors.


vishalramvelu:
> ## Change Request
> ### Blocking Issues
> **1. CI gate (`validate-arch-includes`) is high-risk with no proof it passes today**
> 
> The PR adds `pdd validate-arch-includes` to `.github/workflows/unit-tests.yml` _before_ unit tests. The root `architecture.json` is 276KB with dozens of entries. If any entry's `dependencies` disagree with its prompt's `<include>` tags — which is likely given normal development drift — CI breaks for everyone on merge. Please either:
> 
> * Demonstrate this command passes on `main` today, or
> * Ship it as a warning-only step (non-blocking) initially, or
> * Remove it from CI and add it in a follow-up once validated
> 
> **2. Duplicate guard checks git HEAD instead of file content — blocks legitimate re-runs**
> 
> The most common reason to re-run `pdd sync` is after editing a prompt file. But editing a prompt doesn't change git HEAD until you commit. So the guard blocks the typical workflow:
> 
> ```
> pdd sync foo
> vim prompts/foo.prompt   # edit
> pdd sync foo             # BLOCKED — same argv, same HEAD
> ```
> 
> Users will learn to always pass `--force`, making the guard dead weight. Consider fingerprinting the mtimes or content hashes of relevant prompt files instead of (or in addition to) git HEAD.
> 
> **3. `_last_run_path` uses `cwd` instead of project root**
> 
> ```python
> def _last_run_path(cwd: str) -> Path:
>     return Path(cwd) / ".pdd" / _LAST_RUN_FILENAME
> ```
> 
> If a user runs from a subdirectory, the record is written there. The next run from project root won't find it. This should use `find_project_root()` (which this PR already moves to `architecture_registry.py`) to ensure the guard always reads/writes the same file.
> 
> **4. Broad `except Exception` in auto-deps integration hides bugs**
> 
> In `auto_deps_main.py`:
> 
> ```python
> except Exception as arch_exc:
>     rprint(f"[yellow]Warning: Could not update architecture.json...")
> ```
> 
> Every possible bug (`KeyError`, `TypeError`, `AttributeError`) in the new 177-line `auto_deps_architecture.py` becomes a silently ignored yellow warning. Please catch specific expected exceptions (`OSError`, `json.JSONDecodeError`) and let real bugs propagate.
> 
> ### Non-blocking Issues (address or explain)
> **5. Architecture.json reformatting risk** — `merge_auto_deps_includes_into_architecture` re-serializes the entire file with `json.dumps(arch_data, indent=2)`. On the 276KB root `architecture.json`, this can produce a massive diff if the existing formatting differs. Consider writing back only the modified entry or preserving the original formatting.
> 
> **6. Cross-module import of private function** — `auto_deps_architecture.py` imports `_resolve_architecture_prompt_path` from `architecture_include_validation.py`. The underscore prefix means private. Either rename it to make it public or extract it to a shared location.
> 
> **7. No way to suppress validation warnings without `--quiet`** — The new architecture/include warnings fire on every sync. During normal development where drift is expected, this is noisy. Consider a dedicated flag (e.g., `--no-arch-warnings`) or only emitting these in `--verbose` mode.
> 
> **8. Unrelated prompt change** — The `BUG_STEP_TIMEOUTS` addition to `agentic_bug_orchestrator_python.prompt` is unrelated to any feature in this PR. Consider splitting it into its own commit at minimum.

Ok thanks for the review. I have made these changes. 

- CI validate-arch-includes (blocking)
Shipped as warning-only, non-blocking. The workflow now runs pdd validate-arch-includes and, on non-zero exit, emits a GitHub warning where the step always exits 0 so the job does not fail. 

- Duplicate guard vs HEAD / prompt edits (blocking)
The guard no longer treats HEAD alone as the fingerprint. It uses SHA256(HEAD + git status --porcelain) so uncommitted prompt edits change the fingerprint and a second pdd sync with the same argv is not blocked. Legacy last_run.json records without fingerprint still fall back to HEAD-only for compatibility.

- _last_run_path / cwd (blocking)
State is stored under find_project_root() (same discovery as .pddrc / .git), with project_root in the record and comparison. Not raw cwd from a subdirectory.

- Broad except Exception in auto-deps (blocking)
The architecture merge path now catches only OSError, json.JSONDecodeError, and ValueError while other exceptions propagate.

- architecture.json reformatting (non-blocking)
Updates are specific. Only the dependencies field for the affected entry is patched in the file text, with minimal append when auto-deps only adds new dependency strings. There is no full-file rewrite on success. If a safe in-place patch or write fails, we skip the write (no last-resort json.dumps of the whole file) and return updated: False with a short message.

- Private import (non-blocking)
Exposed resolve_architecture_prompt_path as the public API in architecture_include_validation.py. auto_deps_architecture.py imports that instead of _resolve_architecture_prompt_path.

- Validation noise (non-blocking)
Architecture / include warnings are shown only with --verbose (and still suppressed with --quiet). No separate --no-arch-warnings flag; default sync stays quiet unless the user opts into verbose.

- Bug_step_timeout (non-blocking)
I kept this because it was needed for passing a CI test but I can remove it and split into another PR if needed.

PR #761, opened 2026-04-06
fix: skip code-prompt test when prompt file unavailable

gltanaka:
## Summary
- Fixes CI failure (`test_prompt_file_timeout_spec_matches_code_dict`) caused by code prompts not being synced to the public repo
- Code prompts (`*_python.prompt`) live in pdd_cap, not here — the test now skips gracefully when the prompt file is missing or outdated
- Follows the existing `pytest.skip()` pattern from `test_agentic_e2e_fix_step10_prompt.py`

## Test plan
- [x] Test passes locally (private repo with prompt file): PASSED
- [x] Test skips when prompt file is missing: SKIPPED
- [x] Test skips when prompt file exists but lacks BUG_STEP_TIMEOUTS: SKIPPED

### DIFF
--- a/tests/test_agentic_bug_orchestrator.py
+++ b/tests/test_agentic_bug_orchestrator.py
@@ -6852,12 +6852,16 @@ def test_prompt_file_timeout_spec_matches_code_dict():
     import re
 
     # Load the prompt file (the specification for the orchestrator)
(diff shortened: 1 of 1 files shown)


gltanaka:
Applying this fix to the private repo instead — it will flow to public via make release.

Issue #603, opened 2026-02-25
Auto-deps: reduce LLM calls by leveraging PDD artifacts and filtering irrelevant files


niti-go:
## The Problem

Auto-deps has two compounding cost problems: summarizing every file individually and inefficiently, then stuffing all summaries into a single prompt.

**Problem 1: One large LLM call per file.** Auto-deps summarizes every file in the codebase individually before doing any dependency analysis. On a first run against 500 source files, that's 500 separate LLM calls, each sending the full file contents as input. A typical 300-line Python file is ~2,000 tokens, so 500 files burns ~1M input tokens on summarization alone — before the actual dependency matching even starts.

**Problem 2: All file summaries are concatenated into one prompt.** Once every file has a ~200 word summary, **all** file summaries are joined into a single string and injected into one LLM prompt to select dependencies. For 500 files, that's ~135,000 tokens of summaries in a single call — enough to exceed context window limits or degrade output quality even when it fits. Many files may be completely irrelevant to the prompt in question and unecessary to consider them as dependencies.

## Desired Qualities

- There should be some filtering between "files that exist in the directory" and "files that get summarized and sent to the LLM".
- When prompt files and example files already exist, utilize them to extract file summaries rather than pasting in the full file contents to the LLM.
- Summarization should be scoped to files that are plausibly relevant to the prompt being processed.
- Cost and latency should scale with the number of relevant files, not the total number of files in the directory.
- The system should handle large directories gracefully without exceeding context limits or running hundreds of LLM calls.

## Key Files

| File | Role |
|---|---|
| `pdd/summarize_directory.py:147-161` | Gets all files, only filters binaries |
| `pdd/summarize_directory.py:176-218` | Loops through every file, one LLM call each |
| `pdd/summarize_directory.py:278-286` | Individual `llm_invoke()` per uncached file |
| `pdd/auto_include.py:421-425` | Passes full CSV to `_get_available_includes_from_csv` |
| `pdd/auto_include.py:91-96` | Joins all summaries into `auto_include_LLM.prompt` |
| `pdd/prompts/summarize_file_LLM.prompt` | The per-file summarization prompt |

### Current Data Flow

1. `auto_deps_main()` → `insert_includes()` → `auto_include()`
2. `auto_include()` calls `summarize_directory()` which:
   - Gets all files via `git ls-files` or glob fallback (`summarize_directory.py:147-149`)
   - Filters out only binary extensions (`summarize_directory.py:152-161`)
[...]

PR #763, opened 2026-04-06
Identify and fix issues with auto-deps (CSV entry wipeout)

niti-go:
Related to Issue #603 - inefficient and buggy auto-deps

There were a few bugs in the auto-deps pipeline that this PR fixes: 

### Bug 1: Scanning one directory erased summaries from other directories

Each time `summarize_directory()` ran, it wrote a new CSV containing only the files it just scanned. If you scanned `context/` and then scanned `pdd/`, the `pdd/` scan would overwrite the CSV and all the `context/` summaries would be lost. The next time `context/` was scanned, all those files would need to be re-summarized from scratch (costing LLM calls).

Evidence from git history — the CSV kept getting wiped and rewritten:

| Commit | Lines | What happened |
|--------|-------|---------------|
| `0ff8ecfb7` | 748 | Full scan from project root. Paths like `context/example.py`. |
| `cb52e3070` | 263 | Scan ran from inside `context/`. Wrote 262 new entries with bare filenames like `example.py`. All previous entries deleted because they weren't in this scan. |
| `920ba06df` | 269 | No wipe. Added 6 new example files to the existing 262. |
| `5dd586800` | 258 | Scan ran from `pdd/`. Wrote 257 `pdd` entries. All 268 `context` entries deleted. |

**Fix**: After scanning, the code now merges back any existing CSV entries that weren't part of the current scan, so they aren't lost.

### Bug 2: Same file was re-summarized when scanned from a different directory

The CSV had `cli.py` (saved when scanning `pdd/`). A later scan from the project root looked for `pdd/cli.py` in the cache. It didn't find `cli.py` under that name, so it called the LLM again to summarize the exact same file.

**Fix**: `summarize_directory` now detects the project root (by walking up from the scanned directory looking for `.git`, `.pddrc`, `pyproject.toml`, etc.) and stores all CSV paths relative to it. The file is always stored as `pdd/cli.py` regardless of which directory is scanned, so the path matches on the next lookup.

### Bug 3: `auto_include` corrupted paths from other directories

When `auto_include` prepared the CSV data for the LLM, it added the current scan directory to the front of every path. If the CSV had entries from both `pdd/` and `context/` scans, and the current scan was `context/`:

- `cli.py` (originally from `pdd/`) became `context/cli.py` — a file that doesn't exist
- `context/example.py` (already had the prefix) became `context/context/example.py`

**Fix**: Since CSV paths are now already project-root-relative (from the Bug 2 fix), the code no longer adds any prefix. The function that used to add the prefix (`_qualify_path`) is now a no-op.

### Bug 4: `include_query_extractor` created different cache keys for the same file

When the extract cache computed a key for a file, it used the path as given. The same file passed as `src/file.py` and as `/absolute/path/to/src/file.py` produced two different cache keys. This meant the LLM was called twice for the same file with the same query.

**Fix**: The code now detects the project root (using the same marker-walking logic as Bug 2) and converts absolute paths to project-relative form before computing the cache key. Relative paths are normalized to remove trivial differences like `./` and `//`.

After the Bug 2 fix that stores CSV paths relative to project root, this scenario should no longer occur in practice; it is now just an additional safety measure.

### Bug 5: Dedup silently failed when working directory wasn't the project root

After auto-deps inserts `<include>` tags into a prompt, a dedup step checks whether the prompt already contains the same content inline and removes the duplicate. It tried to open each included file using `Path(file_path).is_file()`, which looks for the file relative to the current working directory. If the working directory was a subdirectory (like `tests/`), a path like `context/helper.py` couldn't be found, and the dedup step was silently skipped.

**Fix**: The code now detects the project root (using the same marker-walking logic as Bug 2) and resolves file paths relative to it instead of the current working directory.

### Bug 6: 

Processing of `insert_includes_LLM.prompt` printed warnings that `<insert>` tags could not be processed. The LLM prompt file text contained the raw text `<insert>` that not intended to be interpreted an actual tag. However, it was getting interpreted as a tag by the processor.

**Fix**: Escape `<include>` mentions in prose so the preprocessor doesn't match them as real tags.

### Bug 7: Symlinked directories caused cache misses and duplicate CSV entries

On platforms where the filesystem has symlinks in common paths (e.g. macOS where `/var` → `/private/var`), `summarize_directory` could produce different path forms for the same file depending on whether `os.path.normpath` or `os.path.realpath` was used. The cache lookup used `normpath` (which preserves symlinks), but `_compute_rel_path` used `realpath` (which resolves them). When a file was first scanned via one path form and then looked up via another, the cache missed and the file was re-summarized. The merge step had the same issue — it couldn't recognize that a scanned relative path and an existing absolute-path entry referred to the same file when symlinks were involved, causing duplicate entries in the CSV.

**Fix**: Cache lookup now checks three path forms: the normalized relative path, the normalized absolute path (`normpath`), and the resolved absolute path (`realpath`). The merge step's `scanned_paths` set also tracks the `realpath` form so that existing entries reachable via symlinks are correctly recognized as already present.

### Primary Fix Applied: Unified project root detection

Previously, each module resolved the project root differently: `summarize_directory` used `git rev-parse`, `include_query_extractor` and `insert_includes` read `get_config()["project_root"]` (which defaulted to CWD if no config existed), and `extracts_prune` did the same. This meant non-git projects had no stable root, and any module without a pdd config file fell back to CWD.

All modules now use `path_resolution.find_project_root_from_path`, which walks up from the given path looking for `.git`, `.pddrc`, `pyproject.toml`, `data/`, or `.env`.

This solution relies on detecting a "project root" (by looking for `.git`, `.pddrc`, `pyproject.toml`, etc.) and storing file paths relative to it. But what happens when none of those markers exist?

My approach in this PR is to store full absolute paths in the CSV (e.g., `/home/user/myapp/src/utils.py`). When a root marker is later detected (e.g., after `git init`), the existing absolute-path entries are still matched by the cache lookup (backward-compat logic already exists), and new entries are written as project-relative. After one scan, the CSV has fully transitioned — no re-summarization needed.

**Pros:**
- No collisions — absolute paths are always unique
- Seamless transition to project-relative paths after `git init` (cache hits, not misses)
- Scanning from any depth produces stable, unambiguous paths

**Cons:**
- Absolute paths leak into the LLM context and `<include>` tags for marker-free projects (functional but verbose)
- If the project directory is moved before adding a root marker, cached absolute paths and any saved `<include>` tags become stale (but adding any marker triggers a clean transition)

The second-best alternative I could think of was just storing file paths relative to whichever directory was scanned. So scanning `src/` stores `utils.py`, and scanning `lib/` stores `helpers.py`. That works well when you only ever scan one directory, and paths are short and readable, but it has worse cons, so I decided not to implement this:

**Cons:**
- `<include>` tags get bare filenames like `utils.py` which can't be resolved by the preprocessor unless CWD happens to be the same directory that was scanned
- If two different directories each contain a file with the same name (e.g., `src/utils.py` and `lib/utils.py`), both are stored as `utils.py` and one overwrites the other in the CSV — silent data loss
- When someone later does `git init`, the path format changes from directory-relative to project-relative, causing a one-time full re-summarization (all cache misses)

## Tests to verify the fixes

**New test files:**
- `test_e2e_auto_deps_pipeline.py` — Full pipeline integration test that runs `auto_deps_main` 4 times with real LLM calls across `context/` and `pdd/`, verifying CSV accumulation, path consistency, caching, and selective re-summarization on file modification. Catches compound interaction bugs across the summarize → auto_include → insert_includes chain that unit tests cannot catch — the bugs only manifest when CSV state flows between modules.
- `test_auto_deps_entry_wipe.py` — Unit tests for CSV wipe and cache-miss scenarios, no-marker absolute-path fallback (collision prevention, seamless git-init transition, abs+rel dedup), symlink cache consistency (`TestSymlinkCacheConsistency`), and a full pipeline test through `insert_includes` without markers. The alternating-scan test (`TestAlternatingDirectoryScans`) reproduces the exact production failure pattern observed in the git history.
- `test_path_resolution.py` — Tests for `find_project_root_from_path`: file vs directory input, each marker type, max_levels, no-marker fallback, nonexistent paths.

**Tests added to existing files:**
- `test_e2e_selective_includes.py` — End-to-end tests for the `auto_include` pipeline with mixed-origin CSV entries. Exercises the full pipeline (summarize → auto_include → insert_includes) with cross-directory CSV entries, catching path corruption that only surfaces in compound flows.
- `test_e2e_selective_includes_integration.py` — Real LLM quality checks (relevant files picked, paths valid, `select=` applied appropriately by file size).
- `test_summarize_directory.py` — Real LLM multi-directory CSV accumulation and cache hit verification.
- `test_auto_include.py` — Cross-directory path qualification, double-prefixing prevention, `_strip_selectors_from_small_files` CWD-independence, and CSV formatting error handling (malformed/missing columns degrade gracefully). Removed duplicate test block that was copy-pasted with inconsistent fixture values.
- `test_include_query_extractor.py` — Cache key normalization (absolute vs. relative paths).
- `test_insert_includes.py` — Dedup correctness when CWD differs from project root.

**Modified tests**:
- `test_extracts_prune.py` — Updated mocks and updated `test_absolute_keyed_entry` to reflect that absolute and relative paths now produce the same cache key.
- `test_e2e_selective_includes.py` — Updated mocks for `include_query_extractor` and `extracts_prune` call sites.
- `test_include_query_extractor.py` — Updated fixture and `test_absolute_vs_relative_path_same_cache_key` mocks.
- `test_insert_includes.py` — Updated dedup CWD-independence test mocks.

**Updated example files** (replaced stale `get_config` mocks with `find_project_root_from_path`):
- `include_query_extractor_example.py`
- `extracts_prune_example.py`

## Manual Testing

In addition to the written tests, I created a small local git repo to visually verify auto-deps end-to-end:

```
/tmp/auto_deps_test/
├── .git/
├── prompts/
│   └── prompt_budget_checker_python.prompt   # asks for a token-budget checker
└── src/
    ├── token_counter.py      # RELATED — tiktoken wrapper
    ├── color_palette.py      # unrelated — UI theming
    └── date_formatter.py     # unrelated — date helpers
```

I ran this auto-deps command to test the CWD-independence fix:

```bash
cd /tmp/auto_deps_test/src
pdd --verbose auto-deps ../prompts/prompt_budget_checker_python.prompt "../src/"
```

I visually verified that:
- The LLM added `src/token_counter.py` as an <include> tag dependency to the prompt file (ignored the two unrelated files)
- CSV paths are project-root-relative (e.g. `src/token_counter.py`), not absolute and not bare filenames — despite running from a subdirectory (Bug 2 fix)
- All 3 files are summarized in run 1
- Re-running auto-deps produced cache hits for all 3 files (Bug 2 fix)
- After modifying one file, only that file was re-summarized; the other two stayed cached

## Prompt files
Prompt files with all these changes will be added to pdd_cap after PR 71 on pdd_cap is merged. That PR adds the initial prompts for many of the files that are involved.


niti-go:
## Summary of findings and fixes

### Issue 1: CSV entry wipeout
**Root cause**: `summarize_directory()` builds `results_data` from scratch each run, only including discovered files. The final CSV is written exclusively from `results_data` — old entries not in the current scan are silently dropped.

**Real-world impact**: Scanning `context/` wipes `pdd/` entries and vice versa. The CSV bounced between ~269 and ~263 lines instead of staying at ~750+.

**Fix**: After processing discovered files, merge any `existing_data` entries whose paths aren't in `results_data`. Also fixed the early-return (no files found) path to preserve existing entries.

### Issue 2: Cache miss when `directory_path` changes
**Root cause**: Cache lookup only matches by relative path, but the relative path depends on `base_dir` which changes with `directory_path`. CSV has `cli.py` (from scanning `pdd/`), but scanning `.` produces `pdd/cli.py` — cache miss, unnecessary LLM call.

**Fix**: Added a secondary `existing_by_hash` index. When path-based lookup misses, fall back to content hash matching. Same content = reuse the summary.

### Issue 3: Narrow scan scope (`context/commands/` instead of `context/`)
Already fixed before us in `construct_paths.py` (commit `0ff8ecfb7`). Our fix for Issue 1 makes this class of bug harmless — even if a narrow scan happens, it no longer wipes unscanned entries.

### Tests written: 11

| Test | What it catches | Status |
|---|---|---|
| `test_entries_for_unscanned_files_are_preserved` | Scan dir_b, dir_a entries survive | Pass (fix 1) |
| `test_deleted_then_readded_file_not_resummarized` | Temp deletion wipes entry, forces rescan | Pass (fix 1) |
| `test_scanning_subset_preserves_full_csv` | Glob `*.py` wipes `.js` entries | Pass (fix 1) |
| `test_parent_dir_scan_reuses_subdir_cache` | CSV has `module.py`, scan produces `pkg/module.py` | Pass (fix 2) |
| `test_subdir_scan_reuses_parent_cache` | CSV has `pkg/module.py`, scan produces `module.py` | Pass (fix 2) |
| `test_directory_vs_glob_path_cache_consistency` | Dir path then glob pattern for same dir | Pass (already worked) |
| `test_trailing_slash_does_not_invalidate_cache` | `dir` vs `dir/` | Pass (already worked) |
| `test_old_format_csv_without_exports_forces_resummarization` | Old CSV format correctly triggers rescan | Pass (expected behavior) |
| `test_absolute_vs_relative_path_in_csv_cache_hit` | Abs path in CSV, rel path in scan | Pass (already worked) |
| `test_alternating_scans_accumulate_entries` | Scan context/ then pdd/ then context/ again | Pass (fix 1) |
| `test_partial_scan_preserves_unscanned_entries` | 5-entry CSV, scan finds 3 files, all 5 preserved | Pass (fix 1) |

niti-go:
## Follow-up: downstream path handling bugs surfaced by the entry preservation fix

The CSV entry preservation fix (issue #1) keeps entries from multiple directory scans in the same `project_dependencies.csv`. This is correct, but it surfaces latent bugs in downstream code that assumed all CSV entries came from a single `directory_path`.

### What I found

**`auto_include.py` — path qualification is broken for mixed-origin CSVs**

`_format_csv_rows_for_llm()` (line 84-87) blindly prepends `directory_path` to every relative path in the CSV. When the CSV has entries from both `pdd/` and `context/` scans:

- `cli.py` (from pdd/) becomes `context/cli.py` — a path that doesn't exist
- `context/example_a.py` (already qualified from a root scan) becomes `context/context/example_a.py`

`_qualify_result_paths()` (line 158-184) then double-qualifies paths that the LLM returns, since the LLM received already-qualified paths from `_format_csv_rows_for_llm`. The `startswith(prefix)` guard at line 153 doesn't help when `directory_path` is an absolute tmp_path like `/abs/path/context/` and the file is `context/helper.py`.

**`include_query_extractor.py` — cache keys aren't stable across invocations**

`compute_cache_key()` (line 75-82) uses `os.path.normpath()` which does NOT resolve to absolute paths. The same file accessed as `src/file.py` vs `/abs/path/src/file.py` (or from different CWDs as `../src/file.py` vs `src/file.py`) produces different cache keys, causing cache misses.

**`insert_includes.py` — dedup silently fails for non-CWD paths**

`_remove_redundant_content()` (line 48-50) uses `Path(file_path).is_file()` which resolves relative to CWD. If `auto_include` returned paths relative to a different base directory, the file can't be found and dedup is silently skipped.

### Tests added

**`test_auto_include.py`** — 3 new failing tests:

| Test | Bug confirmed |
|---|---|
| `test_cross_directory_entries_not_misqualified` | `cli.py` from pdd/ wrongly becomes `context/cli.py` |
| `test_already_qualified_paths_not_double_qualified` | `context/example_a.py` becomes `context/context/example_a.py` |
| `test_new_include_already_qualified_not_doubled` | `_qualify_result_paths` double-qualifies after `_format_csv_rows_for_llm` |

Plus 2 passing tests documenting `_qualify_path` behavior.

**`test_include_query_extractor.py`** — 3 new passing tests documenting the cache key limitation (relative vs absolute paths produce different keys for the same file).

**`test_insert_includes.py`** — 2 new passing tests documenting that dedup silently fails when include paths aren't resolvable from CWD.

### Why these bugs matter now

Previously these bugs were invisible because the CSV wipeout meant the CSV only ever had entries from one directory at a time. Now that we preserve cross-directory entries, `_format_csv_rows_for_llm` will misqualify them. The root cause across all these files is the same: **the CSV stores paths relative to a variable `base_dir` with no record of what that `base_dir` was.** A proper fix would store paths relative to the git repo root (or as absolute paths).

gltanaka (changes requested):
Thanks for digging into this. I was able to confirm that this PR addresses several real auto-deps problems on `main`, and I appreciate how much testing you added around the behavior.

I do have one remaining blocking issue before this is ready to merge:

In repos with nested `.pddrc` files, the path normalization still is not stable across scan roots. `find_project_root_from_path()` returns the nearest marker, so `summarize_directory()` can still store the same file under different relative paths depending on where the scan starts.

I reproduced it with a temp repo shaped like this:
- repo root has `.git`
- `module/.pddrc`
- `module/context/a.py`

Then:
1. Scan `module/context` -> CSV stores `context/a.py`
2. Scan `.` with that CSV as input -> CSV stores `module/context/a.py`
3. The file is re-summarized and both entries can remain in the CSV

So the cache/path-stability issue this PR is trying to solve still exists for nested PDD project layouts.

If you can normalize to one consistent project root across scans for that case, I think this will be in much better shape. The overall direction here is valuable, and I’d be happy to take another look after that fix. Thanks again for the careful work on this.

niti-go:
Thank you for the review. I have implemented the following new approach:

`.git` is most likely to indicate project/repo scope compared to other markers. This should be prioritized as the main project root marker. 

Other markers (.pddrc, data/, .env, pyproject.toml) are still valuable, but are weaker because there could be multiple nested inside a single project/repo. We should only consider these if we don't see any `.git` file. So, this is the new walk logic for determining the project root from a file:

Start at the file/directory you are determining the project root of.
Walk upward up to `max_levels` (default: 10).
At each level:  
- If directory has `.git` → return it immediately as the project root. Done.
- If directory has a weak marker → remember it as a candidate, keep walking.
- If the walk ends without finding a `.git` file → return the highest weak candidate seen (outermost, not nearest).
- If no marker at all → return None (existing absolute-path fallback kicks in).

There are some downsides to this approach.
- This is error-prone for projects that are 10+ levels deep. 
- Projects that don't use `.git` will always pay the efficiency cost of searching through 10 levels, for every file scan.
- For projects with git submodules, if users scan inside a submodule, the project root would resolve to the submodule root, not the project root.
- If a project initially does not use git and has multiple nested .pddrc markers, then the project root initially gets resolved to one of the nested directories. Later, if someone runs `git init` higher up in the true project root, this function now returns a new root and CSV keys now have the wrong prefix.
- A user might initially run auto-deps on a project that has some root markers (e.g. pyproject.toml). This generates the CSV with relative paths. Then, they delete the markers. Now, they re-run auto-deps. Auto-deps rescans all files and adds them to the CSV with absolute paths, essentially duplicating all entries. 

However, this is the best approach I could come up with with the least error cases. I'm not sure if there's any other clean way to determine a consistent project root when there are no indicators like .git, while also avoiding using absolute paths in case the project later *does* become a git project. 

Another alternative I considered is storing the project root path in the CSV itself. However, it would have to be stored as an absolute path, which would break across machines. I also want to avoid storing any absolute paths in the CSV, because then <include> tags would have absolute paths, and later transitions to git would be difficult.

gltanaka:
target 4/14

pdd-cli (approved):
This looks great, Niti. The tests are comprehensive, and standardizing the project root detection was definitely the right call for fixing these path resolution issues.

gltanaka (approved):
Revalidated this on a clean cherry-pick branch and through Cloud Batch before merging locally. I also reproduced the original cache/path issues on baseline main first and verified they were fixed after applying these changes.

Thanks for the contribution. This looks good to merge now.

gltanaka:
Validation summary from April 15, 2026:

I validated this in three layers before merging.

1. Baseline repro on pre-fix main
- Reproduced CSV entry wipeout: after scanning one directory and then another, the second scan overwrote prior CSV rows instead of preserving them.
- Reproduced scan-root cache miss: the same file was re-summarized when the scan root changed because paths were not stable.
- Reproduced incorrect path qualification in auto-includes: `pdd/cli.py` could become `context/pdd/cli.py`.
- Reproduced dedup failure from a non-root working directory.
- Reproduced nested `.git` + nested `.pddrc` path instability, where the same source file could land under different CSV keys depending on scan root.

2. Cherry-pick validation on a clean branch from `gltanaka/pdd` main
- Created a clean branch and worktree, then cherry-picked the PR commits there.
- Re-ran the same manual repros after applying the changes.
- Confirmed CSV entries were preserved across scans.
- Confirmed changing scan roots no longer caused the same file to be re-summarized.
- Confirmed `pdd/cli.py` stayed `pdd/cli.py` instead of being incorrectly prefixed.
- Confirmed dedup worked from a non-root CWD.
- Confirmed the nested `.git` + `.pddrc` case was stable for the actual source file path, and nested `.pddrc` without `.git` was also stable.

3. Automated validation
- Focused pytest validation on the cherry-pick branch:
  `env -u PDD_PATH pytest tests/test_auto_deps_entry_wipe.py tests/test_include_query_extractor.py tests/test_insert_includes.py tests/test_path_resolution.py tests/test_e2e_auto_deps_pipeline.py -q`
  Result: `125 passed, 1 skipped`
- Full Cloud Batch validation from the cherry-pick worktree using `make cloud-test`:
  Result: `77 passed, 0 failed, 0 errors`

After that validation, I merged the equivalent cherry-pick branch into `gltanaka/pdd` main and then merged this PR here. Thanks again for the contribution.

PR #786, opened 2026-04-22
Fix pdd example postprocessing and path resolution

gltanaka:
## Summary
- pass the requested language through postprocess() so non-Python examples are extracted with the correct structured-output settings
- prefer existing basename-derived example/test artifacts when architecture.json points to a nested filepath stem like lib/sse.ts
- add regression coverage for both behaviors

## Testing
- python3 -m pytest tests/test_postprocess.py tests/test_issue_225_paths_and_includes.py -q

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Fix
+
+- forward `language` from `postprocess()` into `llm_invoke()` so non-Python example extraction no longer trips Python-syntax cache-bypass retries
+- when `architecture.json` points to a nested filepath stem (for example `lib/sse.py`) but basename-derived artifacts already exist (`lib_sse_example.py`, `test_lib_sse.py`), prefer the existing flattened artifacts and log when that override activates
+
 ## v0.0.215 (2026-04-21)
 
 ### Fix
--- a/pdd/postprocess.py
+++ b/pdd/postprocess.py
@@ -121,6 +121,7 @@ def postprocess(
             strength=strength,
             temperature=temperature,
             time=time,
+            language=language,
             output_pydantic=ExtractedCode,
             verbose=verbose,
         )
--- a/pdd/sync_determine_operation.py
+++ b/pdd/sync_determine_operation.py
@@ -947,8 +947,20 @@ def get_pdd_file_paths(basename: str, language: str, prompts_dir: str = "prompts
                 example_path = project_root / f"{example_dir}{code_stem}_example.{extension}"
                 test_path = project_root / f"{test_dir}test_{code_stem}.{extension}"
 
+                # If the flattened prompt basename already has corresponding example/test
+                # artifacts, prefer those over the architecture filepath stem. This keeps
+                # command summaries and sync behavior aligned with repos that intentionally
+                # namespace files such as lib_sse_example.ts or test_api_route.ts.
+                if name != code_stem:
+                    basename_example_path = project_root / f"{example_dir}{name}_example.{extension}"
+                    basename_test_path = project_root / f"{test_dir}test_{name}.{extension}"
+                    if basename_example_path.exists():
+                        example_path = basename_example_path
+                    if basename_test_path.exists():
+                        test_path = basename_test_path
+
                 test_dir_path = test_path.parent
-                test_stem = f"test_{code_stem}"
+                test_stem = glob.escape(test_path.stem)
                 if test_dir_path.exists():
                     matching_test_files = sorted(test_dir_path.glob(f"{test_stem}*.{extension}"))
                 else:
--- a/pdd/sync_determine_operation.py
+++ b/pdd/sync_determine_operation.py
@@ -954,10 +954,23 @@ def get_pdd_file_paths(basename: str, language: str, prompts_dir: str = "prompts
                 if name != code_stem:
                     basename_example_path = project_root / f"{example_dir}{name}_example.{extension}"
                     basename_test_path = project_root / f"{test_dir}test_{name}.{extension}"
+                    preferred_example = False
+                    preferred_test = False
                     if basename_example_path.exists():
                         example_path = basename_example_path
+                        preferred_example = True
                     if basename_test_path.exists():
                         test_path = basename_test_path
+                        preferred_test = True
+                    if preferred_example or preferred_test:
+                        logger.info(
+                            "Preferring basename-derived artifacts for %s over architecture stem %s "
+                            "(example=%s, test=%s)",
+                            name,
+                            code_stem,
+                            preferred_example,
+                            preferred_test,
+                        )
 
                 test_dir_path = test_path.parent
                 test_stem = glob.escape(test_path.stem)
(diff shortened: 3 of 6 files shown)


gltanaka:
Follow-up summary with reproduction details, decision analysis, and validation.

## What I reproduced before opening the PR

I reproduced both issues against a real downstream repo (`Generative-Video-Studio`) using the local `pdd` CLI workflow.

### 1. `pdd example` was validating TypeScript as Python during extraction

Repro path:
- `pdd --local example prompts/lib_sse_TypeScript.prompt lib/sse.ts --output examples/lib_sse_example.generated.ts`

Observed behavior on the unpatched CLI:
- example generation completed, but postprocessing logged:
  - `[WARNING] Detected invalid Python syntax in code fields ... Retrying with cache bypass...`
- it then attempted the cache-bypass retry path, which is the wrong behavior for valid TypeScript and adds unnecessary token/cost churn
- on an earlier `/tmp`-targeted run it also fell through into an unnecessary Gemini retry path that produced a 404

Root cause:
- `postprocess()` called `llm_invoke(..., output_pydantic=ExtractedCode, ...)` but did not pass `language`
- in `llm_invoke`, the invalid-code retry guard treats `language in (None, "python")` as Python-validation mode
- so non-Python extracted code in structured-output fields could hit `_has_invalid_python_code()` and trigger a false-positive cache-bypass retry

### 2. `get_pdd_file_paths()` resolved the wrong example/test artifact names when `architecture.json` used a nested filepath stem

Repro shape:
- prompt basename: `lib_sse`
- architecture filepath: `lib/sse.ts`
- real repo artifacts: `examples/lib_sse_example.ts`, `tests/test_lib_sse.ts`

Observed behavior on the unpatched CLI:
- command summary / path resolution preferred:
  - `examples/sse_example.ts`
  - `tests/test_sse.ts`
- that mismatched the repo’s actual flattened artifact naming

Root cause:
- path resolution derived `code_stem = "sse"` from `architecture.json` and always built example/test names from that stem
- it did not check whether the repo already had basename-derived artifacts matching the prompt basename (`lib_sse`)

## Why I still think both changes belong in the same PR

I agree with the review framing: once “multi-purpose PR” stops being the objection, the real question is expected value vs expected cost per change.

### Change 1: language passthrough in `postprocess()`

Why it is worth merging:
- behavior change is extremely narrow: it only affects structured extraction when `language` is something other than `None` / `python`
- for non-Python languages, skipping Python AST validation is the correct behavior
- there is real recurring cost in not merging it: every false-positive retry on extracted TypeScript/JS/Go/Rust/etc. burns tokens and adds latency for no benefit

Risk assessment:
- mechanically small change: a single forwarded kwarg
- failure mode is low-risk and very easy to reason about
- original forwarding-only unit test was a decent floor; I strengthened coverage after review (details below)

### Change 2: prefer existing basename-derived artifacts when they already exist

Why it is worth merging:
- activation is narrow: it only applies when both of these are true:
  1. `name != code_stem`
  2. the basename-derived example/test files already exist on disk
- that is not a broad semantic rewrite; it is a targeted fallback for a specific, already-observed repo shape
- `get_pdd_file_paths()` is already filesystem-state-dependent throughout; this change is consistent with the existing design rather than introducing a new kind of dependency on disk state

Risk assessment:
- worst realistic failure mode is that a repo with both naming schemes present prefers the basename-derived file if it exists
- that is recoverable and visible; it is not a destructive or silent-data-loss class failure
- the paired `glob.escape(test_path.stem)` change is part of making that override glob correctly, not an unrelated semantic change

## Review follow-ups I implemented after the PR was opened

Based on the review feedback, I added the following follow-up improvements on top of the original two-file fix:

### 1. Stronger regression coverage for the non-Python retry bug

Added in `tests/test_llm_invoke.py`:
- `test_llm_invoke_skips_python_validation_retry_for_typescript_language`

What it proves:
- valid TypeScript in a structured-output code field does **not** trigger the invalid-Python cache-bypass retry path when `language="typescript"`
- retry count stays at 1 completion call
- no retry-warning text appears in logs

This is stronger than the original forwarding-only `postprocess()` assertion because it exercises the actual retry guard behavior at the `llm_invoke` layer.

### 2. Debuggability log for the basename-artifact override

Added in `pdd/sync_determine_operation.py`:
- an `INFO` log when basename-derived artifacts are preferred over the architecture filepath stem

Example log shape:
- `Preferring basename-derived artifacts for lib_sse over architecture stem sse (example=True, test=True)`

This makes the filesystem-dependent override visible when someone asks why sync chose one artifact path over another.

### 3. Test coverage for the new log behavior

Extended `tests/test_issue_225_paths_and_includes.py`:
- `test_get_pdd_file_paths_prefers_existing_basename_artifacts_with_architecture_filepath` now also asserts that the override log is emitted

### 4. Changelog note

Added an `Unreleased` note in `CHANGELOG.md` for:
- the `postprocess()` language passthrough fix
- the basename-derived artifact preference behavior

## Follow-up analysis I did but did not turn into code changes

I also audited the other current `postprocess(...)` callers mentioned in review:
- `pdd/code_generator.py`
- `pdd/context_generator.py`
- `pdd/generate_test.py`
- `pdd/increase_tests.py`
- `pdd/bug_to_unit_test.py`

Those call sites already pass `language`, so I did **not** add extra code changes there.

## Validation

### Downstream repro validation

Using the patched local source CLI (`PYTHONPATH=/Users/gregtanaka/Documents/pdd_cloud/pdd python3 -m pdd.cli --local ...`) against `Generative-Video-Studio`:

- `pdd example` for `lib/sse.ts` no longer emitted the false invalid-Python warning for TypeScript extraction
- the command summary resolved the real repo artifact names correctly:
  - `examples/lib_sse_example.ts`
  - `tests/test_lib_sse.ts`
- the generated example ran successfully from the repo-local output path

### Local test validation for the PR

Original PR validation:
- `python3 -m pytest tests/test_postprocess.py -q` -> `45 passed`
- `python3 -m pytest tests/test_issue_225_paths_and_includes.py -q` -> `7 passed`

After the follow-up review changes:
- `python3 -m pytest tests/test_postprocess.py -q` -> `45 passed`
- `python3 -m pytest tests/test_issue_225_paths_and_includes.py -q` -> `7 passed`
- `python3 -m pytest tests/test_llm_invoke.py -k "test_llm_invoke_retries_on_invalid_python_code or test_llm_invoke_skips_python_validation_retry_for_typescript_language or test_llm_invoke_uses_repaired_code_without_retry or test_llm_invoke_no_retry_in_batch_mode" -q` -> `4 passed`

I also ran the full `tests/test_llm_invoke.py -q` once while validating. That file still has one unrelated pre-existing failure:
- `test_default_base_model_can_be_none`

That failure is unrelated to this PR’s changes, so I did not fold it into this fix.

## Final state of the PR

The PR now contains two commits:
1. `4b171a572` - `Fix pdd example postprocessing and path resolution`
2. `a6c8cd24a` - `Add follow-up coverage and logging for CLI fixes`

I think this lands the right balance:
- the original fixes remain small and targeted
- the follow-up changes improve observability and strengthen the most important regression path
- no unrelated cleanup was pulled into the branch

gltanaka:
Superseded by gltanaka/pdd#1284, which is opened on the source-of-truth repository.

PR #835, opened 2026-05-05
fix: preserve path-qualified sync dependencies

### LINKED ISSUE #834: pdd sync loses architecture dependency order for path-qualified branch-diff modules
## Problem

Cloud `pdd-sync` can detect changed prompt modules from a branch diff as path-qualified targets such as `src/config`, `src/models`, and `src/clients/firestore_client`. Agentic sync needs those path-qualified names to route each child `pdd sync` through the correct `.pddrc` context.

However, `build_dep_graph_from_architecture_data()` only matched architecture dependencies through `filename`-derived basenames such as `config` and `firestore_client`. In repositories whose architecture entries store bare prompt filenames plus a `filepath`, the path-qualified target set does not match the architecture dependency endpoints.

The result is an empty dependency graph, so large dependent modules can start in parallel instead of waiting for foundational modules.

## Observed failure

`promptdriven/pdd_cloud#1340` hit this shape after `pdd change` created PR `promptdriven/pdd_cloud#1345`.

`pdd-sync` detected five modules:

- `src/config`
- `src/models`
- `src/clients/firestore_client`
- `src/services/solving_orchestrator`
- `src/worker_app`

The small modules succeeded, while the larger dependent modules all hit the per-module timeout together:

- `src/clients/firestore_client`: `Timeout after 1800s`
- `src/services/solving_orchestrator`: `Timeout after 1800s`
- `src/worker_app`: `Timeout after 1800s`

Local graph reproduction before the fix returned all-empty deps for those exact targets.

## Expected behavior

When architecture entries include `filepath`, agentic sync should match targets against both identities:

- prompt filename identity: `config_Python.prompt` -> `config`
- code filepath identity: `src/config.py` -> `src/config`

The graph should preserve the original path-qualified target strings while restoring dependency edges.

## Acceptance criteria

- Path-qualified branch-diff targets match architecture entries by `filepath` as well as `filename`.
- The dependency graph preserves the original target strings used by the runner.
- Regression test covers the pdd_cloud module shape from issue #1340.
- Existing agentic sync tests continue to pass.

Serhan-Asad:
## Summary

Fixes #834.

Agentic `pdd sync` can receive branch-diff modules as path-qualified targets such as `src/config` and `src/clients/firestore_client`. Those names must be preserved for `.pddrc` context routing, but the architecture dependency graph previously matched only prompt `filename` basenames such as `config` and `firestore_client`.

This caused architecture dependencies to be dropped for repos whose entries use bare prompt filenames plus `filepath`, which let large dependent modules start in parallel and hit cloud timeouts. This matches the observed `promptdriven/pdd_cloud#1340` / PR `promptdriven/pdd_cloud#1345` failure shape.

Related downstream mirror issue/PR: `gltanaka/pdd#1382`, `gltanaka/pdd#1383`.

## Changes

- Match architecture entries by both prompt `filename` aliases and code `filepath` aliases.
- Preserve the original requested sync target strings as graph keys and dependency values.
- Update the generated prompt contract for `agentic_sync_runner`.
- Add a regression test using the pdd_cloud module shape: `src/config`, `src/models`, `src/clients/firestore_client`, `src/services/solving_orchestrator`, `src/worker_app`.

## Verification

- `PYTHONPATH=$PWD python -m pytest tests/test_agentic_sync.py tests/test_agentic_sync_runner.py -q` -> 228 passed
- `PYTHONPATH=$PWD python -m pytest tests/test_agentic_sync.py::TestBuildDepGraphFromArchitecture::test_preserves_dependencies_for_path_qualified_targets -q` -> 1 passed
- `PYTHONPATH=$PWD python -m ruff check pdd/agentic_sync_runner.py` -> passed
- `git diff --check` -> passed

Note: `ruff check tests/test_agentic_sync.py` still reports pre-existing lint issues in that file; this PR does not touch those unrelated cleanups.

### DIFF
--- a/pdd/agentic_sync_runner.py
+++ b/pdd/agentic_sync_runner.py
@@ -116,6 +116,49 @@ def _basename_from_architecture_filename(filename: str) -> Optional[str]:
     return str(path.parent / basename)
 
 
+def _basename_from_architecture_filepath(filepath: str) -> Optional[str]:
+    """Return a sync basename from an architecture filepath."""
+    if not filepath:
+        return None
+
+    path = Path(filepath)
+    if path.suffix:
+        path = path.with_suffix("")
+    return path.as_posix()
+
+
+def _architecture_entry_aliases(entry: Dict[str, Any]) -> set[str]:
+    """Return all sync basenames that may identify an architecture entry."""
+    aliases: set[str] = set()
+
+    filename_basename = _basename_from_architecture_filename(
+        str(entry.get("filename", "") or "")
+    )
+    if filename_basename:
+        aliases.add(filename_basename)
+
+    filepath_basename = _basename_from_architecture_filepath(
+        str(entry.get("filepath", "") or "")
+    )
+    if filepath_basename:
+        aliases.add(filepath_basename)
+
+    return aliases
+
+
+def _target_basename_aliases(target_basename: str) -> set[str]:
+    """Return aliases that should match a requested sync target."""
+    aliases = {target_basename}
+    path = Path(target_basename)
+    stripped_name = extract_module_from_include(path.name + ".prompt")
+    if stripped_name:
+        if path.parent == Path("."):
+            aliases.add(stripped_name)
+        else:
+            aliases.add((path.parent / stripped_name).as_posix())
+    return aliases
+
+
 def build_dep_graph_from_architecture(
     arch_path: Path, target_basenames: List[str]
 ) -> DepGraphFromArchitectureResult:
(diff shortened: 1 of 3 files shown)


Serhan-Asad:
Moved to the private repository because we do not use the public pdd repo for this work. Follow-up PR: gltanaka/pdd#1386. Tracking issue: gltanaka/pdd#1382. Closing this public PR.

Serhan-Asad:
Closed in favor of private PR gltanaka/pdd#1386.

PR #846, opened 2026-05-07
fix(#813): preserve Claude OAuth when stale Anthropic key is set

### LINKED ISSUE #813: agentic_common: subprocess CI=1 + inherited ANTHROPIC_API_KEY silently bypasses Claude Code Max OAuth
## Summary

When `agentic_common._run_provider_cli` spawns the `claude` CLI for any agentic-architecture step, it sets `CI=1` in the subprocess env while forwarding the rest of the user's env unchanged. If the user has a stale or depleted `ANTHROPIC_API_KEY` exported in their shell (a common situation — many third-party tools and historical PDD docs prompt users to set it), the `CI=1` flag causes Claude Code to take its non-interactive auth path and prefer the API key over the user's stored Max OAuth credential. The result: every "anthropic" provider call inside pdd fails with HTTP 400 "Credit balance is too low" (or similar), even though running `claude` interactively in the same shell uses Max successfully.

## Repro

Shell with a stale/depleted `ANTHROPIC_API_KEY` exported (e.g. left over in `~/.bashrc` or a conda activation hook) and Claude Code logged in via Max:

```
$ claude /status
   Login method:     Claude Max account

$ CI=1 claude --print --output-format json "say hi" </dev/null
{"type":"result","subtype":"success","is_error":true,"api_error_status":400,
 "duration_ms":253,"duration_api_ms":0,"num_turns":1,
 "result":"Credit balance is too low",...}
```

`duration_api_ms: 0` indicates the failure happens during auth resolution before the request leaves — confirming Claude Code switched to API-key mode under `CI=1`. Removing `CI=1` (or unsetting `ANTHROPIC_API_KEY`) makes the same invocation succeed via Max.

## Impact

Whole-run failure of `pdd generate` (and any other agentic command that delegates to `claude`) for users who keep `ANTHROPIC_API_KEY` in their shell env even when their *intent* is to use the Claude Max subscription via Claude Code. The failure mode is opaque because:

- Interactive `claude /status` shows "Claude Max account" — pointing diagnostics in the wrong direction.
- pdd's progress messages show each step "completing" because the orchestrator silently falls back to a different provider (see related issue on log visibility).

## Code reference

`pdd/agentic_common.py` `_run_provider_cli`, env construction (in installed `pdd-cli==0.0.227`):

```python
env = os.environ.copy()
env["TERM"] = "dumb"
env["NO_COLOR"] = "1"
env["CI"] = "1"
env.pop("PDD_OUTPUT_COST_PATH", None)
env["GIT_WORK_TREE"] = str(cwd)
```

Grep across the package shows no `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` mutation anywhere — the depleted key is purely inherited from the parent shell.

## Suggested fix

When `provider == "anthropic"` and pdd is delegating to the `claude` CLI (which has its own credential store), pop the relevant Anthropic auth env vars from the subprocess env so that Claude Code uses whatever credential it would use interactively:

```python
if provider == "anthropic":
    env.pop("ANTHROPIC_API_KEY", None)
    env.pop("ANTHROPIC_AUTH_TOKEN", None)
```

Alternatively (or additionally), detect the conflict (`CI=1` + `ANTHROPIC_API_KEY` set + `claude /status` reports Max OAuth) and emit a loud warning before starting the run. The same consideration likely applies to the openai/google branches when delegating to their respective CLIs.

## Environment

- `pdd-cli==0.0.227` (Linux, conda env)
- Claude Code 2.1.126
- Behavior reproduced with the trimmed one-liner above; original failure surfaced via `pdd generate <issue-url>`.

---

**Related:**
- #814 — once the auth issue triggers a 400, "Credit balance is too low" is treated as transient and retried, amplifying the failure window
- #816 — the silent fallback to a different provider is invisible in `.pdd/agentic-logs/`, which hid this root cause during diagnosis

Serhan-Asad:
## Summary
- Fixes #813 by stripping stale `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` only for Claude subprocesses when first-party Claude OAuth is confirmed.
- Preserves API-key-only setups by treating failed/missing/non-JSON auth probes as "no OAuth" and leaving the key in place.
- Updates setup/CLI detection/docs so OAuth-backed Claude/Gemini/Codex users are not pushed into adding an API key for agentic workflows, while documenting which direct prompt paths still require LiteLLM API keys.
- Adds pytest isolation for cwd and heavily mocked orchestrator globals so the public unit workflow is not order-dependent under xdist.

## Review Note Addressed
- Claude Code 2.1.38 rejects `claude auth status --json`; the probe now tries the documented `claude auth status` command first and falls back once to `claude auth status --json` for older builds.
- Added regression coverage for both the documented-command-first path and the legacy `--json` fallback.

## Regression Guards
- API-key-only Anthropic flows keep `ANTHROPIC_API_KEY` intact when no OAuth login is detected.
- Bedrock/Vertex routes bypass the strip.
- `PDD_KEEP_ANTHROPIC_API_KEY=1` forces API-key billing.
- Codex/Gemini credential handling is not scrubbed speculatively; only documented/detected auth availability is used for setup UX.
- OAuth-only setup summary does not recommend direct LiteLLM commands that still need API keys.

## Verification
- GitHub Actions on final head `353983819`: CodeQL, Analyze (actions), Analyze (javascript-typescript), Analyze (python), Public CLI Regression, and Run Unit Tests all passed.
- `conda run -n pdd python -m pytest tests/test_agentic_e2e_fix_orchestrator.py tests/test_e2e_issue_468_not_a_bug_early_exit.py tests/test_e2e_issue_903_convergence.py -q --tb=short --timeout=60` — 306 passed.
- `conda run -n pdd python -m pytest tests/test_agentic_common.py tests/test_agentic_common_worktree.py tests/test_agentic_common_issue_813_anthropic_api_key_oauth_shadow.py tests/test_cli_detector.py tests/test_setup_tool.py tests/test_e2e_issue_468_not_a_bug_early_exit.py tests/test_e2e_issue_903_convergence.py tests/test_generate_model_catalog.py -q` — 475 passed.
- `conda run -n pdd python -m pytest tests/test_agentic_common_issue_813_anthropic_api_key_oauth_shadow.py tests/test_e2e_issue_468_not_a_bug_early_exit.py tests/test_e2e_issue_903_convergence.py -q` — 55 passed.
- `conda run -n pdd python -m compileall -q pdd/agentic_common.py pdd/cli_detector.py pdd/setup_tool.py`.
- `git diff --check`.
- `conda run -n pdd pdd checkup --validate-arch-includes --project-root tests/fixtures/arch_include_validate_ok`.
- `conda run -n pdd pdd checkup --validate-arch-includes`.
- `conda run -n pdd env CI=true make regression-public`.

Fixes #813


gltanaka (approved):
I think this is mergeable now. The prior blocker has been addressed for supported/current Claude Code builds: `_probe_claude_auth_status()` now tries the documented `claude auth status` command first and falls back to `claude auth status --json`, with regression coverage for both paths.

I re-ran the focused local checks on this branch:

- `pytest -q tests/test_agentic_common_issue_813_anthropic_api_key_oauth_shadow.py tests/test_agentic_common.py::test_environment_sanitization tests/test_agentic_common.py::test_anthropic_api_key_preserved_when_explicitly_set` — 49 passed
- `pytest -q tests/test_cli_detector.py tests/test_setup_tool.py` — 89 passed

GitHub checks are green as well. The remaining caveat is only for older Claude Code versions that do not support `claude auth status` at all; the probe intentionally fails open there to preserve API-key-only setups. Given current docs and current/stable Claude Code support the auth-status command, I do not consider that a merge blocker.

Issue #798, opened 2026-04-30
Support OpenCode as an agentic CLI backend


gltanaka:
## Summary

Add OpenCode CLI support as an agentic backend for PDD workflows that currently use Claude Code, Gemini CLI, and Codex CLI.

This should cover agentic workflows such as `pdd change`, `pdd bug`, `pdd fix`, `pdd crash`, and `pdd update`, where PDD delegates repository-aware work to an external coding agent.

## Motivation

OpenCode is an open-source, provider-agnostic coding agent with a CLI (`opencode run`) and a headless server/API (`opencode serve`). Supporting it would give PDD users another agentic backend, especially for teams already using OpenCode provider configuration, OAuth flows, or GitHub Copilot/OpenRouter/local model setups through OpenCode.

There is an older closed umbrella issue for agentic coder CLI support (#61), but no open issue dedicated to OpenCode integration.

## Proposed approach

Implement OpenCode as a first-class agentic provider in PDD, probably named `opencode` in `PDD_AGENTIC_PROVIDER`.

Core changes:

- Add `opencode` to agentic provider detection and preference handling in `pdd/agentic_common.py`.
- Add `opencode` to setup/detection/install guidance in `pdd/cli_detector.py` and setup summaries.
- Add README/setup docs for installing OpenCode and configuring a model/provider.
- Add tests for detection, command construction/API invocation, output parsing, cost parsing, and provider preference ordering.

## Model selection and `llm_model.csv`

OpenCode support should use PDD's model catalog instead of hardcoding a default model.

Expected behavior:

1. Prefer explicit `OPENCODE_MODEL` when set.
2. If `OPENCODE_MODEL` is unset, derive a candidate from `llm_model.csv` using PDD's existing strength/model-selection semantics where possible.
3. Translate LiteLLM-oriented CSV model IDs to OpenCode provider/model IDs where they differ.
   - Example: PDD CSV uses `github_copilot/...`, while OpenCode uses `github-copilot/...`.
   - OpenCode expects `provider/model`, and model IDs may themselves contain slashes, e.g. `openrouter/openai/gpt-5.3-codex`.
4. Validate the selected model against OpenCode's configured/available providers when using the server API or `opencode models`.
5. Use OpenCode-reported cost from JSON events when available; fall back to CSV pricing only when OpenCode does not report cost.

Suggested env vars:

- `OPENCODE_MODEL=provider/model`
- `OPENCODE_AGENT=build` or another OpenCode agent name
[...]

PR #848, opened 2026-05-07
Fix generated prompt include sanitization

gltanaka:
## Summary
- Validate attributed and self-closing prompt `<include>` tags, including selected includes backed by `ContentSelector`.
- Sanitize generated `.prompt` output before `change`, `update`, `split`, and `auto-deps` persist prompt files.
- Update prompt specs and add regressions for the issue #798 bad-selector shape.

## Root Cause
`pdd issue` produced a prompt change that selected `TokenMatch` symbols from `context/llm_invoke_example.py`, where those symbols do not exist. The existing validator only handled bare `<include>path</include>` tags and only checked file existence, so attributed selected includes were persisted and later caused `pdd sync`/preprocess to fall back to full-file context with a `ContentSelector failed` warning.

## Test Plan
- `conda run -n pdd --no-capture-output env PDD_PATH=$PWD PYTHONPATH=$PWD python -m pytest tests/test_issue_225_paths_and_includes.py tests/test_change_main.py tests/test_update_main.py tests/test_split_main.py tests/test_auto_deps_main.py tests/test_preprocess.py tests/test_code_generator_main.py tests/test_agentic_sync_runner.py tests/test_auto_include.py tests/test_insert_includes.py -q`
- `conda run -n pdd --no-capture-output env PYTHONPATH=$PWD python -m py_compile pdd/validate_prompt_includes.py pdd/change_main.py pdd/update_main.py pdd/split_main.py pdd/auto_deps_main.py`
- Focused issue #798 selector reproduction using `origin/change/issue-798`: validator catches the bad selected include, and preprocess emits no `ContentSelector failed` warning after sanitization.
- `git diff --check`

### DIFF
--- a/pdd/auto_deps_main.py
+++ b/pdd/auto_deps_main.py
@@ -9,6 +9,7 @@ from rich import print as rprint
 from . import DEFAULT_STRENGTH, DEFAULT_TIME
 from .construct_paths import construct_paths
 from .insert_includes import insert_includes
+from .validate_prompt_includes import sanitize_prompt_output
 
 
 def auto_deps_main(
@@ -107,6 +108,15 @@ def auto_deps_main(
             # Save the modified prompt
             output_path = output_file_paths["output"]
             if output_path:
+                modified_prompt, invalid_includes = sanitize_prompt_output(
+                    modified_prompt,
+                    output_path,
+                )
+                if invalid_includes and not ctx.obj.get("quiet", False):
+                    rprint(
+                        "[yellow]Warning: Cleaned invalid <include> tag(s) "
+                        f"before saving {output_path}.[/yellow]"
+                    )
                 with open(output_path, 'w', encoding='utf-8') as f:
                     f.write(modified_prompt)
                 try:
--- a/pdd/change_main.py
+++ b/pdd/change_main.py
@@ -23,6 +23,7 @@ from .change import change as change_func
 from .process_csv_change import process_csv_change
 from .get_extension import get_extension
 from .user_story_tests import run_user_story_tests, discover_prompt_files
+from .validate_prompt_includes import sanitize_prompt_output
 
 # Set up logging
 logger = logging.getLogger(__name__)
@@ -433,6 +434,15 @@ def change_main(
 
                         try:
                             logger.debug("Attempting to save file to: %s", individual_output_path)
+                            modified_content, invalid_includes = sanitize_prompt_output(
+                                modified_content,
+                                individual_output_path,
+                            )
+                            if invalid_includes and not quiet:
+                                rprint(
+                                    "[yellow]Warning: Cleaned invalid <include> tag(s) "
+                                    f"before saving {individual_output_path}.[/yellow]"
+                                )
                             with open(individual_output_path, 'w', encoding='utf-8') as output_file:
                                 output_file.write(modified_content)
                             logger.debug("Saved modified prompt to: %s", individual_output_path)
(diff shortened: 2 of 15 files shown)


gltanaka:
Do not merge as-is. The selector-validation direction is useful, but this PR currently regresses existing prompt dependency behavior and CI is red.

Required changes:

1. Fix the sanitizer so it does not remove valid module prompt includes before `auto-deps` updates `architecture.json`. The failing CI case is `tests/test_auto_deps_main.py::test_auto_deps_main_updates_architecture_json_after_write`: on Linux, `<include>parent_python.prompt</include>` is replaced with `<!-- Missing: parent_python.prompt -->` even though `parent_Python.prompt` exists and the architecture layer intentionally maps this include to `parent_Python.prompt`. Because sanitization now runs before `merge_auto_deps_includes_from_cwd`, the newly inserted dependency is erased before the architecture merge can see it.

2. Make include validation mirror the project’s real resolution semantics for module prompt includes, including case-insensitive prompt/language suffix matching where existing sync/architecture logic supports it. A bare module prompt include that resolves through those semantics should remain an include; only truly missing files and genuinely invalid selectors should be replaced.

3. Add/adjust a Linux-safe regression test for the mixed-case module prompt include case so this does not depend on macOS case-insensitive filesystem behavior. The full unit-test check must be green before merge.

Verification observed: GitHub Actions `Run Unit Tests` failed with exactly this assertion, changing `%\n<include>parent_python.prompt</include>\n` into `%\n<!-- Missing: parent_python.prompt -->\n`. `git diff --check` and `py_compile` are clean, but the behavioral regression is merge-blocking.

gltanaka:
Additional required change before merge: the sanitizer needs to distinguish real include directives from literal include syntax examples.

The broadened matcher in `pdd/validate_prompt_includes.py` now catches attributed forms everywhere in the prompt text. That means examples such as `` `<include query="...">path/to/file</include>` `` or `` `<include select="def:foo">path/to/file</include>` `` are treated as real includes and rewritten to comments when the target path is not present. I verified locally that `validate_prompt_includes()` turns those inline-code examples into `` `<!-- Missing: path/to/file -->` ``.

That is risky now that `change`, `update`, `split`, and `auto-deps` all call `sanitize_prompt_output()` before writing `.prompt` files: prompt specs that document include syntax can be corrupted even though those tags are not intended to be resolved. Please make validation mirror the preprocessor’s code-span protection, or otherwise skip/include-gate tags inside inline code and fenced code blocks, and add regressions for literal `<include select=...>` and `<include query=...>` examples.

Also keep the existing Linux CI failure as merge-blocking until the mixed-case auto-deps fixture/behavior is corrected and the full unit-test check is green.

gltanaka:
Addressed the legitimate review points in `cd13ab100`:

- Preserved valid module prompt includes that resolve by sync/architecture semantics, including Linux case-sensitive `parent_python.prompt` -> `parent_Python.prompt` matching.
- Skipped `<include ...>` tags inside inline code and fenced code so prompt syntax examples are not rewritten.
- Added regressions for the mixed-case module prompt include and literal `select=` / `query=` include examples.

Local validation:
- `tests/test_issue_225_paths_and_includes.py tests/test_auto_deps_main.py::test_auto_deps_main_updates_architecture_json_after_write`: 15 passed
- Full CI-equivalent unit command: 7602 passed, 29 skipped, 1 xfailed
- Architecture include checkups passed
- Issue #798 selector reproduction passed with no `ContentSelector failed` warning
- `py_compile` and `git diff --check` passed

PR #851, opened 2026-05-08
feat(split): add per-child verify gate

### LINKED ISSUE #850: pdd split: per-child verify gate with repair loop, resume from failed child, pdd checkup integration
Migrated from private tracking issue gltanaka/pdd#1390 per repo policy to keep PRs/issues public.\n\n## Summary

`pdd split` (agentic mode) currently runs a single end-of-pipeline verify (\\\`7a_verify_local\\\`) after every child has been extracted. Two failure modes follow:

1. **Cascading damage** — when verify finds N issues across M children, the agent has no way to know *which* child broke *which* invariant. The repair pass at step 8 sees the whole pipeline's failures at once and can't isolate.
2. **No resume granularity** — if the run dies between extracting child 5 and child 6 (crash, timeout, OOM, signal), state is lost on the in-flight child. The next invocation either re-extracts everything or skips the broken child entirely.

We've also seen evidence that prompt+code+example+tests don't stay in lock-step automatically (Greg's strict rule from CLAUDE.md) — and \\\`pdd checkup --pr\\\` exists but is not wired into the verify gate.

### A. Per-child verify gate with repair loop
- [ ] After each child extracts (Step 6), run \\\`validate_extraction(plan, current_work_dir)\\\` filtered to that child's files only — not the whole plan.
- [ ] On any error-severity failure, route to a repair sub-loop with the \\\`ValidationFailure.message\\\` strings as structured feedback in the RETRY NOTICE. Up to N=2 repair attempts per child.
- [ ] On all-pass, mark child as \\\`verified\\\` in state and continue.
- [ ] Today's end-of-pipeline \\\`7a_verify_local\\\` step continues to run as a final cross-cutting check (catches issues only visible at full-package scope, e.g. circular imports between children).

### B. Granular per-child state schema
- [ ] \\\`state[\"children_extracted_status\"]\\\` is a dict keyed by child name (or index): \\\`pending\\\` | \\\`extracted\\\` | \\\`verified\\\` | \\\`failed_extract\\\` | \\\`failed_verify\\\`.
- [ ] Resume picks up at the first non-\\\`verified\\\` child instead of restarting the loop. Children already \\\`verified\\\` are skipped without re-billing.
- [ ] State persistence (\\\`save_workflow_state\\\`) happens after every status transition so a crash between transitions doesn't lose work.

### C. Architecture conformance via \\\`pdd checkup\\\` integration
- [ ] \\\`7a_verify_local\\\` invokes \\\`pdd checkup --pr\\\` (or the equivalent scoped command) on the new package after the per-child gate passes. Catches prompt↔code drift that's invisible to file-level checks.
- [ ] Failures are surfaced through the same structured-feedback channel into step 8 repair.

### D. Tests
- [ ] State-transition tests: a \\\`failed_verify\\\` child triggers exactly one repair pass; success after the second try transitions to \\\`verified\\\`; second failure marks \\\`failed\\\`.
- [ ] Resume tests: state with 2 \\\`verified\\\`, 1 \\\`failed_verify\\\` resumes at the failed child, not from scratch.
- [ ] Integration: \\\`pytest tests/test_agentic_split_orchestrator.py\\\` passes.

### E. PDD sync
- [ ] \\\`pdd/prompts/agentic_split_orchestrator_python.prompt\\\` documents the per-child gate, new state schema, repair-loop bound.
- [ ] If the per-child verify exposes new public helpers, they're synced to prompt+example+tests.

## Why this matters

Today's pipeline can spend $30 extracting 8 children, then catch 311 test failures at the end (real number from the May 5 \\\`solving_orchestrator.py\\\` run). Even with #1388's runtime test-seam check landing, structural validation runs once; with this change every child is gated independently, so the failure budget is bounded per-child instead of compounding.

## Reuses (don't reinvent)

- Verify-then-repair pattern at \\\`pdd/agentic_e2e_fix_orchestrator.py:302-343\\\` (\\\`_verify_tests_independently\\\` + \\\`_handle_verification_failure\\\` classification + retry).
- State persistence: \\\`load_workflow_state\\\` / \\\`save_workflow_state\\\` in \\\`pdd/agentic_common.py\\\`.
- Static patch-path resolution \\\`_resolve_patch_path\\\` at \\\`pdd/split_validation.py:116-264\\\` — the per-child gate filters this same logic to one child's files.
- \\\`pdd checkup --pr\\\` invocation pattern from \\\`pdd/checkup_review_loop.py\\\`.

## Related

- gltanaka/pdd#1384 — defensive resilience (independent; this builds on top of #1387's --max-cost ceiling).
- gltanaka/pdd#1388 — plan & extraction quality (independent; this complements the runtime test-seam check from #1389 with structural verify per-child).
- promptdriven/pdd_cloud#1320 — originating run.

Serhan-Asad:
## Summary

Public migration of the per-child `pdd split` verify/resume work. Adds a deterministic Step 6v gate after each child extraction, bounded per-child repair attempts, granular per-child resume state, and final `pdd checkup --validate-arch-includes` integration before global repair/assess steps.

Closes #850

## Key changes

- Adds per-child extraction status tracking: `pending`, `extracted`, `verified`, `failed_extract`, `failed_verify`, `failed`.
- Reuses extracted children on resume while still running 6v validation.
- Stops Step 6 immediately on unrecovered extraction or terminal per-child verify failure, before spending budget on later siblings or final gates.
- Persists terminal child failure details and clears stale failure state when children recover.
- Supports explicit split-plan `new_example` / `new_test` validation paths and child-scoped seam failure filtering.
- Updates prompts, README, CHANGELOG, architecture metadata, examples, and focused tests.

## Verification

- `python -m py_compile` on changed Python/test files
- `python -m json.tool architecture.json`
- `git diff --check`
- `PDD_PATH=/tmp/pdd-public-850 python -m pytest -q tests/test_agentic_split_orchestrator.py tests/test_split_validation.py tests/test_architecture_registry.py tests/test_ci_drift_heal.py` — 321 passed
- `pdd checkup --validate-arch-includes --project-root /tmp/pdd-public-850` — no mismatches

## Migration note

This replaces the private repo tracking issue/PR with public issue #850 and this public PR.


gltanaka (changes requested):
Requesting changes because the new checkup gate is wired before the architecture it is supposed to validate exists/is refreshed.

`pdd/agentic_split_orchestrator.py`: `ordered_steps` still runs `7a_verify_local` before `7c_arch_sync`, and `_run_split_checkup(...)` is called from inside `7a_verify_local`. But the step-6 prompt and workflow say `architecture.json` is derived in `7c` from the newly written `<pdd-*>` prompt metadata, and step 6 explicitly tells extractors not to write architecture directly. As a result, the checkup validates the old/stale architecture state and can pass before `7c_arch_sync` creates or rewrites the new package architecture. Any architecture/include drift introduced by `7c` will not be routed through the checkup failure path or global repair loop.

Please move the `pdd checkup --validate-arch-includes --project-root <worktree>` gate to run after `7c_arch_sync` and before assessment/global repair, or add an equivalent post-`7c` verification step that feeds failures into `verify_failures` for step 8.

Serhan-Asad:
Addressed the requested checkup ordering in b00baecdb: the arch/include check now runs as 7d after 7c_arch_sync, before assessment/global repair, and step 8 re-syncs architecture before each checkup retry. Added regression coverage for both the post-7c ordering and repair recheck path. Local affected suites and GitHub CI are green.

gltanaka (approved):
The previous blocking issue is addressed. `7d_checkup` now runs after `7c_arch_sync`, and the repair loop refreshes architecture before re-running checkup, so architecture/include drift is checked against freshly synced metadata and still feeds `verify_failures` for step 8.

I rechecked the updated head locally: py_compile on the changed Python/test files, `python -m json.tool architecture.json`, `git diff --check`, the focused pytest set (`tests/test_agentic_split_orchestrator.py tests/test_split_validation.py tests/test_architecture_registry.py tests/test_ci_drift_heal.py`), and `python -m pdd checkup --validate-arch-includes --project-root ...` all pass.

PR #856, opened 2026-05-08
Fix sync runner dependency failure scheduling

gltanaka:
## Summary

- Continue scheduling modules whose own dependencies are satisfied after an unrelated sync module fails.
- Report pending modules as `Skipped (blocked)` only when they actually depend on a failed module; otherwise report them separately as `Skipped (not run)`.
- Update the progress footer to show that independent modules are still continuing after a failure, plus refresh the prompt/context/architecture metadata for `agentic_sync_runner`.

## Root Cause

In the #798 rerun, `cli_detector` failed architecture conformance while `agentic_common` was still running. `setup_tool` was correctly blocked because it depends on `cli_detector`, but `agentic_update` only depends on `agentic_common`. The runner set `self.failed = True` after the first failure and then globally stopped starting new modules, so `agentic_update` was stranded even though its own dependency later succeeded. The final summary then labeled every pending module as `Skipped (blocked)`, hiding the difference between a real failed-dependency block and a global fail-fast skip.

## Tests

- `pytest tests/test_agentic_sync_runner.py -q` — 118 passed
- `pytest tests/test_agentic_sync.py tests/test_agentic_sync_runner.py -q` — 268 passed
- `python -m compileall -q pdd/agentic_sync_runner.py`
- `python -m json.tool architecture.json >/dev/null`
- `git diff --check`
- `python -m pdd checkup --validate-arch-includes`

### DIFF
--- a/architecture.json
+++ b/architecture.json
@@ -7193,7 +7193,7 @@
   },
   {
     "reason": "Parallel sync engine with env-configurable per-module timeout (PDD_MODULE_TIMEOUT_SECONDS, default 2700s) plus per-run --timeout-adder forwarded via sync_options['timeout_adder'], phase-aware heartbeat, and live GitHub progress updates.",
-    "description": "Manages pdd sync subprocesses using ThreadPoolExecutor, preserves architecture subdirectory basenames, respects dependency ordering from architecture.json, enforces total-budget sequential execution when requested, posts live progress to GitHub issue comments, and pauses on failure.",
+    "description": "Manages pdd sync subprocesses using ThreadPoolExecutor, preserves architecture subdirectory basenames, respects dependency ordering from architecture.json, enforces total-budget sequential execution when requested, posts live progress to GitHub issue comments, keeps unrelated ready modules running after a module failure while blocking transitive failed-dependency dependents, and stops all new scheduling when total budget is exhausted.",
     "dependencies": [
       "architecture_sync_python.prompt",
       "agentic_langtest_python.prompt",
--- a/context/agentic_sync_runner_example.py
+++ b/context/agentic_sync_runner_example.py
@@ -98,6 +98,12 @@ def main():
     print("\n--- Module States ---")
     for name, state in runner.module_states.items():
         print(f"  {name}: {state.status} (cost: ${state.cost:.2f})")
+    print(
+        "A failed module blocks only modules that depend on it; unrelated "
+        "modules continue when their own dependencies are satisfied. The "
+        "blocked set is transitive, and total-budget exhaustion still stops "
+        "all new scheduling."
+    )
 
     # --- Comment Body Preview ---
     print("\n--- GitHub Comment Preview ---")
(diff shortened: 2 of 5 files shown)


gltanaka:
Do not merge yet. I found two required changes in `pdd/agentic_sync_runner.py` before this is safe to land:

1. Total-budget exhaustion can be missed after a failed module. In `_run_inner`, the loop skips the top-level budget check once `self.failed` is true, and the post-result budget check only runs in the success branch. Repro: with `total_budget=1.0`, independent modules `a` and `b`, and `a` returning `(False, 1.0, ...)`, the runner still starts `b`; it only marks `budget_exhausted=True` after `b` finishes. That violates the requirement that total-budget exhaustion is the one global stop condition. Please check `_budget_exhausted()` after every completed future, including failures, before scheduling more work, and add a regression test.

2. Blocked-module classification is only direct, not transitive. `_get_blocked_modules()` only marks a pending module blocked when one of its immediate dependencies is `failed`. For `a -> b -> c`, if `a` fails, the summary reports `b` as `Skipped (blocked)` but `c` as `Skipped (not run)`, even though `c` is also blocked by the failed dependency chain. Please compute blocked pending modules transitively, or otherwise classify pending modules whose dependency path reaches a failed module as blocked, and add a regression test for the chain case.

Existing PR tests pass locally: `pytest tests/test_agentic_sync_runner.py -q` returned `118 passed`.

gltanaka:
Reviewed the PR comments.

Legit and fixed:

1. Total-budget exhaustion after failed modules. The runner now checks `_budget_exhausted()` after every completed future, including failed module results, before scheduling more work. Added `test_total_budget_exhaustion_after_failed_module_stops_new_modules`.

2. Transitive blocked-module classification. `_get_blocked_modules()` now marks pending modules blocked when any dependency path reaches a failed module. Added `test_blocked_by_transitive_failed_dep`.

Not actionable:

- Greptile free-trial notice is not a code review finding.

Validation after the fix and rebase:

- `pytest tests/test_agentic_sync_runner.py -q` — 120 passed
- `pytest tests/test_agentic_sync.py tests/test_agentic_sync_runner.py -q` — 270 passed
- `python -m compileall -q pdd/agentic_sync_runner.py`
- `python -m json.tool architecture.json >/dev/null`
- `git diff --check`
- `python -m pdd checkup --validate-arch-includes`

gltanaka:
Re-reviewed the updated PR. The two previous blockers are resolved:

- total-budget exhaustion is now checked after failed module results before scheduling more work;
- blocked-module classification now handles transitive failed-dependency chains.

Local verification passed:

- `pytest tests/test_agentic_sync_runner.py -q` -> `120 passed`
- `pytest tests/test_agentic_sync.py tests/test_agentic_sync_runner.py -q` -> `270 passed`
- `python -m compileall -q pdd/agentic_sync_runner.py`
- `python -m json.tool architecture.json >/dev/null`
- `git diff --check origin/main...HEAD`
- `python -m pdd checkup --validate-arch-includes`

No remaining required changes from my review. This looks mergeable.

PR #857, opened 2026-05-08
fix: clean public release flow

gltanaka:
## Summary
- split PyPI upload out of the build target
- make publish build and upload only to PyPI
- guard make release so it only runs from a public promptdriven/pdd origin with a clean worktree
- remove legacy publish-public and post-release changelog steps from the release path

## Verification
- python scripts/check_deps.py
- git diff --check -- Makefile
- make check-release-remote (expected failure in this checkout because origin still points to gltanaka/pdd)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,10 +54,11 @@ help:
 	@echo "Build & Deployment:"
 	@echo "  make install                 - Install pdd"
 	@echo "  make build                   - Build pdd package"
-	@echo "  make publish                 - Build & upload current version"
+	@echo "  make upload-pypi             - Upload dist/*.whl to PyPI"
+	@echo "  make publish                 - Build & upload current version to PyPI"
 	@echo "  make publish-public          - Copy artifacts to public repo only"
 	@echo "  make check-deps              - Check pyproject.toml and requirements.txt are in sync"
-	@echo "  make release                 - Bump version and build package (runs check-deps first)"
+	@echo "  make release                 - Bump version, tag, and upload to PyPI (public origin only)"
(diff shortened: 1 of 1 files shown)


gltanaka:
I would not merge this as-is. The release path still has a couple of merge-blocking safety gaps that should be fixed first.

Requested changes:

1. Make the release recipe fail fast before publishing. In the `else` branch, `python -m commitizen bump ...` and `git push origin main --tags` are followed by `;` inside a single shell block, so a failure in either command does not stop the later `$(MAKE) publish`. Please use `set -e`, explicit `&&`, or split commands so a failed bump or push cannot still upload to PyPI.

2. Guard that release is running from the expected branch and synced state, not just from the expected remote URL. A clean feature branch with `origin` set to `promptdriven/pdd` would currently pass `check-release-remote` and `check-release-clean`, then build/publish from the wrong checkout while pushing `origin main --tags`. Please require the current branch to be `main` and verify it is aligned with `origin/main` before allowing release.

3. Add `release` to `.PHONY` while touching this target. Without it, a file or directory named `release` can cause `make release` to become a no-op.

The CI checks are green, but these are release-flow correctness issues that the current checks do not exercise.

gltanaka:
Addressed the valid release-flow comments:

- Added `release` to `.PHONY`.
- Added `check-release-branch` so release only runs from `main` and requires local `HEAD` to match freshly fetched `origin/main`.
- Added `set -e` to the release shell block so a failed Commitizen bump, push, or publish stops the recipe before later release actions.

The Greptile free-trial note is non-actionable.

PR #867, opened 2026-05-08
Harden issue sync prompt-contract validation

gltanaka:
## Summary
- fail prompt-contract preflight when a prompt-local interface declares existing symbols without existing module source context
- run the same prompt-contract preflight during agentic dry-run validation before accepting a module cwd
- label review-loop fixer attempts as verified/unverified in final reports so unverified fixes are not treated as authoritative
- update the prompt specs for the changed prompt-driven modules

## Root cause
The failed #798 run had two bad prompt/code contracts. `agentic_common` declared a broad existing interface with only partial source context, and `setup_tool` declared a prompt-local interface without any existing source context. Plain `pdd sync --dry-run` could pass those modules, so the agentic issue flow accepted them and only failed later during real sync/conformance. The later review-loop pause was a correct safety refusal, but the final report made fixer attempts too easy to confuse with verified fixes.

## Test plan
- pytest -q tests/test_prompt_contract_validation.py tests/test_agentic_sync.py tests/test_agentic_sync_runner.py tests/test_checkup_review_loop.py
- reproduced PR #858 prompt-contract failures for `agentic_common` and `setup_tool` with the updated preflight
- end-to-end temp project: plain `pdd sync bad --dry-run --agentic --no-steer` passes, while `_run_dry_run_validation(["bad"], root)` now fails at zero cost with `prompt contract preflight failed`

Related to #798


gltanaka:
Do not merge yet.

The new prompt-local-interface/no-self-include hard failure is global, but the PR leaves the current first-party prompt set in a newly failing state. On this PR, running the validator across the root `architecture.json` reports 26 failures; the same check on `origin/main` reports 0. Example affected modules include:

- `setup_tool_python.prompt -> pdd/setup_tool.py`
- `architecture_sync_python.prompt -> pdd/architecture_sync.py`
- `architecture_registry_python.prompt -> pdd/architecture_registry.py`
- `core/cli_python.prompt -> pdd/core/cli.py`
- `core/duplicate_cli_guard_python.prompt -> pdd/core/duplicate_cli_guard.py`

A direct repro also shows the new agentic dry-run gate now rejects an existing repo module at zero cost:

```py
from pathlib import Path
from pdd.agentic_sync import _run_dry_run_validation
print(_run_dry_run_validation(["setup_tool"], Path.cwd(), quiet=True))
```

This returns `ok=False` with:

```text
setup_tool: prompt contract preflight failed:
- .../pdd/prompts/setup_tool_python.prompt: prompt-local interface declares 1 public symbols already present in .../pdd/setup_tool.py but includes no existing module source context: missing run_setup.
```

Because `validate_prompt_contract_context` is used by the normal sync preflight as well as by the new agentic dry-run validation, merging this would immediately make a set of existing first-party modules unsyncable until their prompts are manually repaired.

1. Either migrate the existing first-party prompts that the new rule rejects by adding full self-includes or selected includes covering every declared existing public symbol, or narrow/stage the new hard failure so current valid repo workflows remain syncable.
2. Add a regression check that runs this prompt-contract validator over the repository architecture (or an equivalent representative fixture) so future changes cannot introduce repo-wide prompt-contract failures silently.
3. Keep the current targeted tests; they pass locally. Full `pytest -q -x` currently stops in `tests/commands/test_connect.py::test_connect_custom_host_port`, but that same failure reproduces on `origin/main`, so I am not treating it as PR-attributable.

gltanaka:
Reviewed the PR comments and fixed the legitimate blocker.

Legit:
- The no-self-include hard failure was too broad and made existing first-party prompts fail the prompt-contract scan.
- `_run_dry_run_validation(["setup_tool"], Path.cwd(), quiet=True)` did reject a current repo module before the fix.
- A repo-wide regression check was missing.

Not PR-actionable:
- Greptile trial/billing note.
- The full `pytest -q -x` `test_connect_custom_host_port` failure noted in the comment is already reproducible on `origin/main`.

Fix pushed in `f663e713a`:
- Kept partial self-include contract failures global.
- Staged the no-self prompt-local hard failure behind `require_prompt_local_source_context`.
- Agentic sync enables that strict no-self rule only when the prompt file changed in git, so newly generated/changed prompt contracts are checked while existing first-party legacy prompts remain syncable.
- Added a repo-wide prompt-contract regression test.

Validation:
- `pytest -q tests/test_prompt_contract_validation.py tests/test_agentic_sync.py tests/test_agentic_sync_runner.py tests/test_checkup_review_loop.py` = 335 passed
- Repo-wide default prompt-contract scan = 0 failures
- `_run_dry_run_validation(["setup_tool"], Path.cwd(), quiet=True)` now returns ok=True
- PR #858 clone still reports prompt-contract errors for `agentic_common` and `setup_tool`
- Temp git end-to-end repro rejects a changed no-self prompt-local contract at zero cost with `prompt contract preflight failed`

gltanaka:
Re-reviewed the updated PR at `f663e713`.

The previous blocker is resolved:
- Repo-wide default prompt-contract scan now reports 0 failures.
- `_run_dry_run_validation(["setup_tool"], Path.cwd(), quiet=True)` now succeeds with no errors.
- The changed PR modules (`agentic_sync`, `architecture_include_validation`, `checkup_review_loop`) also pass the new strict prompt-contract gate.
- Focused validation passed locally: `pytest -q tests/test_prompt_contract_validation.py tests/test_agentic_sync.py tests/test_agentic_sync_runner.py tests/test_checkup_review_loop.py` = 335 passed.
- GitHub status checks are green and the PR is mergeable.

I do not see remaining blocking reasons to hold this PR. OK to merge.

PR #900, opened 2026-05-10
fix(llm): clamp temperature to 1.0 for Gemini 3 models (litellm warning)

gltanaka:
## Summary

litellm prints the following warning whenever a Gemini 3 model is invoked with `temperature < 1.0`:

> Warning: Setting temperature < 1.0 for Gemini 3 models (gemini-3-flash-preview) can cause infinite loops, degraded reasoning performance, and failure on complex tasks. Strongly recommended to use temperature = 1.0 (default).

PDD's default temperature is `0.1` (`llm_invoke(... temperature: float = 0.1, ...)`) and several call sites pass `temperature=0.0`. Path C validation of #1405 / #899 surfaced the warning with `vertex_ai/gemini-3-flash-preview`, exposing PDD to the documented production failure modes (infinite loops, degraded reasoning).

This PR adds a narrow, model-family-scoped clamp: whenever the resolved model is in the Gemini 3 family (`gemini-3`, `gemini-3.x` with any routing prefix — `gemini/`, `vertex_ai/`, `gmi/google/`, `github_copilot/`, or the bare Vertex AI name), PDD forces `temperature=1.0` before handing kwargs to litellm.

### What's intentionally NOT in this PR

- Other model families' temperature defaults are unchanged.
- The litellm warning itself is not silenced — that's litellm's concern; we just stop triggering it.
- No litellm version bump, no broader LLM/temperature refactor.

## Why the change is structured this way

An inline `'gemini-3' in model_name.lower()` override already existed at the `litellm.completion` call site, but it had two gaps:

1. **Batch path bypass.** `litellm.batch_completion` was called earlier in the same `if/else`, before the override, so callers using `use_batch_mode=True` got no protection.
2. **Substring matching is too loose** — `'gemini-3' in name` would also match a hypothetical `gemini-30-something`. A boundary-anchored regex (`gemini-3(?:\.\d+)?(?=$|[-_./\s])`) matches the documented Gemini 3 family identifiers and only those.

The fix extracts the logic into two helpers (`_is_gemini_3_model`, `_maybe_clamp_gemini_3_temperature`) and invokes the clamp once, BEFORE the batch/single split, so both `litellm.completion` and `litellm.batch_completion` are covered. The redundant late-stage inline override is replaced with a pointer comment so future readers know where the single source of truth lives.

## Where the override lives

- Helpers: [`pdd/llm_invoke.py:1640-1708`](https://github.com/promptdriven/pdd/blob/fix/gemini-3-temperature-clamp/pdd/llm_invoke.py#L1640-L1708)
- Call site (early, BEFORE batch/single split): [`pdd/llm_invoke.py:3183-3189`](https://github.com/promptdriven/pdd/blob/fix/gemini-3-temperature-clamp/pdd/llm_invoke.py#L3183-L3189)
- Tests: [`tests/test_llm_invoke.py::TestGemini3TemperatureClamp`](https://github.com/promptdriven/pdd/blob/fix/gemini-3-temperature-clamp/tests/test_llm_invoke.py) (8 cases, including helper unit test, batch path coverage, and negative cases for Gemini 2.x / Gemini 1.5 / non-Gemini)
- Prompt-as-source-of-truth updated: [`pdd/prompts/llm_invoke_python.prompt:143`](https://github.com/promptdriven/pdd/blob/fix/gemini-3-temperature-clamp/pdd/prompts/llm_invoke_python.prompt#L143)

## Empirical evidence

Ad-hoc verification script (`/tmp/verify_gemini3_clamp.py`) calls `llm_invoke` with multiple model-name shapes and captures the `temperature` kwarg that reaches litellm:

```
MODEL                                               INPUT  BATCH    GOT  EXPECT  OK?
------------------------------------------------------------------------------------------
gemini/gemini-3-flash-preview                         0.1  False      1       1  YES
gemini/gemini-3-pro-preview                           0.0  False      1       1  YES
vertex_ai/gemini-3-flash-preview                      0.0  False      1       1  YES
gemini-3.1-pro-preview                                0.7  False      1       1  YES
gmi/google/gemini-3-pro-preview                       0.5  False      1       1  YES
vertex_ai/gemini-3-flash-preview                      0.0   True      1       1  YES
gemini/gemini-2.5-flash                               0.1  False    0.1     0.1  YES
gemini-1.5-pro                                        0.2  False    0.2     0.2  YES
gpt-4o-mini                                           0.1  False    0.1     0.1  YES

OVERALL: PASS
```

Every Gemini 3 shape receives `temperature=1`; every non-Gemini-3 model keeps its caller-supplied temperature.

## TDD evidence

- `7d59bd7f1` — `test(llm): add failing tests for Gemini 3 temperature clamp`. Two tests fail: the helper classifier (helper didn't exist) and the batch-mode clamp (existing late override didn't cover batch). Six already passed thanks to the existing late inline override — kept as regression coverage.
- `5546f6ae1` — `fix(llm): clamp temperature to 1.0 for Gemini 3 models (litellm warning)`. Adds helpers, moves the clamp earlier, removes the redundant late inline override.

## Test plan

- [x] `pytest tests/test_llm_invoke.py::TestGemini3TemperatureClamp` — 8 passing
- [x] `pytest tests/test_llm_invoke.py tests/test_llm_invoke_csv_model_registration.py tests/test_llm_invoke_integration.py tests/test_llm_invoke_nested_schema.py tests/test_llm_invoke_retry_cost.py tests/test_llm_invoke_vertex_retry.py` — 307 passed, 3 skipped (pre-existing skips)
- [x] `pylint pdd/llm_invoke.py` — no new warnings (score 6.79/10 before, 6.79/10 after; pre-existing import-order issues only)
- [x] `pylint --disable=all --enable=E,F tests/test_llm_invoke.py` — 10.00/10
- [x] Ad-hoc verification script confirms actual temperature kwarg reaching litellm
- [ ] `/gcbrun` (Cloud Batch full suite — posting below)

### DIFF
--- a/pdd/llm_invoke.py
+++ b/pdd/llm_invoke.py
@@ -1637,6 +1637,78 @@ def _model_disallows_temperature(model_name: Any) -> bool:
     return "claude-opus-4-7" in model_lower or "claude-opus-4.7" in model_lower
 
 
+# Regex anchored to the Gemini 3 family identifier so that:
+#   - matches: gemini-3, gemini-3-flash, gemini-3-pro, gemini-3.1-..., gemini-3.5-...
+#   - does NOT match: gemini-2.5-*, gemini-1.5-*, gemini-30-* (hypothetical
+#     unrelated families), claude-3, etc.
+# The lookahead requires a separator (-, _, /, ., end-of-string, or whitespace)
+# after the version number so that "gemini-30-something" cannot be mistaken for
+# a Gemini 3 variant.
+_GEMINI_3_RE = re.compile(r"gemini-3(?:\.\d+)?(?=$|[-_./\s])", re.IGNORECASE)
+
+
+def _is_gemini_3_model(model_name: Any) -> bool:
+    """Return True when *model_name* belongs to the Gemini 3 family.
+
+    See ``_GEMINI_3_RE`` for the matching contract. This intentionally covers
+    every routing prefix (``gemini/``, ``vertex_ai/``, ``gmi/google/``,
+    ``github_copilot/``, bare Vertex AI names like ``gemini-3.1-pro-preview``)
+    by anchoring on the model family identifier itself rather than the prefix.
+    """
+    if model_name is None:
+        return False
+    return bool(_GEMINI_3_RE.search(str(model_name)))
+
+
+def _maybe_clamp_gemini_3_temperature(
+    litellm_kwargs: Dict[str, Any], model_name: Any, verbose: bool
+) -> bool:
+    """Force ``temperature=1`` in *litellm_kwargs* when the model is Gemini 3.
+
+    WHY this override exists (do NOT remove without checking the litellm
+    upstream behavior):
+
+      litellm prints the following warning for Gemini 3 models invoked with
+      temperature < 1.0::
+
+          Warning: Setting temperature < 1.0 for Gemini 3 models
+          (gemini-3-flash-preview) can cause infinite loops, degraded
+          reasoning performance, and failure on complex tasks. Strongly
+          recommended to use temperature = 1.0 (default).
+
+      PDD's defaults pass ``temperature=0.1`` and many callers pass 0.0
+      explicitly, which would trip this warning on every Gemini 3 call and
+      expose us to the documented production failure modes (infinite loops,
+      degraded reasoning). The fix is a narrow, model-family-scoped clamp
+      that only fires for Gemini 3; every other model keeps the
+      caller-supplied temperature unchanged.
+
+    Returns True iff the clamp fired (used to keep ``current_temperature``
+    in sync at the call site).
+    """
+    if "temperature" not in litellm_kwargs:
+        # Caller already removed temperature (e.g. provider rejects it) —
+        # nothing to clamp.
+        return False
+    if not _is_gemini_3_model(model_name):
+        return False
+    current = litellm_kwargs.get("temperature")
+    try:
+        if current is not None and float(current) >= 1.0:
+            return False
+    except (TypeError, ValueError):
+        return False
+    if verbose:
+        logger.info(
+            "[INFO] Gemini 3 model detected: forcing temperature=1 (was %s). "
+            "See litellm warning re infinite loops / degraded reasoning at "
+            "temperature < 1.0.",
+            current,
+        )
+    litellm_kwargs["temperature"] = 1
+    return True
+
+
 def _response_usage_summary(response: Any) -> Dict[str, Any]:
     """Return token and finish metadata from a LiteLLM response."""
     responses = response if isinstance(response, list) else [response]
@@ -3108,6 +3180,14 @@ def llm_invoke(
                 # Use a local adjustable temperature to allow provider-specific fallbacks.
                 litellm_kwargs["temperature"] = current_temperature
 
+            # Gemini 3 family clamp: temperature < 1.0 is documented by litellm
+            # to cause infinite loops and degraded reasoning. Apply BEFORE the
+            # batch/single split so both litellm.completion and
+            # litellm.batch_completion paths are protected. See
+            # _maybe_clamp_gemini_3_temperature for the rationale.
+            if _maybe_clamp_gemini_3_temperature(litellm_kwargs, model_name_litellm, verbose):
+                current_temperature = 1
+
             # --- Resolve API key / credentials ---
             # The CSV api_key field may be:
             #   - Single env var (e.g. "ANTHROPIC_API_KEY") → pass as api_key=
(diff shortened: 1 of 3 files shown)


gltanaka (commented):
## Verdict (iter 1)
APPROVE

None.

## Majors (should fix; merge if implementer commits to follow-up)
None.

## Minors (nits / suggestions)

1. **Prompt clause wording — slight over-promise.** `pdd/prompts/llm_invoke_python.prompt:143` says "Force temperature=1 *regardless of caller-supplied temperature*", but the actual helper only clamps when `current < 1.0` — a caller passing `temperature=1.5` will pass through unchanged (verified empirically). Not a behavior bug (the litellm warning is only about `< 1.0`), but a future regen from this prompt could write stricter code that overrides explicit `1.5 -> 1`. Consider rewording to "raise temperature to 1.0 whenever the caller supplied a value below 1.0".

2. **`_is_gemini_3_model` accepts non-string types via `str(...)`.** `_is_gemini_3_model(['gemini-3-flash'])` returns True because `str([...])` embeds the literal substring. Not exploitable in practice (callers always pass strings from CSV), and the helper correctly rejects `None`, but a stricter `isinstance(model_name, str)` gate would harden the contract. Pure nit.

3. **Empty string is handled by `_GEMINI_3_RE.search('')` returning `None`, not by an explicit guard.** Works correctly; just noting that None vs '' use different code paths (the `None` check at line 1658 short-circuits; empty string falls through to `re.search`).

## What this PR does well

- **TDD discipline is real, not theatrical.** I checked out commit `7d59bd7f` with `origin/main`'s `pdd/llm_invoke.py` and confirmed exactly 2 tests fail (`test_is_gemini_3_model_classifier` — helper didn't exist; `test_gemini_3_batch_mode_temperature_clamped` — old inline override was in the `else` branch of `use_batch_mode`, never ran for batch). The 6 already-passing tests are valid regression coverage for the prior substring-match override. Matches the implementer's claim exactly.
- **Regex boundary-anchoring is rigorous.** I constructed 35 adversarial cases including `gemini-30-flash`, `gemini-3xx`, `GEMINI-3X` (no separator after `3`), `gemini-flash-3` (reversed naming), `gemini-3.10-flash`, `gmi/google/gemini-3-pro-preview`, `claude-3`, mixed-case, leading/trailing whitespace. All 35 produce the correct verdict.
- **Call site placement is correct.** `_maybe_clamp_gemini_3_temperature` runs at `pdd/llm_invoke.py:3188`, BEFORE the `if use_batch_mode: ... else: ...` split at line 3737. The previous inline override at `pdd/llm_invoke.py:3679-3688` (pre-fix) is correctly removed and replaced with a pointer comment — verified with `git show origin/main:pdd/llm_invoke.py`. No duplication, no silent bypass.
- **Retry loop interaction is safe.** When `current_temperature` is adjusted for Claude thinking errors at line 4411, the retry re-enters at line 3148, re-applies `litellm_kwargs["temperature"] = current_temperature` at line 3181, and the Gemini 3 clamp re-fires at line 3188. The two override paths (Claude thinking, Gemini 3) gate on different model-name predicates and cannot collide.
- **No scope creep.** +472/-10 lines decomposes as: +85/-10 in `llm_invoke.py` (70 lines of helper + docstrings, 15 lines call site changes), +1 in the prompt, +386 in tests (8 thorough end-to-end tests, each ~40-50 lines because of CSV+monkeypatch ceremony consistent with surrounding test conventions). No refactor, no model default changes, no litellm bump.

## Verification I did

- Cloned `pdd` fresh into `/tmp/pr-900-review` and ran `git log --oneline origin/main..HEAD` — TDD order confirmed (`7d59bd7f` test commit precedes `5546f6ae` fix commit).
- Checked out the failing-test commit with `origin/main`'s production code and ran the new tests: exactly 2 failed as advertised, 6 passed as regression coverage.
- Restored the PR state and ran the full `tests/test_llm_invoke.py` suite — 281/281 passed.
- Read the helper at `pdd/llm_invoke.py:1640-1709` and the call site at `pdd/llm_invoke.py:3183-3189`. Verified the helper is called before the batch/single split.
- Ran 35 adversarial cases through `_GEMINI_3_RE` and `_is_gemini_3_model` directly — all pass.
- Ran the implementer's `/tmp/verify_gemini3_clamp.py` and reproduced the 9-case empirical table verbatim.
- Constructed and ran additional helper tests for edge cases: `temperature=None` (clamps to 1 when key present — debatable but defensible), `temperature='invalid'` (no-op, returns False), `temperature=1.5` (no-op, > 1.0), `temperature=1.0` (no-op).
- Ran `pylint pdd/llm_invoke.py`: score 6.79/10 (slightly *higher* than the implementer's reported 6.79/10 vs previous 6.75/10 — pylint actually awarded +0.04 for the new docstrings).
- Re-checked `gh pr checks 900`: all 7 checks pass (Analyze actions/javascript-typescript/python, CodeQL, Package Preprocess Smoke, Public CLI Regression, Run Unit Tests).

## Open questions

- The implementer flagged that there was previously an inline `'gemini-3' in model_name.lower()` override at the `litellm.completion` call site. I verified `git show origin/main:pdd/llm_invoke.py` at lines 3679-3688 shows the old substring-match override; the PR correctly removes it and replaces with a pointer comment. The `batch_completion` path was indeed bypassing it before this PR.
- The `_is_gemini_3_model` accepts non-string types by coercing via `str(...)`. A `list(['gemini-3-flash'])` would match because `str(['gemini-3-flash'])` contains the substring. Not exploitable; just a robustness nit if you want to lock down the contract.

**Conditions for merge:** None. CI is green, tests are real, the override is narrow and correct. Safe to merge.

PR #901, opened 2026-05-10
fix(llm): polish Gemini 3 clamp helper — prompt alignment + type guards (PR #900 follow-up)

gltanaka:
## Summary

Follow-up polish for PR #900 (now merged as `2d27b55c3`). The reviewer flagged three advisory minors that did not block merge; this PR addresses all three with TDD evidence.

Refs: promptdriven/pdd#900

## Reviewer's verbatim minors

> 1. Prompt clause says "regardless of caller-supplied temperature" but helper only clamps when `< 1.0` — minor over-promise that could mislead future regen.
> 2. `_is_gemini_3_model` accepts non-string types via `str(...)` coercion; `['gemini-3-flash']` would match. Pure robustness nit.
> 3. Empty string handled by `re.search` returning None, not an explicit guard — works correctly.

### 1. Prompt-code alignment

**File:** `pdd/prompts/llm_invoke_python.prompt:143`

Reworded the Gemini 3 clause from `"Force temperature=1 regardless of caller-supplied temperature"` to `"When the caller-supplied temperature is below 1.0, clamp it up to temperature=1 (callers that explicitly pass temperature>=1.0 are left untouched)"`. The previous phrasing would have led a future regen of `llm_invoke.py` to over-clamp.

### 2. Non-string type guard

**File:** `pdd/llm_invoke.py:1650` (`_is_gemini_3_model`)

Replaced `if model_name is None: return False / str(model_name)` with an explicit `isinstance(model_name, str)` guard. Before the fix, a list like `["gemini-3-flash"]` stringified to `"['gemini-3-flash']"` and falsely matched the regex (verified locally — `_is_gemini_3_model(['gemini-3-flash'])` returned `True`). After the fix it returns `False` for any non-string type (list, dict, tuple, int, bool, bytes).

The only call site (`_maybe_clamp_gemini_3_temperature`) is unchanged — non-string input now correctly does NOT trigger a clamp instead of triggering one with an unintended container.

### 3. Empty-string explicit guard

**File:** `pdd/llm_invoke.py:1650` (`_is_gemini_3_model`)

Empty string already returned `False` because `re.search` returned `None`, but pinned the explicit guard (`not model_name`) so a future regex change can't accidentally make `""` truthy.

## TDD evidence

Commits land in red-then-green order so the reviewer can verify the failing-test-first contract:

- `18ff1c6a9` — `test(llm): add failing tests for Gemini 3 helper polish (PR #900 minors)`
  - `test_is_gemini_3_model_rejects_non_string_types` FAILS on `upstream/main` with `assert True is False` for `['gemini-3-flash']`.
  - `test_is_gemini_3_model_empty_string_explicit_guard` PASSES (behavior unchanged, pins contract).
- `debc786a1` — `fix(llm): polish Gemini 3 clamp helper — prompt alignment + type guards`
  - All 10 tests in `TestGemini3TemperatureClamp` pass (8 originals from PR #900 + 2 new).

## Test plan

- [x] `pytest tests/test_llm_invoke.py -v -k "Gemini3 or _is_gemini_3"` — 10/10 pass locally
- [x] Confirmed `test_is_gemini_3_model_rejects_non_string_types` FAILS on `upstream/main` (TDD red phase)
- [x] All 8 original `TestGemini3TemperatureClamp` tests (from PR #900) still pass
- [x] Full `tests/test_llm_invoke.py` suite passes (283/283)
- [ ] CI green (`/gcbrun` will follow this PR creation)

## Scope discipline

- Only three files touched: `pdd/llm_invoke.py`, `pdd/prompts/llm_invoke_python.prompt`, `tests/test_llm_invoke.py`.
- No refactoring of `_maybe_clamp_gemini_3_temperature` or surrounding code.
- No changes to the regex `_GEMINI_3_RE`.


gltanaka (commented):
## Verdict (iter 1)
APPROVE

(Posting as `--comment` per branch protection: GitHub blocks self-approval on the same org's PRs. Treat this as a green-light review.)

None.

## Verification I did
Checked out fresh clone of the branch at `pr-901-review`, ran all checks below myself rather than trusting the PR description.

### 1. TDD ordering reproduced (red → green)
- `git log --oneline origin/main..HEAD` confirms `18ff1c6a` (failing tests) precedes `debc786a` (source fix).
- `18ff1c6a` touches **only** `tests/test_llm_invoke.py` (`git show --stat 18ff1c6a` — clean TDD, no source sneak-in).
- Reproduced the red phase by checking out the new tests from `18ff1c6a` while keeping `pdd/llm_invoke.py` at `origin/main`:
  - `test_is_gemini_3_model_rejects_non_string_types` **FAILS** on the `is_g3([\"gemini-3-flash\"]) is False` assertion — `str(model_name)` coercion produced `\"['gemini-3-flash']\"`, which the regex matched. This is exactly the minor the PR #900 reviewer flagged.
  - `test_is_gemini_3_model_empty_string_explicit_guard` passes on red too (since `re.search` already returned None for `\"\"`), which is the expected behavior pinned by the test — the test exists to lock the contract, not to fail on red.

### 2. Helper behavior (PR HEAD) probed against every case in the brief
| Input | Result | Expected |
|---|---|---|
| `None` | False | False |
| `[\"gemini-3-flash\"]` | False | False |
| `[\"gemini-3-pro-preview\", \"gpt-5\"]` | False | False |
| `{\"model\": \"gemini-3-flash\"}` | False | False |
| `(\"gemini-3-flash\",)` | False | False |
| `\"\"` | False | False |
| `\"gemini-3-flash\"` | True | True |
| `0` / `3` / `3.1` | False | False |
| `True` / `False` | False | False |
| `b\"gemini-3-flash\"` | False | False |
| `\"GEMINI-3-FLASH\"` | True | True (case-insensitive regex) |
| `\"vertex_ai/gemini-3-pro-preview\"` | True | True |
| `\"gemini-2.5-flash\"` | False | False |
| `\"gemini-1.5-pro\"` | False | False |

All 19 probes match expectations.

### 3. Prompt clause (line 143) now matches helper behavior
- New wording: `\"When the caller-supplied temperature is below 1.0, clamp it up to temperature=1 (callers that explicitly pass temperature>=1.0 are left untouched)\"`.
- `_maybe_clamp_gemini_3_temperature` returns `False` (no clamp) when `float(current) >= 1.0`. Wording and implementation agree, fixing the over-promise from `\"regardless of caller-supplied temperature\"` that the PR #900 reviewer flagged.
- Bonus: the prompt also now requires the family classifier to reject non-strings and treat empty strings as non-matches — this keeps any future regen of `llm_invoke.py` from re-introducing the `str(...)` coercion bug.

### 4. Test runs
- `pytest tests/test_llm_invoke.py -v -k Gemini3` → **10 passed** (the 8 originals from PR #900 + the 2 new defensive tests).
- `pytest tests/test_llm_invoke.py` → **283 passed in 4.77s** — matches the implementer's 283/283 claim exactly.

### 5. CI
- `gh pr checks 901` reports all 7 checks green (Analyze actions / js-ts / python, CodeQL, Package Preprocess Smoke, Public CLI Regression, Run Unit Tests).
- `mergeable: MERGEABLE`.

## Notes
- Scope is tight: 3 files, +58/-3 lines, all addressing the exact minors the PR #900 reviewer raised. No drift.
- The `_maybe_clamp_gemini_3_temperature` call site is unchanged. The behavioral delta is contained to `_is_gemini_3_model`: previously, handing a non-string container to the clamp helper could trigger an unintended `temperature=1` override at the call site; that path is now correctly inert.
- No new external dependencies, no API surface change, no schema change.

Approving (as a comment). Safe to merge.

Issue #814, opened 2026-05-04
agentic_common: "Credit balance is too low" 400s should be classified permanent, not retried


jamesdlevine:
## Summary

`_is_permanent_error` in `pdd/agentic_common.py` does not include the "Credit balance is too low" string (or general 400-class billing/auth signatures) in its pattern list. As a result, when an anthropic-provider call fails with HTTP 400 "Credit balance is too low", pdd's agentic orchestrator burns all retries on the same provider before falling back to the next one in the configured provider order (default `["anthropic", "google", "openai"]`).

## Repro / observed behavior

Captured `.pdd/agentic-logs/session_*.jsonl` from a failing `pdd generate` run shows 12 sequential anthropic calls, each with `success: False`, `cost_usd: 0.0`, and a 513-byte response body containing:

```
{...,"is_error":true,"api_error_status":400,"result":"Credit balance is too low",...}
```

across labels `step1`, `step1b`, `step2`, `step2b`, `step3`, `step4`, `step5`, `step5b_attempt1`, `step5b_fix1`, `step5b_attempt2`, `step5b_fix2`, `step5b_attempt3`. Each was retried until exhausted before the orchestrator moved on.

## Impact

- Wastes ~25–30 seconds per attempt on a clearly permanent error (~5–6 minutes of wasted wall-clock in the captured run).
- Obscures the root cause — there is no clear "your credit balance is empty / your provider is unavailable" signal at the top of the run.

## Suggested fix

Add a case-insensitive `"credit balance is too low"` (and likely other clear billing/auth-shaped 400 strings) to the permanent-error patterns in `_is_permanent_error`. Optionally also classify any combination of `is_error: true` + `api_error_status: 400` + a billing/auth-shaped result string as permanent. On classification, surface a clear top-of-output diagnostic ("anthropic provider unavailable: <reason> — skipping to next provider") rather than silently advancing.

## Code reference

`pdd/agentic_common.py` `_is_permanent_error` permanent-error patterns (approximate lines 309–341 in `pdd-cli==0.0.227`).

## Environment

- `pdd-cli==0.0.227`
- Surfaced during `pdd generate <issue-url>` after a `CI=1`-induced auth issue caused all anthropic calls to 400.

---

**Related:**
- #813 — the upstream auth issue that triggered the 400 storm motivating this report
- #816 — the logging gap that combined with this retry behavior to hide what was actually happening

Serhan-Asad:
5/11/2026

[...]

PR #931, opened 2026-05-11
chore: PDD sync for #923

### LINKED ISSUE #923: checkup_review_loop: no fallback when primary reviewer fails — verification dead-stops on infra-class errors
## Summary

When the primary reviewer (default: `codex`) fails during `pdd checkup --pr --review-loop`, the loop **breaks immediately at `pdd/checkup_review_loop.py:346-351`** with `stop_reason = "Primary reviewer <X> could not complete: failed."`. There is no automatic fallback to another reviewer role, so verification dead-stops and the downstream verdict adapter cannot determine merge-readiness.

Result: `pdd-issue` autonomous solving cannot finish even when the PR's fix itself is correct.

## Observed in production

Issue #814 / PR #914 — `pdd-issue` produced a correct credit-balance fix and a mergeable PR, but the Step 7/8 Review Loop Final Report showed:

```
reviewer-status: codex=failed claude=fixer fresh-final=missing
fresh-final-review: missing
max-rounds-reached: false
max-cost-reached: false
max-duration-reached: false

Summary
Primary reviewer codex could not complete: failed.
```

`pdd checkup --pr` then exited as "Required reviewer not clean", verification failed, and three retries all hit the same wall before the user stopped the job at $30.58 spend.

## Root cause (verified)

1. `pdd/checkup_review_loop.py:_run_review` (line 590) calls `_run_role_task` → `_forced_provider(<role>)` pins `PDD_AGENTIC_PROVIDER` to the single primary-reviewer role, so the existing in-task provider fallback in `run_agentic_task` is intentionally **defeated** for review-loop calls (`pdd/checkup_review_loop.py:846-857`, `pdd/agentic_common.py:_forced_provider`).
2. When `_run_role_task` returns `success=False`, `_failure_status(output)` (line 1865) decides `degraded` vs `failed` based only on a narrow set of degraded markers (`rate limit`, `quota`, `timeout`, `context length`, …). Any other CLI/auth/network failure becomes `failed`.
3. `failed` is in `HARD_NOT_CLEAN_STATES` (line 111), so the for-round loop hits the `break` at lines 346–351 and never tries another reviewer.
4. The "fixer" role (default: `claude`) is not a secondary reviewer; auto-promoting it would defeat two-eye review (see `_resolve_roles` line 524 enforcing `reviewer != fixer`).

## Proposed fix

Two-step, minimal-surface:

**a) Broaden `_failure_status` degraded markers** so transient CLI/infrastructure failures (auth, network, exec, sandbox) classify as `degraded` instead of `failed`. `degraded` is still not-clean, but pairs well with `--continue-on-reviewer-limit`.

**b) Add `--reviewer-fallback` config** (default empty). When the primary reviewer status lands in `HARD_NOT_CLEAN_STATES` and a fallback role is configured and differs from the fixer, the loop tries the fallback reviewer **once** before breaking. The original reviewer's `failed`/`degraded` status is preserved in `reviewer_status`; the fallback gets its own status entry; if the fallback also fails, the loop breaks as today.

Existing flag `--continue-on-reviewer-limit` already exists; this proposal extends what it covers + adds an opt-in reviewer-of-last-resort distinct from the fixer.

## Caveat — environment cause

The actual codex failure in pdd_cloud's job artifacts was not available to confirm root cause (`output.txt` lives in pdd_cloud, not local). Codex CLI works fine locally with the same prompt shape, so the failure is most likely an environment issue in pdd_cloud's executor (auth handshake, sandbox, or missing binary). A separate pdd_cloud-side investigation is warranted; the structural fix proposed here is needed regardless, because *some* class of provider failures will always occur.

## Acceptance criteria

- A test in `tests/test_checkup_review_loop.py` that stubs `_run_role_task` to fail for the primary reviewer and succeed for the fallback, and asserts the loop produces a verdict using the fallback's findings.
- A test asserting that with no `--reviewer-fallback`, behavior is unchanged (default policy preserved).
- A test asserting auth/network/exec failure strings classify as `degraded` not `failed`.

---

**Related:**
- #814 — credit-balance permanent-error fix that exposed this
- #739 — earlier review-loop reliability work

Serhan-Asad:
## Summary (reliability fix)

This is a reliability PR for the `pdd checkup --pr --review-loop` orchestrator, not a correctness change. It fixes a real $30+ retry failure on issue #814 / PR #914 where pdd-issue could not finalize an otherwise-correct PR because the pinned primary reviewer (`codex`) failed in pdd_cloud's executor and the loop had no escape hatch.

**Scope is initial primary-review failure only.** Verify-step failures are still subject to the existing dead-stop behavior — that path isn't observed in production and isn't changed here.

## Changes

- **`--reviewer-fallback TEXT`** — opt-in secondary reviewer, distinct from the fixer. When the **initial** primary review for a round lands in `HARD_NOT_CLEAN_STATES`, the loop invokes the fallback once. If the fallback returns a usable review, it takes over as primary for all subsequent rounds (fix + verify + later cycles); the original primary's status is preserved in `reviewer-status`. If the fallback also fails, the loop breaks with both statuses recorded.
- **Role-alias normalization** for the fallback. `--fixer claude --reviewer-fallback anthropic` previously slipped past the `fallback != fixer` gate because the raw strings differed even though both alias to the same role. Now normalized via `_normalize_reviewers` before comparison (`anthropic`→`claude`, `openai`/`chatgpt`→`codex`, `google`→`gemini`).
- **Narrower transient-failure classification** in `_failure_status`. Specific multi-word phrases (`authentication failed`, `connection refused`, `permission denied`, `failed to create sandbox`, …) + a non-zero-exit-code regex. Avoids the substring false-positives that would have come from naive markers (`"auth"` → `"author"`, `"login"` → `"logging"`). Still gated on `--continue-on-reviewer-limit`, so the default policy is unchanged.

## What this PR is NOT

- Not a verify-step fallback — only the **initial review** for a round triggers the fallback path. The verify call later in the same loop body, and any subsequent rounds' reviews using the (now-promoted) fallback, do not chain another fallback.
- Not a default-behavior change — without `--reviewer-fallback`, behavior is byte-identical to current `main`. Without `--continue-on-reviewer-limit`, the broadened markers do not fire.

## Why a fallback distinct from the fixer

The fixer (default: `claude`) cannot be auto-promoted to reviewer — that would defeat two-eye review (the fixer would be approving its own fixes). The structural fix has to be an opt-in role distinct from both primary reviewer and fixer.

## Test plan

- [x] `pytest tests/test_checkup_review_loop.py tests/commands/test_checkup.py -q` — 62 pass locally.
- [x] Includes both true-positive and true-negative cases for `_failure_status` (e.g. `"Author: Greg"` → `failed`, not `degraded`).
- [x] `test_reviewer_fallback_takes_over_subsequent_rounds` mutation-verified: removing the role-swap line makes the test fail.
- [x] `test_reviewer_fallback_provider_alias_of_fixer_is_ignored` + `..._of_primary_is_ignored` cover the normalization gates.
- [x] `test_no_reviewer_fallback_preserves_legacy_behavior` confirms default path unchanged.
- [ ] CI checks pass.
- [ ] Manual: run `pdd checkup --pr <pr> --issue <issue> --review-loop --reviewer-fallback gemini` against a PR where codex auth is broken — confirm gemini takes over and the loop reaches a verdict.

Closes #923


gltanaka (changes requested):
This fallback is needed: #923 documents a real review-loop gap where the primary reviewer is pinned through `PDD_AGENTIC_PROVIDER`, so the normal provider fallback path is bypassed and an infra/auth failure can dead-stop an otherwise valid PR. The opt-in fallback reviewer, distinct from the fixer, is the right shape.

Requesting changes before merge because the current PR still fails the repository whitespace check:

```
git diff --check origin/main...HEAD
context/pdd_completion_example.fish:17: new blank line at EOF.
context/pdd_completion_example.sh:20: new blank line at EOF.
context/pdd_completion_example.zsh:19: new blank line at EOF.
```

Please remove those trailing blank lines and rerun `git diff --check origin/main...HEAD`. I also ran the focused tests locally and they pass: `pytest tests/test_checkup_review_loop.py tests/commands/test_checkup.py tests/test_agentic_checkup.py -q` -> 102 passed.

gltanaka (approved):
The requested whitespace issue is fixed: `git diff --check origin/main...HEAD` now passes locally, and the focused review-loop/checkup tests still pass (`102 passed`). The fallback remains needed for #923 and the implementation is appropriately opt-in and distinct from the fixer.

Issue #865, opened 2026-05-08
pdd sync should repair architecture conformance failures instead of failing generically


Serhan-Asad:
## Problem

`pdd sync` can generate code that does not satisfy the public interface declared in `architecture.json`, then fail early instead of repairing the generated code.

This was seen while investigating #861. The failing module was `agentic_sync_runner`: `architecture.json` required `AsyncSyncRunner.run`, but the generated `pdd/agentic_sync_runner.py` did not expose that symbol, so architecture conformance failed and the GitHub App reported a generic sync failure.

## Why this matters

`architecture.json` is the contract between prompt/spec changes and generated code. If `pdd change` updates a prompt or architecture entry, `pdd sync` should either:

- generate code that satisfies the architecture contract, or
- retry/fix generation with the exact missing exports, or
- fail with an actionable error showing the missing symbols and repro command.

A generic `Sync operation failed` makes it look like a local setup problem and hides the real contract mismatch.

## Desired behavior

When architecture conformance fails, treat it as a repairable generation failure.

Example repair instruction passed into the retry/fix loop:

```text
The generated code does not satisfy architecture.json.

Required missing exports:
- AsyncSyncRunner.run

Add these exact exports. Do not modify architecture.json. Do not remove existing valid exports.
Regenerate until all architecture-required symbols exist.
```

## Proposed implementation

- Add a typed architecture conformance failure object/error that includes:
  - prompt filename
  - generated output path
  - matched architecture entry
  - expected symbols
  - found symbols
  - missing symbols
- In `pdd sync` / code generation, catch that failure and retry generation or run a focused repair step with the missing symbols.
- Limit retries, for example 2-3 attempts.
- Do not advance to example/test generation until architecture conformance passes.
- If retries fail, print a structured diagnostic block with the exact missing symbols and a local repro command.
- In the GitHub App path, avoid collapsing this into `Sync operation failed`; surface the conformance block in the issue/PR comment.
- Also log enough environment detail to rule out stale runner setup: `pdd.__file__`, `pdd --version`, git SHA, dirty status, and whether PDD is being imported from the checked-out repo or from site-packages.

## Files likely involved

- `pdd/code_generator_main.py`
- `pdd/agentic_sync_runner.py`
- `pdd/server/jobs.py`
- `pdd/prompts/code_generator_main_python.prompt`
- `pdd/prompts/agentic_sync_runner_python.prompt`
- `tests/test_code_generator_main.py`
[...]

PR #939, opened 2026-05-11
fix(code-gen): enforce <pdd-interface> signature in architecture conformance (#928)

### LINKED ISSUE #928: pdd sync: regen silently drops features when generated code violates the prompt's <pdd-interface> contract
## Observed
On promptdriven/pdd#871, the `pdd sync` run for `update_main` reported SUCCESS (5 phases, 1m 45s, \$2.82) but the regenerated `pdd/update_main.py` is missing the entire `--sync-metadata` feature that the prompt declares:

**Prompt (`pdd/prompts/update_main_python.prompt`) declares:**
- `<pdd-interface>` signature includes `sync_metadata=False` (lines 6-19).
- `<pdd-dependency>metadata_sync_python.prompt</pdd-dependency>` (line 22).
- Lines 41-46 spell out `run_metadata_sync` orchestration for single-file, regeneration, and repo modes.

**Regenerated `pdd/update_main.py` (line 1006):**
\`\`\`python
def update_main(
    ctx: click.Context,
    input_prompt_file: Optional[str],
    ...
    base_branch: str = "main",
    budget: Optional[float] = None,
    dry_run: bool = False,
) -> Optional[Tuple[str, float, str]]:
\`\`\`

`grep` for `sync_metadata|metadata_sync|run_metadata_sync` in `pdd/update_main.py` returns zero hits.

This caused 6 unit-test failures on PR promptdriven/pdd#920 (`test_update_command.py::TestUpdateCommandStrengthFallback`, `test_pddrc_true_e2e.py::TestPddrcUpdateCommandE2E`) — the sibling regen of `pdd/commands/modify.py` correctly added `--sync-metadata` and passes it as a kwarg, but the call into `update_main` raises `TypeError: unexpected keyword argument 'sync_metadata'`, which the CLI error handler swallows.

## Root cause
`pdd sync`'s success criterion does not include verifying that the regenerated code's public interface matches the prompt's `<pdd-interface>` declaration. Verifier phase only ran unit tests; the new tests for `sync_metadata` didn't exist yet (would have been written by a downstream regen of `tests/test_update.py`).

## Expected
After regen, `pdd sync` should:
1. Statically extract the public signature of all symbols declared in the prompt's `<pdd-interface>`.
2. Diff against the regenerated code's actual signature.
3. Fail the phase loudly when there is a mismatch (missing kwarg, missing function, type drift).

## Acceptance
- A test fixture where the prompt declares a kwarg the regen drops causes `pdd sync` to exit non-zero with an actionable error.
- The interface-contract check runs unconditionally after the code-generation phase.

Filed as part of the "make pdd-issue work end-to-end on #871" effort.

Serhan-Asad:
## Summary

Extends `pdd/code_generator_main.py::_verify_architecture_conformance` to enforce the prompt's `<pdd-interface>` signature contract — not just symbol existence — so `pdd sync` can no longer silently produce code missing a kwarg the prompt declares.

Closes #928.

## Root cause (verified)

On PR #920's branch, `pdd/prompts/update_main_python.prompt` declared `<pdd-interface>` with `sync_metadata=False`. `pdd sync` chose the `fix` command (per `.pdd/meta/update_main_python.json`'s `"command": "fix"`), produced `pdd/update_main.py` with **no** `sync_metadata` parameter, and reported SUCCESS in 5 phases. The existing conformance check only verifies that the symbol `update_main` exists — not that its declared kwargs are present — so the gap slipped through and 6 downstream unit tests failed with `TypeError: unexpected keyword argument 'sync_metadata'` (swallowed by the CLI error handler).

## Changes

- `pdd/code_generator_main.py`
  - Refactor: extracted the existing arch.json symbol-existence check into a private helper `_verify_architecture_json_conformance` (behavior unchanged); `_verify_architecture_conformance` is now a thin orchestrator that runs the arch.json check then the new `<pdd-interface>` check.
  - New: `_verify_pdd_interface_signatures` parses the prompt's `<pdd-interface>` JSON (via canonical `architecture_sync.parse_prompt_tags`), extracts declared parameter names for each declared function/CLI command (via `ast.parse` on the signature string — robust to defaults, type hints, kw-only markers, positional-only markers), AST-introspects the matching `FunctionDef` in the generated code, and raises `ArchitectureConformanceError` when a declared param is absent.
  - New optional `repair_directive` override on `ArchitectureConformanceError` so the new check can surface a `<pdd-interface>`-specific repair message.
  - `prompt_content: Optional[str] = None` added to `_verify_architecture_conformance` signature (default-compatible — existing callers unaffected).
- `pdd/prompts/code_generator_main_python.prompt` — one new sub-bullet under Requirement 5 declaring this check so future regen of `code_generator_main` stays consistent.
- `tests/test_code_generator_main.py` — new `TestPddInterfaceSignatureConformance` class with 9 tests: missing kwarg in function, missing kwarg in CLI command, present kwarg passes, `**kwargs` does not satisfy a declared named param, no `<pdd-interface>` skips silently, malformed JSON warns and skips, extra params in code do not raise, function absent from code is owned by the existing symbol check, and a regression test confirming `ArchitectureConformanceError` fields stay populated for the agentic_sync_runner repair loop.

## Robustness rules

- No `<pdd-interface>` in the prompt → new check is a silent no-op (preserves existing behavior).
- Malformed JSON inside `<pdd-interface>` → `logger.warning`, skip (don't break sync on bad metadata).
- `**kwargs` in code does **not** satisfy a named declared param — callers in the codebase pass these as named kwargs, so the catch-all is not a valid substitute.
- Extra params in generated code beyond what the prompt declares → fine, no raise (one-way: declared ⊆ actual).
- Class methods other than `__init__` are out of scope for this PR (functions + CLI commands cover the #928 case).

## Tests

- `pytest tests/test_code_generator_main.py` — 147 → 156 (9 new), all pass.
- `pytest tests/test_agentic_sync_runner.py tests/test_architecture_sync.py tests/test_sync_main.py tests/test_sync_orchestration.py` — 512 tests, all pass (no regressions).

## Independent review

Independently audited by a reviewer sub-agent. Sabotage-tested by flipping `if param in actual_params:` → `if param not in actual_params:` in the new check; 6/9 gate tests failed as expected, confirming the tests genuinely exercise the gate. Signature parsing also probed directly with 4 edge cases (positional-only, kw-only, varargs, type hints + defaults) — all correct. Verdict: clean, no required fixes.

## Replicating #928 with this fix

With the same setup that produced #928 (prompt declares `<pdd-interface>{"...sync_metadata=False"}`, code lacks the kwarg), `_verify_architecture_conformance` now raises:

```
ArchitectureConformanceError: Architecture conformance error for update_main_python.prompt: the prompt's <pdd-interface> declares parameter(s) missing from the generated code: update_main.sync_metadata.
missing_symbols: ['update_main.sync_metadata']
repair_directive: <full directive enumerating each missing param>
```

This routes through the existing repair-loop machinery added in PR #865, so `agentic_sync_runner` will surface the diagnostic and (where applicable) retry with the prompt's directive embedded.


gltanaka (changes requested):
This check is needed for #928, but I would not merge this PR as-is. The new <pdd-interface> gate currently fails against the repo prompt/code pairs, so the branch introduces an enforcement rule that the current tree does not satisfy.

I verified this by running the PR implementation of _verify_architecture_conformance across existing Python architecture entries twice: once with only architecture.json conformance and once with prompt_content enabled. The new <pdd-interface> pass adds these failures beyond the pre-existing architecture.json mismatches:

- agentic_common_python.prompt -> pdd/agentic_common.py: substitute_template_variables.variables is declared, but the implementation parameter is context.
- split_main_python.prompt -> pdd/split_main.py: split_main.legacy is declared, but split_main has no legacy parameter.
- core/cli_python.prompt -> pdd/core/cli.py: cli.context is declared, but the implementation parameter is context_override.
- core/duplicate_cli_guard_python.prompt -> pdd/core/duplicate_cli_guard.py: save_last_run.record is declared, but the implementation takes argv_tail and subcommand.

Because this PR makes the prompt the source of truth, it should either fix these prompt/code mismatches in the same PR or introduce an explicit staged rollout/allowlist so current modules are not broken by the new gate. Otherwise a future sync of any of these modules can fail purely because the repository already contains stale <pdd-interface> signatures.

The targeted PR tests passed locally: `conda run -n pdd pytest tests/test_code_generator_main.py::TestPddInterfaceSignatureConformance tests/test_agentic_sync_runner.py::TestSyncOneModule -q` -> 54 passed.

Serhan-Asad:
Addressed in 5c3b127f1: reconciled the four flagged prompt/code mismatches (`substitute_template_variables`, `split_main`, `cli`, `save_last_run`) — updated each `<pdd-interface>` declaration and matching architecture.json signature to reflect the current code. Swept every Python prompt/code pair in the repo through `_verify_architecture_conformance` with `prompt_content` enabled: zero remaining `<pdd-interface>` failures. The new gate is now mergeable against the current tree without a staged rollout or allowlist.

gltanaka (approved):
The previous blocker is resolved. I reran the repo-wide prompt/code conformance scan against the updated PR head with prompt_content enabled, and the new pdd-interface gate now introduces zero additional failures beyond the pre-existing architecture.json-only mismatches:

- architecture-only failures: 13
- full conformance failures with prompt_content: 13
- new/changed pdd-interface failures: 0

Local verification also passed:

- `conda run -n pdd pytest tests/test_code_generator_main.py::TestPddInterfaceSignatureConformance tests/test_agentic_sync_runner.py::TestSyncOneModule -q` -> 54 passed
- `conda run -n pdd pytest tests/test_code_generator_main.py tests/test_agentic_sync_runner.py -q` -> 303 passed

GitHub checks are green and the PR is mergeable. One non-blocking follow-up: a couple prompt prose references still mention old names/signatures even though the enforced pdd-interface contracts are now correct; that can be cleaned separately.

PR #945, opened 2026-05-12
chore: PDD sync for #814

### LINKED ISSUE #814: agentic_common: "Credit balance is too low" 400s should be classified permanent, not retried

Serhan-Asad:
Replica of #914 opened under @Serhan-Asad so the `heal` job authorizes (the bot-authored original failed the `pdd_cloud collaborator` gate).

## Summary

Extends `prompts/agentic_common_python.prompt` requirement 14 so `_is_permanent_error` classifies Anthropic-style billing/credit-exhaustion 400s as permanent. This stops the orchestrator from burning all retries on a provider whose response is clearly non-recoverable and lets `PDD_AGENTIC_PROVIDER` fall back to the next provider immediately.

Closes #814

### Prompts Modified
- `pdd/prompts/agentic_common_python.prompt` — Requirement 14 now lists billing/credit exhaustion alongside auth/quota as permanent, and calls out the three new case-insensitive substrings to match (`credit balance is too low`, `plans & billing`, `insufficient credit`/`balance`/`funds`). Cites Issue #814 inline so the rationale survives future regenerations.

### Other Files
- `architecture.json` — trailing-newline normalization only (no semantic change).

## Review Checklist

- [ ] Prompt syntax is valid
- [ ] PDD conventions followed
- [ ] Documentation is up to date

## Next Steps After Merge

1. Regenerate code from modified prompts in dependency order:
   ```bash
   ./sync_order.sh
   ```
   Or manually run the `pdd sync` chain starting with `pdd sync agentic_common` (see the workflow's sync order script for the full list).
2. Run tests to verify functionality (`pytest -vv tests/`).
3. Confirm a forced 400 "Credit balance is too low" response now advances to the next provider on the first failure rather than after retry exhaustion.

---
Supersedes #914 (bot-authored, auto-heal authorization blocked).


gltanaka (changes requested):
I think the behavioral fix is needed: current `main` does not classify the Anthropic `Credit balance is too low` 400 body as permanent, so #814's retry/fallback problem is real. I do not think this branch should merge as-is, though.

Findings:

1. Blocker: `.pdd/meta/agentic_common_python.json:5` is stale. On this branch, `calculate_current_hashes()` returns current composite `prompt_hash=e07d14a8c3ebac93bc4e9f085143a45445a2c43912dd0fc35ce8115db0953cc4`, but the fingerprint stores `ffe91baeea34d6398466dc56d2ed9fc9f21da29c7d5d0fe65d7261b15b3c1a3b`. Running `sync_determine_operation('agentic_common', 'python', prompts_dir='pdd/prompts')` immediately returns `auto-deps` with `Prompt changed - regenerating`, so the PR would land already dirty and trigger another regeneration cycle.

2. Blocker: `.pdd/meta/agentic_common_python_run.json:7` is also stale. The run report records `test_hash=69988552...` for `tests/test_agentic_common.py`, but the final test file and fingerprint use `4d3ea9bc...` after the later Rich-markup/quiet-mode test additions. The run report therefore does not validate the final branch state and will be rejected by the existing stale-run-report checks.

3. Blocker: `project_dependencies.csv:1` and the rest of the file were rewritten with CRLF line endings. `git diff --check origin/main...origin/pr-945` fails with trailing-whitespace diagnostics for all 280 CSV lines, and the diff is dominated by line-ending churn. Please normalize this file back to LF and keep only the semantic dependency updates needed for this PR.

Focused verification I ran: `conda run -n pdd pytest tests/test_agentic_common.py -k 'Issue814 or billing or permanent_error_diagnostic or anthropic_is_error_json_envelope' -q` passes (`12 passed, 255 deselected`). The code path looks directionally correct; the merge blockers are the stale generated metadata/run report and the CRLF CSV rewrite.

Serhan-Asad:
Addressed @gltanaka's three blockers in cc8d34446:

1. `.pdd/meta/agentic_common_python.json` — refreshed `prompt_hash` to the current composite `e07d14a8c3ebac93bc4e9f085143a45445a2c43912dd0fc35ce8115db0953cc4`. `sync_determine_operation('agentic_common', 'python', prompts_dir='pdd/prompts', target_coverage=90.0)` now returns `operation=nothing`, `reason="All required files synchronized"`.
2. `.pdd/meta/agentic_common_python_run.json` — refreshed `test_hash` and the `test_files["test_agentic_common.py"]` entry to the current content hash `4d3ea9bc77286259c0c698d89f5eb9b09d6cf3192f97cd6385b660ac5ba5cb1c`, bumped `tests_passed` to the verified 379 (`pytest tests/test_agentic_common.py tests/test_agentic_common_issue_813_*.py tests/test_agentic_common_worktree.py → 379 passed in 2.94s`). Run report is no longer stale relative to the final test files.
3. `project_dependencies.csv` — reverted to `upstream/main` verbatim. LF restored, the 280-line CRLF rewrite is gone, and the unrelated dependency drift is dropped since this PR is prompt-only and does not need CSV updates.

`git diff --check upstream/main...HEAD -- project_dependencies.csv` is now clean.

gltanaka (approved):
The previous blockers look resolved. I rechecked the updated head:

- `git diff --check origin/main...origin/pr-945` is clean.
- `project_dependencies.csv` is back to LF line endings with only semantic row updates.
- `agentic_common` fingerprint and run report now match the current files; `sync_determine_operation('agentic_common', 'python', prompts_dir='pdd/prompts', target_coverage=90)` returns `nothing`.
- The touched local tests pass: `conda run -n pdd pytest tests/test_agentic_common.py tests/test_preprocess.py -q` → `402 passed`.
- GitHub checks are green.

The #814 behavior is still needed, and the updated classifier is better than the prior version because it avoids misclassifying transient 429 / Plans & Billing hints as permanent while keeping hard quota/credit failures fail-fast. This now looks reasonable to merge.

Issue #941, opened 2026-05-11
auto-heal Cloud Build: pdd verify hangs on device-flow auth (keyring missing in runner image)


Serhan-Asad:
## Summary

The `auto-heal` workflow's Cloud Build step could hang because PDD fell back to **interactive GitHub device-flow authentication** inside a headless Cloud Build container. When the runner had no usable JWT cache or keyring refresh token, the CLI printed a GitHub verification URL/code and waited for a human who could not interact with the build.

Observed on PR #920, Cloud Build `dfea0df3-926c-46c2-9401-8d40e3db42eb` (heal run [`25702322008`](https://github.com/promptdriven/pdd/actions/runs/25702322008)). This was not PR-specific; any auto-heal run that reached a cloud-authenticated path without cached credentials could hit the same wall.

## Evidence From Original Failure

```text
Step #1 - "auto-heal-public-pdd-pr": ✗ Healing update_main (operation=verify) timed out
Step #1 - "auto-heal-public-pdd-pr": To authenticate, visit: https://github.com/login/device
Step #1 - "auto-heal-public-pdd-pr": Enter code: 53B3-291F
Step #1 - "auto-heal-public-pdd-pr": Opening browser for authentication...
Step #1 - "auto-heal-public-pdd-pr": Waiting for authentication...
```

```text
WARNING: Failed to retrieve refresh token from keyring:
No recommended backend was available. Install a recommended 3rd party
backend package; or, install the keyrings.alt package if you want to use
the non-recommended backends.
```

## Root Cause

`pdd verify` can enter a code path that needs PDD cloud authentication. In Cloud Build, keyring access is unavailable, so refresh-token lookup fails and the old behavior fell through to GitHub device-flow OAuth. Device flow is appropriate for a human terminal, but unsafe in CI because it waits for manual verification until timeout.

## Resolution

Implemented in PR #949: https://github.com/promptdriven/pdd/pull/949

- `pdd.get_jwt_token.get_jwt_token()` now refuses device-flow auth when `PDD_NO_INTERACTIVE` or `CI` is truthy.
- The guard runs after `PDD_JWT_TOKEN`, JWT cache, and keyring refresh paths, so non-interactive callers with valid cached credentials still authenticate silently.
- `pdd.ci_drift_heal._build_ci_env()` now sets `PDD_NO_INTERACTIVE=1` for auto-heal subprocesses.
- The generated prompt spec and tests were updated so future PDD maintenance preserves the env contract.
[...]

PR #946, opened 2026-05-12
ci(auto-heal): `/heal` comment actively clears `heal` check-run on PR head

Serhan-Asad:
## Summary

Adds a "Mirror conclusion to `heal` check-run on PR head" step to `.github/workflows/auto-heal.yml` that, on every `issue_comment` `/heal` run, posts a fresh `heal` check-run on the PR head SHA with the workflow's final conclusion. Result: commenting `/heal` actively clears (or refreshes) the red `heal` check-run on the PR, matching the operator mental model.

## Problem

Today the `heal` check-run on a PR head SHA reflects the conclusion of the most recent `pull_request_target` run for that SHA. That breaks two operator flows:

1. **Bot-authored PRs** (e.g. PRs opened by `prompt-driven-github[bot]` — see PR #920): the auth-gate exits 1 because the bot is not a `pdd_cloud` collaborator, leaving a permanent `heal: failure` check-run on the PR head. The intended escape hatch — a collaborator commenting `/heal` — runs the `issue_comment` workflow in default-branch context, so GitHub Actions' auto-posted check-run for that run lands on `main`'s SHA, not the PR head. The PR head's `heal: failure` is left untouched even after a successful `/heal`.

2. **Cloud-Build flakes on human PRs** (#941 device-flow timeout): `heal: failure` sticks on the PR head even after `/heal` re-runs Cloud Build successfully, for the same reason.

## Fix

Add one step after `Finalize check_run` that explicitly creates a new `heal` check-run on the PR head SHA via `gh api` on `issue_comment` runs. GitHub's checks-API rollup uses the most recently updated check-run per `(name, app)` on a SHA, and `GITHUB_TOKEN` runs as the same `github-actions` app that auto-posts the original — so this fresh check-run supersedes the stale one. No new check-run name; branch-protection unchanged.

## Scope guards (matches existing patterns)

- `github.event_name == 'issue_comment'` — `pull_request_target` already gets a fresh GHA-auto-posted `heal` check-run on the PR head.
- `fork == 'false'` — fork PRs short-circuit to `neutral` earlier and never reach this step.
- `skip != 'true'` — `/healthcheck`, `/healz`, etc. don't trigger the refresh.
- `always()` — failures (auth-gate reject of commenter, Cloud Build fail) refresh too, so stale state never lingers.

Conclusion mapping mirrors `Finalize check_run`: only `success` → `success`; cancelled / timed-out / failed all collapse to `failure` (branch protection can't be tricked by a cancelled run).

## Test plan

- [ ] Workflow YAML valid (`python -c "import yaml; yaml.safe_load(open('.github/workflows/auto-heal.yml'))"` — verified locally before push).
- [ ] After merge: comment `/heal` on PR #920 (bot-authored) → previously-permanent `heal: failure` on its head SHA flips to `heal: success`.
- [ ] Comment `/heal` on a hypothetical human PR with a flaked `heal: failure` from a prior Cloud Build → also flips to `heal: success`.
- [ ] `/healthcheck` / `/healz` / non-`/heal` comments do NOT post a new `heal` check-run (covered by existing `skip` gate).
- [ ] Fork PRs continue to short-circuit to `neutral` and don't reach the new step.

## Why a separate PR

`pull_request_target` workflows execute from `main`, not from the PR head (see the file's own line 23-25 security comment). A workflow fix shipped inside a bot-authored PR like PR #920 cannot self-apply for its own checks. Splitting this out means it can land on `main` first, after which `/heal` on existing bot PRs will clear their stale red.

## Out of scope

- The underlying `pdd_cloud` collaborator auth-gate behavior (unchanged).
- The #941 Cloud Build device-flow timeout (separate infra issue).
- README — the `auto-heal.yml` workflow has no dedicated README section; rationale is captured in a 25-line inline comment on the new step.


gltanaka (changes requested):
Blocking: I do not think this should merge as-is.

The required status check on `main` is currently `auto-heal` only (`required_status_checks.contexts=["auto-heal"]`), and this workflow already creates/updates that explicit `auto-heal` check-run on the PR head for both `pull_request_target` and `/heal` `issue_comment` runs (`Post check_run` + `Finalize check_run`). The new step at `.github/workflows/auto-heal.yml:366` therefore does not unblock branch protection; it only synthesizes a second `heal` check-run to overwrite a non-required Actions job signal.

That is risky because the synthetic check-run uses the exact same name/app as the real GitHub Actions job (`heal`). It makes the PR check list less trustworthy by allowing a manually created Checks API result to supersede the real job-level check in GitHub's rollup. If `heal` is ever made required, or if tooling reads checks by name, this can satisfy the context with a synthetic check instead of the actual Actions job, while the existing `auto-heal` context is already the supported source of truth.

Please drop this workflow change unless there is a concrete required-check or automation consumer that actually needs `heal` to be green. If the goal is operator clarity only, prefer documenting that `auto-heal` is the required gate or adjusting the naming/source-of-truth model rather than shadowing the Actions job check.

Serhan-Asad:
Audit 2026-07-21: closing this PR as superseded by merged PR #943. This head still has Changes Requested and is no longer a current-main delivery path; its branch and review history remain recoverable here.

Issue #1017, opened 2026-05-14
Add explicit global sync architecture scan policy for examples and fixtures


Serhan-Asad:
## Problem
No-argument global sync now correctly preserves nested architecture scope, but root scans currently discover architecture.json files under examples and tests fixtures, including:

- `examples/arch/architecture.json`
- `examples/prompts_linter/architecture.json`
- `examples/template_example/architecture.json`
- `tests/fixtures/arch_include_validate_ok/architecture.json`

Real example workspaces can be syncable and may need committed `.pdd/meta`, but docs-only demos and fixture architectures produce "no syncable prompt file found" warnings.

## Desired behavior
Define an explicit scan policy:

- Include nested directories with `.pddrc` and real prompt/code/test layouts.
- Exclude docs-only examples and test fixtures by default.
- Optionally allow explicit include/exclude patterns in `.pddrc` or CLI flags.
- Keep `find_architecture_for_project` behavior predictable.

## Context
The resync PR chose to keep nested `.pdd/meta` for real example workspaces and ignore `**/.pdd/core_dumps/*`.

Issue #1016, opened 2026-05-14
Reduce no-argument global sync dry-run noise for non-Tier-1 states


Serhan-Asad:
## Problem
After the PDD resync cleanup, `pdd sync --dry-run --skip-tests --skip-verify` correctly reports `0 stale module(s)`, but still prints many warnings for modules requiring non-Tier-1 operations (`example`, `verify`, `test`, `update`) and for sample/fixture architecture entries with no syncable prompt.

That makes the success signal harder to read and can make operators think the repo is still stale even when Tier 1 prompt/code sync is clean.

## Desired behavior
- Keep Tier 1 global sync focused on `generate` / `auto-deps`.
- Summarize out-of-scope states in a compact section.
- Consider separate flags/modes for Tier 2 audits (`--include-examples`, `--include-tests`, `--include-verify`, or `pdd sync-audit`).
- Keep "0 stale modules" visually unambiguous.

## Evidence
In the resync sandbox, final dry-run reported:

```
Global sync analysis: 0 stale module(s), 177 already synced or skipped.
```

It then followed with many "outside Tier 1 prompt-staleness scope" warnings.

Serhan-Asad:
5/15/2026

PR #1018, opened 2026-05-14
chore: resync PDD metadata

Serhan-Asad:
Splits the generated/state cleanup out from the sync tooling fixes in #1019.\n\nIncludes:\n- refresh root .pdd/meta fingerprints for current trusted files\n- add nested example metadata under examples/prompts_linter and examples/template_example\n- move example-only files from examples/ into context/\n- update prompts_linter architecture metadata\n\nLocal checks:\n- PYTHONPATH=. python -m pdd checkup --validate-arch-includes\n- git diff --check\n- python -m py_compile moved context examples\n\nRelated follow-up issues: #1016, #1017.


Serhan-Asad:
Closing because this 299-file generated-metadata snapshot predates current main and cannot be safely replayed; ownership and profile evidence must be generated from current main.

PR #1043, opened 2026-05-17
fix(cloud-batch/submit.sh): clean up gcloud's leaked multiprocessing workers

gltanaka:
## Summary
- `gcloud storage cp` and `gcloud batch jobs submit` leak ~18 Python multiprocessing workers on exit (re-parented to init, idle at 0% CPU). They hold the original stdout pipe open, hanging any shell pipeline that wraps `make cloud-test | tee log` long after the script has exited 0.
- Source confirmed: system Python `gcloud` SDK uses (`/Library/.../Python.app/...`), not any venv; pattern is gcloud's internal parallel-upload pool.
- Add a surgical EXIT trap in `submit.sh` that snapshots `multiprocessing.spawn|resource_tracker` PIDs at script start, then on exit kills only newly-spawned ones that are now re-parented to init. TERM -> 1s -> KILL.

## What's NOT touched
- No change to gcloud invocations themselves (parallel uploads still happen).
- No change to the temp-dir cleanup trap (composed with the new cleanup).
- No new dependencies.

## Test plan
- [x] `bash -n ci/cloud-batch/submit.sh` clean.
- [ ] CI green.
- [ ] Maintainer manual smoke: run `make cloud-test | tee /tmp/log` to completion; afterward `pgrep -f 'multiprocessing.(spawn|resource_tracker)' | wc -l` returns 0 (or at least the count present before the run).

### DIFF
--- a/ci/cloud-batch/submit.sh
+++ b/ci/cloud-batch/submit.sh
@@ -33,21 +33,21 @@ trap 'cleanup_leaked_gcloud_workers' EXIT
 # Sweep gcloud's leaked multiprocessing workers (re-parented to init) that this script spawned.
 cleanup_leaked_gcloud_workers() {
     [ -z "${_pre_gcloud_workers+x}" ] && return 0
-    local now_workers new_workers ppid pid term_pids=""
+    local now_workers new_workers ppid pid
+    local -a term_pids=()
     now_workers="$(pgrep -f 'multiprocessing\.spawn|multiprocessing\.resource_tracker' 2>/dev/null | sort -u || true)"
     [ -z "${now_workers}" ] && return 0
     new_workers="$(comm -13 <(printf '%s\n' "${_pre_gcloud_workers}") <(printf '%s\n' "${now_workers}") 2>/dev/null || true)"
     [ -z "${new_workers}" ] && return 0
-    for pid in ${new_workers}; do
+    while IFS= read -r pid; do
+        [ -z "${pid}" ] && continue
         ppid="$(ps -o ppid= -p "${pid}" 2>/dev/null | tr -d ' ' || true)"
-        [ "${ppid}" = "1" ] && term_pids="${term_pids} ${pid}"
-    done
-    [ -z "${term_pids}" ] && return 0
-    # shellcheck disable=SC2086 # intentional word-splitting of PID list
-    kill -TERM ${term_pids} 2>/dev/null || true
+        [ "${ppid}" = "1" ] && term_pids+=("${pid}")
+    done < <(printf '%s\n' "${new_workers}")
+    [ "${#term_pids[@]}" -eq 0 ] && return 0
+    kill -TERM "${term_pids[@]}" 2>/dev/null || true
     sleep 1
-    # shellcheck disable=SC2086 # intentional word-splitting of PID list
-    kill -KILL ${term_pids} 2>/dev/null || true
+    kill -KILL "${term_pids[@]}" 2>/dev/null || true
 }
 
 # ── Prepare source path allowlist ─────────────────────────────────────────
--- a/ci/cloud-batch/submit.sh
+++ b/ci/cloud-batch/submit.sh
@@ -26,6 +26,30 @@ _with_timeout() {
     return "$rc"
 }
 
+# Snapshot pre-existing multiprocessing PIDs so we can later identify gcloud-leaked workers.
+_pre_gcloud_workers="$(pgrep -f 'multiprocessing\.spawn|multiprocessing\.resource_tracker' 2>/dev/null | sort -u || true)"
+trap 'cleanup_leaked_gcloud_workers' EXIT
+
+# Sweep gcloud's leaked multiprocessing workers (re-parented to init) that this script spawned.
+cleanup_leaked_gcloud_workers() {
+    [ -z "${_pre_gcloud_workers+x}" ] && return 0
+    local now_workers new_workers ppid pid term_pids=""
+    now_workers="$(pgrep -f 'multiprocessing\.spawn|multiprocessing\.resource_tracker' 2>/dev/null | sort -u || true)"
+    [ -z "${now_workers}" ] && return 0
+    new_workers="$(comm -13 <(printf '%s\n' "${_pre_gcloud_workers}") <(printf '%s\n' "${now_workers}") 2>/dev/null || true)"
+    [ -z "${new_workers}" ] && return 0
+    for pid in ${new_workers}; do
+        ppid="$(ps -o ppid= -p "${pid}" 2>/dev/null | tr -d ' ' || true)"
+        [ "${ppid}" = "1" ] && term_pids="${term_pids} ${pid}"
+    done
+    [ -z "${term_pids}" ] && return 0
+    # shellcheck disable=SC2086 # intentional word-splitting of PID list
+    kill -TERM ${term_pids} 2>/dev/null || true
+    sleep 1
+    # shellcheck disable=SC2086 # intentional word-splitting of PID list
+    kill -KILL ${term_pids} 2>/dev/null || true
+}
+
 # ── Prepare source path allowlist ─────────────────────────────────────────
 cd "${REPO_ROOT}"
 SOURCE_PATHS=(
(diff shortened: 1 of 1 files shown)


gltanaka:
LGTM. Verified against the diff on disk:

- **Double-trap composition is intentional and correct.** The early trap (line 29) covers the `set -euo pipefail` early-exit window; the later trap (line 162) re-registers with both `rm -rf "${STREAMING_DIR}"` and `cleanup_leaked_gcloud_workers` to preserve temp-dir cleanup. Both code paths call the cleanup.
- **Surgical filtering.** Snapshot + `comm -13` + PPID==1 filter together limit the kill to processes that (a) didn't exist at script start and (b) are now re-parented to init. The residual risks — PID reuse over a multi-minute polling run, or collateral kill of an unrelated foreign orphan whose parent died mid-run — are low-probability and low-blast-radius on a dev workstation.
- **Portability.** `pgrep -f 'a|b'` ERE alternation works on Darwin BSD pgrep (empirically tested) and procps-ng on Linux. `comm` and `ps -o ppid= -p PID` are POSIX. `pgrep` excludes its own PID by default (macOS man: "the current pgrep or pkill process and all of its ancestors are excluded").
- **No-op behavior.** All early-return guards fire before any `kill`; the cleanup is silent when nothing leaked.
- **Caller surface.** Only caller is `Makefile:564` (`bash ci/cloud-batch/submit.sh`). No sourcing anywhere in the repo, so the EXIT trap won't fire on a parent shell.
- **Shellcheck.** SC2086 disables are tightly scoped to the two `kill` lines with explanatory comments for intentional word-splitting of the PID list.

One low-priority follow-up note (not a blocker): the 1s TERM→KILL window is fine for these specific leaked workers (no significant state), but if gcloud ever leaks workers that hold mutex locks, the window might need to grow.

gltanaka:
## Live validation

Exercised the new `cleanup_leaked_gcloud_workers` function with a synthetic test on macOS Darwin 25.3 — no Cloud Batch run needed.

**Setup:**
1. Pre-cleaned any stray `multiprocessing.spawn|resource_tracker` procs.
2. Spawned 2 orphan Python procs whose command line matches the pgrep pattern (PPID=1 via `( nohup … & )`).
3. Snapshotted `_pre_gcloud_workers` exactly as `submit.sh` does.
4. Spawned 3 more orphans (simulating gcloud's leak).
5. Sourced `cleanup_leaked_gcloud_workers` from this PR's `submit.sh`.
6. Ran it under `bash set -x`.

**Trace (excerpt):**
```
+ for pid in ${new_workers}
++ ps -o ppid= -p 3271
+ ppid=1
+ '[' 1 = 1 ']'
+ term_pids=' 3271'
… (repeats for 3273, 3275) …
+ kill -TERM 3271 3273 3275
+ sleep 1
+ kill -KILL 3271 3273 3275
```

**Assertions, all PASS:**
- ✅ PID 3271 (NEW) was killed
- ✅ PID 3273 (NEW) was killed
- ✅ PID 3275 (NEW) was killed
- ✅ PID 3252 (PRE) survived
- ✅ PID 3254 (PRE) survived

**Worth noting** (not a blocker — the shebang is correct): the function relies on bash word-splitting of unquoted `${new_workers}` in `for pid in ${new_workers}`. If anyone ever sources this function from zsh or another shell that doesn't word-split unquoted variables, it would silently iterate once with the multi-line PID string and no-op. With the current `#!/usr/bin/env bash` shebang and no other callers (`grep -rn submit.sh` returns only `Makefile:564 bash $(CLOUD_BATCH_DIR)/submit.sh`), this isn't reachable in production. Flagging in case the script ever gets reorganized.

gltanaka:
## Follow-up: hardened for non-bash shells (083994e52)

Addressed the zsh-fragility note from the live validation. The cleanup function no longer depends on bash word-splitting:

- `term_pids` is now a proper `local -a` array.
- Iteration switched from `for pid in ${new_workers}` to `while IFS= read -r pid; ... < <(printf '%s\n' "${new_workers}")`.
- Kill calls use quoted array expansion `"${term_pids[@]}"`. The two `# shellcheck disable=SC2086` comments are no longer needed and were removed.

**Verification under both shells** (synthetic test: 2 pre-existing + 3 post-snapshot orphan workers; assert exactly the 3 post-snapshot are killed):

| Shell | Result |
|---|---|
| bash 5.3.9 | PASS — `kill -TERM 8232 8234 8236` / `kill -KILL 8232 8234 8236`; pre-existing 8217 8219 survived |
| zsh 5.9 | PASS — `kill -TERM 8292 8294 8296` / `kill -KILL 8292 8294 8296`; pre-existing 8279 8281 survived |

Static checks clean:
- `bash -n ci/cloud-batch/submit.sh` — exit 0
- `shellcheck` — only pre-existing findings remain (SC2329 info on the trap-only function, SC2001 style on an untouched line)

Functionally identical under bash; defensive against future reorganization or sourcing.

gltanaka:
## Codex Round 2 review (083994e52)

LGTM on `083994e52`. The array-based rewrite is correct and safe. `local -a term_pids=()` is well-supported in bash (the only shell that matters given the shebang and the sole `bash submit.sh` caller in the Makefile). The early `[ -z "${new_workers}" ] && return 0` guard is unchanged and still fires before the loop, so the empty-input path has no regression. `printf '%s\n' "${new_workers}"` adds a trailing newline, and `IFS=` is scoped to the `read` command — no leak. `${#term_pids[@]}` is safe under `set -u` after `local -a term_pids=()`, and the array-length guard ensures `kill` is never reached with an empty array (and the `2>/dev/null || true` is a backstop if that guard ever didn't fire). Process substitution `< <(...)` runs in the current shell, so `term_pids+=` updates are visible after the loop — no subshell-swallows-variable bug. No new failure modes introduced.

gltanaka:
## Strengthening: dropped PPID==1 filter (fbb8506cb)

Live validation revealed a real gap with the previous version: gcloud's workers don't always re-parent to init by the time the EXIT trap fires — some pool parents stay alive past the gcloud-cli command exit and only die later. The PPID==1 filter silently no-op'd against those.

**What changed:**
- The PPID==1 check inside the iteration was removed.
- `term_pids` now accumulates every PID from `comm -13` (the snapshot-and-diff).
- The `ppid` local variable is no longer declared.

**Safety bound:** the pre-script snapshot is still the gate. `comm -13 pre now` only emits workers that didn't exist before the script started, so unrelated workers belonging to other pre-existing processes are never killed. The remaining risk — killing multiprocessing workers from a different process that started concurrently DURING this script's run — is narrow on a dev workstation and only affects script-time-bound concurrent invocations.

**Re-validation:**

| Test | Result |
|---|---|
| Synthetic, bash 5.3.9 (2 PRE + 3 NEW orphans) | PASS — PRE survived, NEW killed |
| Synthetic, zsh 5.9 (same setup) | PASS — PRE survived, NEW killed |
| Real `gcloud storage cp` smoke | this run gcloud didn't leak (timing-dependent), so trivially clean; previous run with PPID filter had 3 surviving workers whose parents were still alive — those would now be caught |
| `bash -n` | clean |
| `shellcheck` | only pre-existing SC2329 / SC2001, neither on changed lines |

gltanaka:
## End-to-end validation: the fix eliminates the leak

Ran a fresh `make test-pdd-cli-release` with this branch's `submit.sh` swapped into the local main-checkout (so the make's normal resolution path picks it up; the worktree path triggered an unrelated pdd-CLI Makefile-regeneration prompt).

**Setup:**
- Pre-cleaned all pre-existing `multiprocessing.spawn|resource_tracker` procs: count = **0**.
- `cp /private/tmp/pdd-fix-gcloud-leak-trap/ci/cloud-batch/submit.sh pdd/staging/public/pdd/ci/cloud-batch/submit.sh`.
- Verified swap: 3 matches for `cleanup_leaked_gcloud_workers` in the active file.
- Active gcloud account: `glt@promptdriven.ai`.

**Result:**

| | Session-start observation (no fix) | This run (fix applied) |
|---|---|---|
| Cloud Batch tasks | 77/77 passed | 77/77 passed |
| LLM smoke Cloud Build | SUCCESS | SUCCESS (2m18s) |
| `MAKE_EXIT` | 0 | 0 |
| **Leaked workers after make exit** | **18** | **0** |

```
$ pgrep -f 'multiprocessing\.spawn|multiprocessing\.resource_tracker' | wc -l
0
```

That's the test the previous synthetic + storage-cp smokes couldn't conclusively give: the trap **does** fire at the right moment to catch the workers gcloud actually leaks in the production code path. Confirms dropping the PPID==1 filter (commit `fbb8506cb`) was the right call.

Main checkout reverted to clean state after the test (`git checkout ci/cloud-batch/submit.sh`).

PR #1045, opened 2026-05-17
fix(cloud-batch): pre-create result file so SIGKILLed tasks report preempted

gltanaka:
## Summary
- SIGKILL bypasses traps, so SPOT-preempted Cloud Batch tasks left no `task_N.json`; the aggregator reported them as opaque `ERR / no result file`, conflating preemption with crashes/OOM/init failures.
- Pre-create `task_N.json` with `status: preempted` immediately after `TASK_INDEX` is known. Normal completion overwrites via `write_result`; SIGTERM overwrites via `term_handler`; SIGKILL leaves the pre-create in place (SIGKILL kills the process but not the file).
- Trap handlers' file-existence guard replaced by a `WROTE_FINAL_RESULT` flag — necessary because the pre-create defeats `[ ! -f ]`. Behavior on normal exit + caught SIGTERM is unchanged.

## Test plan
- [x] `bash -n ci/cloud-batch/entrypoint.sh`
- [x] Local SIGKILL simulation: pre-create persists with `status: "preempted"`
- [x] Local normal-exit simulation: `write_result` overwrites to `"passed"`
- [x] Local SIGTERM simulation: `term_handler` overwrites to `"error"/"terminated"`
- [ ] Observe next `make test-pdd-cli-release` run: preemption-induced tasks should now report `preempted` instead of `no result file`

### DIFF
--- a/ci/cloud-batch/entrypoint.sh
+++ b/ci/cloud-batch/entrypoint.sh
@@ -23,6 +23,31 @@ WORK_DIR="/workspace"
 RESULT_JSON="${RESULTS_DIR}/task_${TASK_INDEX}.json"
 RESULT_LOG="${RESULTS_DIR}/task_${TASK_INDEX}.log"
 
+# ── Pre-create result file so SPOT preemption is visible ──────────────────
+# Cloud Batch SPOT VMs receive SIGTERM then SIGKILL ~30s later when preempted.
+# SIGKILL bypasses traps, so if the test is mid-execution when SIGKILL fires
+# no task_N.json is ever written and the aggregator reports an opaque
+# "ERR / no result file" — indistinguishable from a real crash. Pre-creating
+# the file with status="preempted" means SIGKILL leaves a diagnosable artifact
+# behind: write_result overwrites on normal completion, term_handler
+# overwrites on SIGTERM, and only true preemption leaves the preempted marker.
+# Single heredoc keeps the write to one syscall (the streaming reader in
+# submit.sh skips files <10 bytes, so a partial pre-create won't poison
+# the aggregator if preempted mid-write).
+mkdir -p "${RESULTS_DIR}"
+WROTE_FINAL_RESULT=0
+cat > "${RESULT_JSON}" <<JSON
+{
+    "task_index": ${TASK_INDEX},
+    "suite": "unknown",
+    "detail": "task killed before completion (likely SPOT preemption)",
+    "status": "preempted",
+    "duration_seconds": 0,
+    "setup_seconds": 0,
+    "timestamp": "$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+}
+JSON
+
 # Task ranges
 PYTEST_START=0
 PYTEST_END=31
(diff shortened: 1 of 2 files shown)


gltanaka:
Round 1 blockers — both verified locally:

- **`ci/cloud-batch/submit.sh:185-209` (BLOCK)**: the streaming poller globs `task_*.json` from GCS every 10s, counts every file as `COMPLETED`, and reports any non-"passed" status as a `!! FAILURE`. With the pre-create, every running task has a `task_N.json` in GCS within seconds of starting, status=`preempted`. The poller would (a) report "77/77 complete" within a minute, masking real progress, and (b) emit a spurious `!! FAILURE: Task N (unknown / task killed before completion)` for every running task. Worse: it marks each file as `seen` (line 195-196), so when `write_result` later overwrites with `passed`, the poller skips it — real successes/failures never reach the streaming output. Fix: in the failure loop, skip `status="preempted"` until the job is terminal; and exclude preempted-status files from the COMPLETED count.

- **`ci/cloud-batch/entrypoint.sh:302-303` (BLOCK)**: the EXIT/TERM traps are installed only AFTER source extract (line 102), pip install (line 106), preflight (lines 114, 126), and JWT exchange (lines 160-264). Any `set -e` failure in those ~170 lines aborts before traps install — `WROTE_FINAL_RESULT` stays 0, no trap fires to update the pre-create, and the file persists as `status="preempted"`. A corrupt tarball or transient pip failure would be mis-reported as preemption. (Some setup paths already call `write_result` explicitly on failure — preflight, JWT for cloud_regression — but tar/pip don't.) Fix: move `trap_handler` and `term_handler` function defs to right after `write_result` (~line 97), move `START_TIME=$(date +%s)` and the two `trap` installations there as well, so any early-exit failure overwrites the pre-create with the real `error/exit_code_N` result.

— [Codex review, round 1 (via Claude Code)](https://github.com/promptdriven/pdd/pull/1045)

gltanaka:
Both Round 1 blockers addressed in 5eb0d843f:

1. **submit.sh poller**: now reads each `task_N.json`'s status and skips `preempted` entries — both in the `COMPLETED` count and in the failure-detection loop. Provisional markers are not marked "seen", so write_result's overwrite is picked up on the next poll cycle.
2. **entrypoint.sh trap installation**: `trap_handler` and `term_handler` definitions + `START_TIME` + the two `trap` lines moved to right after `write_result` (line ~106-127), before any setup work. A `set -e` abort from tar/pip/preflight now fires the trap and overwrites the pre-create with `error/exit_code_N`.

Verified locally with four simulations: sigkill → `preempted` ✓, sigterm → `error/terminated` ✓, normal exit → `passed` ✓, set-e setup failure → `error/exit_code_1` ✓ (this last one is the regression check for blocker #2). `bash -n` clean on both files.

gltanaka:
LGTM. Both round-1 blockers are fully resolved. The trap is now installed at `entrypoint.sh:126-127`, immediately after `WROTE_FINAL_RESULT=0` (line 38) and `START_TIME=$(date +%s)` (line 125) — both globals are initialized before any trap can fire, and `write_result` references `${SETUP_SECONDS:-0}` (line 88) so a trap during setup produces valid JSON under `set -u`. The remaining pre-create → trap window (lines 51–125) contains only pure variable assignments and function definitions with no I/O, so a signal there correctly leaves `status=preempted` as the intended fallback — and the window is materially narrower than round 1. The `submit.sh` preempted-skip logic is correct: `status=preempted` files are excluded from the COMPLETED count (line 196) and from the failure loop (line 213) without marking them `seen`, so a later tick picks up the real result once `write_result` overwrites. One non-blocking note: the COMPLETED loop at lines 192–198 spawns `python3` for every file before the `preempted` skip, giving ~77 invocations per 10s tick at full scale — acceptable for a long-running CI job, but worth revisiting if the file count grows significantly.

— [Codex review, round 2 (via Claude Code)](https://github.com/promptdriven/pdd/pull/1045)

PR #1047, opened 2026-05-17
feat(llm_invoke): add reasoning_type='adaptive' for Anthropic Opus 4.7

gltanaka:
## Summary
Anthropic's Claude Opus 4.7 dropped the legacy `thinking={"type":"enabled","budget_tokens":N}` payload and only accepts the new adaptive thinking API: `thinking={"type":"adaptive"}` plus top-level `output_config.effort`. Existing `reasoning_type=budget` and `reasoning_type=effort` branches both serialise to the legacy shape via LiteLLM, so every Opus 4.7 call 400s.

This PR adds a third branch, `reasoning_type=adaptive`, parallel to the existing `budget` / `effort` / `none` cases in `pdd/llm_invoke.py`. For Anthropic provider it sets:

```python
thinking         = {"type": "adaptive", "display": "summarized"}
reasoning_effort = time_to_effort_level(time)   # "low" | "medium" | "high"
```

LiteLLM 1.80+ maps `reasoning_effort` to `output_config.effort` server-side for direct Anthropic when the model matches `AnthropicConfig._is_claude_opus_4_5`. Until LiteLLM ships native matching for newer Opus models, downstream callers can monkey-patch that helper (documented inline in the new branch).

For non-Anthropic providers the branch logs a warning and skips — adaptive thinking is currently Anthropic-only.

## Smoke evidence (empirical, 2026-05-16, downstream consumer environment)
Direct Anthropic SDK and LiteLLM-via-monkey-patch both confirmed the wire payload reaches the API as:
```json
{"model":"claude-opus-4-7","max_tokens":...,"output_config":{"effort":"medium"},"thinking":{"type":"adaptive","display":"summarized"}}
```
with header `anthropic-beta: effort-2025-11-24` auto-injected by LiteLLM. 4/4 effort variants (low/medium/high, plus default) returned 200.

Caveat: adaptive mode on Opus 4.7 currently surfaces reasoning inline in the `text` content block rather than emitting a structured `thinking` content block, even with `display:"summarized"`. Callers reading `thinking_blocks` will see them empty — same observable behaviour as `display:"omitted"`. This is server-side adaptive behaviour, not a client-side defect.

## Tests
Two new tests under `TestReasoningParameters` in `tests/test_llm_invoke.py`:
- `test_adaptive_reasoning_anthropic` — asserts the thinking + reasoning_effort kwargs shape and that no legacy `budget_tokens` leaks through.
- `test_adaptive_reasoning_non_anthropic_provider_skipped` — asserts the warn-and-skip path.

All 5 `TestReasoningParameters` tests pass locally against `upstream/main` HEAD `9a07076ba`.

## Out of scope
- PyPI release / version bump — separate.
- Downstream consumer changes (monkey-patch `_is_claude_opus_4_5` to also match `opus-4-7`, flip CSV row to `reasoning_type=adaptive`, update guard test) — separate PR in the consumer repo after this lands and gets released.

## History
This is a re-do of a PR that was originally opened on `gltanaka/pdd#1431`, which has since been confirmed deprecated. Codex review of that prior PR posted an LGTM with three substantive claims verified (no external membership filter, branch sits inside `if time > 0` guard, central temperature-forcing logic at lines ~3691-3703 covers `thinking OR reasoning_effort`); the same code change ports cleanly here because the dispatch shape is identical between the two repos.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -252,6 +252,7 @@ yarn.lock
 !.pdd/meta/get_comment_python.json
 !.pdd/meta/get_test_command_python.json
 !.pdd/meta/git_update_python.json
+!.pdd/meta/llm_invoke_python.json
 !.pdd/meta/summarize_directory_python.json
 .pdd/meta/*_run.json
 # Issue #1006 follow-up: these two modules' tests pass quickly but
--- a/.pdd/meta/llm_invoke_python.json
+++ b/.pdd/meta/llm_invoke_python.json
@@ -1,0 +1,24 @@
+{
+  "pdd_version": "0.0.241",
+  "timestamp": "2026-05-17T18:33:34.290918+00:00",
+  "command": "example",
+  "prompt_hash": "08f20f2e33886688d14103186bc55760a745b23e9bbe2626a0a7a6c4e1c1528c",
+  "code_hash": "42370b9d59b0caca264991021fb612f02bdae906c8043451288a64697af894ae",
+  "example_hash": "48c5aece0ddd153f95ec8a53802d2173d4da12a920b398c5feebb307b9958417",
+  "test_hash": "630ae15410752b1bda2608d72ab104add66b1f9cbbd134c003183339b46353f9",
+  "test_files": {
+    "test_llm_invoke.py": "630ae15410752b1bda2608d72ab104add66b1f9cbbd134c003183339b46353f9",
+    "test_llm_invoke_csv_model_registration.py": "1583b5e076fe5227a11f3d1079034ab4a21bc6131a3433f37bd68e35a99ba49e",
+    "test_llm_invoke_integration.py": "2eb4bd2565761a4148762c6fb73c887cb6e12ff962742e05f5c2d4d62b47aaf9",
+    "test_llm_invoke_nested_schema.py": "c983a19874abacc3e0ea9d6ca2ec87495960d2dd96d4e93be4039ed1bc995b9b",
+    "test_llm_invoke_retry_cost.py": "bfdfe7b814b78813f86b8bc6d081831daf531187feff66358260f6282925958c",
+    "test_llm_invoke_vertex_retry.py": "eafe91ba9376dc7e2da36faf3cd2de3cef50f70cf1c4fc5655e373deb7f50c26"
+  },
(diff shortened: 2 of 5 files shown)


gltanaka:

LGTM for the adaptive reasoning change. The same three checks were verified in detail on the prior identical PR (`gltanaka/pdd#1431`, since closed as `gltanaka/pdd` is deprecated); the dispatch shape in this repo is identical and the verification carries over:

1. **External membership filters?** `git grep -n reasoning_type` across the repo (excluding `pdd/llm_invoke.py` and `tests/`) finds no code-level `if reasoning_type in (...)` filter that would silently skip `'adaptive'`. Hits are all in docs / CSV / prompt files.
2. **`time == 0` short-circuit?** The new `elif reasoning_type == 'adaptive':` sits inside the existing `if time > 0:` guard. Chain order: budget → effort → **adaptive** → none → else (unknown).
3. **Temperature-forcing parity?** The Claude `temperature=1` rule lives centrally and triggers on `'thinking' in litellm_kwargs OR 'reasoning_effort' in litellm_kwargs`. The new branch sets both, so it's covered without needing a per-branch `temperature=1` assignment.

No blockers found.

gltanaka:
Update: rebased onto post-#1050 main and pushed. First auto-heal attempt with the released 0.0.241 progressed past the symlink+check-ignore bug (both filter messages fired correctly) and reached the final verification step, where it failed with: \`metadata staging verification failed: missing .pdd/meta/llm_invoke_python.json\`.

Root cause: \`.gitignore\` line 238 (\`.pdd/meta/*.json\`) ignores all per-module fingerprint files by default, and the allowlist (\`!.pdd/meta/<module>_python.json\` from line 239) is curated per-module. \`llm_invoke_python.json\` was missing from the allowlist, so auto-heal's heal fingerprint got dropped by the gitignore filter before staging and the downstream verification step then complained.

Fixed in \`c6167737f\` by adding \`!.pdd/meta/llm_invoke_python.json\` to the allowlist alongside the other long-running modules (agentic_bug, fix_code_loop, summarize_directory, etc.). One-line, scoped fix.

Re-running auto-heal now.

PR #1048, opened 2026-05-17
fix(ci_drift_heal): skip symlink-traversal paths in git add

gltanaka:
## Summary
The auto-heal CI workflow currently fails on every PR that modifies `pdd/*.py` because of a symlink-staging bug. `auto-heal` is a required status check on `promptdriven/pdd:main`, so this blocks all code PRs from merging.

## Root cause
`_git_relative_path_candidates` in `pdd/ci_drift_heal.py` returns **both** path forms for a given source — the symlinked form (e.g. `prompts/foo.prompt` via the tracked `prompts -> pdd/prompts` shim) and the canonical form (`pdd/prompts/foo.prompt`). Both end up in the flat list passed to `_git_add_pathspecs`. Recent Git rejects the symlinked form with:
```
fatal: pathspec 'prompts/foo.prompt' is beyond a symbolic link
```
… which tanks the whole heal commit. Build step `auto-heal-public-pdd-pr` returns `FAILURE`; the GHA `auto-heal` check fails; merge is blocked.

Concrete example from the latest reproducer: build `6e2ac8b1-e7d0-45e1-bfae-51e206e1856b` (auto-heal for PR #1047) — heal generated the prompt cleanly (`✓ prompt: ok (24246 chars)`, `Auto-heal summary: healed=1 failed=0 skipped=0`), then crashed at `git add prompts/llm_invoke_python.prompt`.

## Fix
In `_git_add_pathspecs`, drop pathspecs whose ancestor directory is a symlink before invoking `git add`. The canonical (resolved) form is normally already present in the list (because `_git_relative_path_candidates` emits both), so the entry isn't lost — the canonical form just gets staged instead.

New helper `_has_symlinked_ancestor(rel, repo_root)` walks the path's ancestor chain under `repo_root` and returns True if any directory along the way is a symlink. Used as a filter alongside the existing `.gitignore` check.

## Tests
Added `TestSymlinkStaging` (3 cases) in `tests/test_ci_drift_heal.py`:
- `_has_symlinked_ancestor` correctly detects a symlinked ancestor directory.
- `_git_add_pathspecs` drops the symlinked form when both forms are passed, staging only the canonical form.
- `_git_add_pathspecs` is unchanged when no symlinks are in play.

All 3 new tests pass locally. 151/152 of the existing `tests/test_ci_drift_heal.py` suite pass; the one failure (`TestDetectDrift::test_pr_touched_metadata_modules_are_not_left_in_drift`) is **pre-existing test fragility unrelated to the fix** — that test runs `detect_drift` against the live working tree, which sees this PR's own edit to `pdd/ci_drift_heal.py` as drift (operation: `verify`). The same drift would auto-heal cleanly once this PR's fix is shipped and the published pdd-cli picks it up.

## Rollout
Auto-heal CI installs `pdd-cli` from PyPI (`pip install --quiet pdd-cli` in `scripts/ci/run_pdd_cli_auto_heal_pr.sh:120`), so this fix only takes effect for new PRs *after* a fresh pdd-cli release that includes it.

### DIFF
--- a/pdd/ci_drift_heal.py
+++ b/pdd/ci_drift_heal.py
@@ -18,7 +18,7 @@ import subprocess
 import sys
 import tempfile
 from dataclasses import dataclass
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import Any, Dict, List, Optional, Sequence, Set, Tuple
 
 from rich.console import Console
@@ -918,6 +918,30 @@ def _healed_module_file_relpaths(module: Any, repo_root: Path) -> List[str]:
     return relpaths
 
 
+def _has_symlinked_ancestor(rel: str, repo_root: Path) -> bool:
+    """Return True if any directory ancestor of ``rel`` under ``repo_root`` is a symlink.
+
+    ``git add`` refuses pathspecs whose ancestor directory is a symlink ("pathspec
+    'X' is beyond a symbolic link"). When ``_git_relative_path_candidates`` returns
+    both the symlinked form (e.g. ``prompts/foo.prompt`` via ``prompts -> pdd/prompts``)
+    and the canonical form (``pdd/prompts/foo.prompt``) for the same source path,
+    the symlinked form will fail staging. Dropping it lets the canonical form get
+    staged instead.
+    """
+    parts = PurePosixPath(rel).parts
+    if len(parts) <= 1:
+        return False
+    cur = repo_root
+    for part in parts[:-1]:
+        cur = cur / part
+        try:
+            if cur.is_symlink():
+                return True
+        except OSError:
+            return False
+    return False
+
+
 def _git_add_pathspecs(pathspecs: Sequence[str], cwd: Optional[Path] = None) -> bool:
     """Stage existing pathspecs with an explicit git-add call.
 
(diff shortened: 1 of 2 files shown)


gltanaka:

LGTM with one non-blocking suggestion that's now applied. Independently verified the invariant Codex flagged:

**Claim:** Dropping the symlinked form from the staging list is safe only if `_git_relative_path_candidates` always emits the canonical form alongside it.

**Verification of `pdd/ci_drift_heal.py:_git_relative_path_candidates`:**
- **Absolute input via symlink** (the auto-heal case): branch 4 emits the symlinked form (`p.relative_to(repo_root)`); branch 5 emits the canonical form (`p.resolve().relative_to(repo_root.resolve())`). Both present.
- **Relative input via symlink**: branch 1 emits the symlinked form; branch 2 emits the canonical form (`(repo_root / p).resolve().relative_to(repo_root.resolve())`); branch 3 emits the symlinked form again (deduped via `set`). Both forms present.
- **Non-symlinked input**: all branches collapse to the same path; one entry in the set.

So the invariant holds for every input shape the function accepts. The filter dropping symlinked-ancestor paths cannot silently miss a file unless that function changes.

Applied Codex's suggestion in commit `f1ffb3025` — added a docstring comment in `_git_add_pathspecs` documenting the invariant so a future reader knows why dropping the symlinked form is safe and where to look (the helper) if it ever becomes unsafe.

Note on Codex's review process this round: the helper's sandbox couldn't refresh its auth, so its review was based on the PR description rather than reading source. I independently re-read the diff to confirm correctness rather than relying on Codex's analysis alone.

PR #1050, opened 2026-05-17
fix(ci_drift_heal): symlink filter must run before git check-ignore (#1048 follow-up)

gltanaka:
## Summary
Follow-up to #1048 addressing a post-merge review that proved the first fix was incomplete. **The symlink filter ran AFTER `git check-ignore`, but check-ignore *also* fatals on symlinked-ancestor pathspecs** — so a mixed batch containing both a symlinked path and a gitignored path would still crash the heal commit at `git add`.

## What the reviewer caught (and reproduced)
Mixed batch of `['prompts/foo.prompt', 'pdd/prompts/foo.prompt', 'pdd/meta/foo_run.json']` in a repo with `prompts -> pdd/prompts` symlink and `*_run.json` in `.gitignore`:

| step | before this PR (i.e. #1048 alone) |
|---|---|
| 1. `git check-ignore -- prompts/foo.prompt pdd/prompts/foo.prompt pdd/meta/foo_run.json` | exits 128: `pathspec 'prompts/foo.prompt' is beyond a symbolic link` |
| 2. helper's exception handling | silently sets `ignored = set()` (rc_code=128 doesn't match `in (0,1)`) |
| 3. symlink filter | drops `prompts/foo.prompt` |
| 4. `git add` receives `pdd/prompts/foo.prompt` AND `pdd/meta/foo_run.json` | refuses: "The following paths are ignored by one of your .gitignore files" |
| **5. helper returns** | **False** |

So heal commits still crashed in production despite #1048.

## Fix
Move the symlink filter to run **first**, before any git invocation. `check-ignore` then sees a symlink-free list and works correctly. The two filters now run as:

1. **Symlink-traversal filter** (new position): drops pathspecs whose ancestor directory is a symlink.
2. **Gitignore filter** (unchanged): now runs `check-ignore` on symlink-free survivors.

Same reproducer after this PR:
```
ci-heal: skipping symlink-traversal paths during stage: ['prompts/foo.prompt']
ci-heal: skipping gitignored paths during stage: ['pdd/meta/foo_run.json']
helper returned ok=True
git status --porcelain:
  A  pdd/prompts/foo.prompt
```

Also adds a defensive warning when every input path gets filtered out (claim 6 from the reviewer's list — silent-drop risk if a future caller passes only the symlinked form with no canonical twin).

## Tests
New `TestSymlinkStagingRealGit` class with three **real-git integration** tests (no `subprocess.run` mocking — spin up a tmp repo, init it, run real git). The first test is the reviewer's exact mixed-batch reproducer; would fail against #1048 alone, passes against this PR.

The original `TestSymlinkStaging` (mocked) suite is retained because it documents the argv shape, but the new suite is what actually proves the bug is fixed.

Also fixed a pre-existing test fragility in `test_pr_touched_metadata_modules_are_not_left_in_drift` — it called `detect_drift` against the live working tree, so any in-flight edit to one of the three listed long-running modules would break it. Now mocks `sync_determine_operation` to make the test hermetic, matching the pattern used by the rest of `TestDetectDrift`.

Final result: **155/155 tests pass locally**, including the previously-fragile one and the three new real-git integration tests.

## Acknowledgment
The reviewer of #1048 was right on every blocking point. Each of the six claims they raised was independently verified:
1. ✅ check-ignore fatals on symlinks too — this PR fixes that
2. ✅ Tests were mocking away the real failure — this PR adds real-git tests
3. ✅ No regression test for the `.gitignore` + symlink combo — `test_mixed_batch_with_symlink_and_ignored_path` covers it
4. ✅ #1048 was merged with red CI via admin-bypass — not doing that here; CI must actually pass
5. ✅ Deployed auto-heal still uses the published pdd-cli — both fixes need to ship in the next release
6. ✅ Silent-drop risk if only the symlinked form is supplied — new yellow warning makes it observable

### DIFF
--- a/pdd/ci_drift_heal.py
+++ b/pdd/ci_drift_heal.py
@@ -945,26 +945,63 @@ def _has_symlinked_ancestor(rel: str, repo_root: Path) -> bool:
 def _git_add_pathspecs(pathspecs: Sequence[str], cwd: Optional[Path] = None) -> bool:
     """Stage existing pathspecs with an explicit git-add call.
 
-    Filters out paths that ``.gitignore`` excludes before invoking ``git add``
-    — otherwise Git refuses the whole batch with "paths are ignored", aborting
-    an otherwise-successful heal commit. Examples include per-run reports
-    under ``.pdd/meta/*_run.json``, which the repo intentionally ignores but
-    which appear in the healed module's metadata pathspecs.
+    Two filters run **before** ``git add`` and **before** ``git check-ignore``:
 
-    Also drops pathspecs whose ancestor directory is a symlink (e.g. the
-    repo-root ``prompts -> pdd/prompts`` shim). Git refuses such paths with
-    "beyond a symbolic link"; the canonical form is normally also present in
-    the list because ``_git_relative_path_candidates`` returns both shapes.
+    1. **Symlink-traversal filter.** Drops pathspecs whose ancestor directory is
+       a symlink (e.g. the repo-root ``prompts -> pdd/prompts`` shim). Recent
+       Git fatals on such paths with "pathspec 'X' is beyond a symbolic link" —
+       and that fatal hits ``git check-ignore`` first, not just ``git add``, so
+       the symlink filter MUST run before any git invocation. The canonical
+       form is normally also present in the list because
+       ``_git_relative_path_candidates`` returns both shapes.
+
+    2. **Gitignore filter.** Run ``git check-ignore`` on the symlink-free
+       survivors so we know which paths to skip — otherwise ``git add`` refuses
+       the whole batch with "paths are ignored" (e.g. per-run reports under
+       ``.pdd/meta/*_run.json``), aborting an otherwise-successful heal commit.
+
+    Returns True on success or on a "nothing to stage" no-op. Returns False
+    only if ``git add`` itself fails after filters have been applied. Emits a
+    warning (not a hard failure) if all input paths get filtered out, since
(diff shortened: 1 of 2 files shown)


gltanaka:

LGTM. Independently verified the two structural claims:

1. **Symlink filter is unconditionally before any `subprocess.run`.** In `pdd/ci_drift_heal.py:_git_add_pathspecs`, `symlinked` is computed at line offset +38 from the function header and `survivors` at +44; the first `subprocess.run` (check-ignore) is at +59, the second (`git add`) at +89. So check-ignore now operates only on symlink-free survivors — addressing the reviewer's claim 1.

2. **`TestSymlinkStagingRealGit` has zero `subprocess.run` patching.** `grep` confirms no `patch.*subprocess` or `patch.*ci_drift_heal` inside that class — the three integration tests actually invoke real git in a tmp repo. The first test (`test_mixed_batch_with_symlink_and_ignored_path`) is the reviewer's exact reproducer; it returns `ok=True` against this PR and `ok=False` against #1048 alone.

3. **`test_pr_touched_metadata_modules_are_not_left_in_drift` fix matches the rest of `TestDetectDrift`.** Same `discover_prompt_files` / `infer_module_identity` / `sync_determine_operation` patch shape as the other tests in the class. Fragility against the developer's live working tree is gone.

No new failure modes introduced — non-symlinked paths flow through the same check-ignore/`git add` path as before, just with the symlink-filter step preceded.

PR #1052, opened 2026-05-17
feat(release): migrate to setuptools-scm; one-command make release

gltanaka:
## Summary

Replace the bump-branch + bump-PR + tag-push release flow with a single `make release` command. Version is derived from the latest `v*` git tag via setuptools-scm — no static version in `pyproject.toml`, no bump commit, no separate bump PR or CI cycle.

After this lands, the entire release flow is:

```
make release    # on main: tag HEAD with next vN.N.N (BUMP=patch|minor|major) and push
```

GHA then builds the wheel, gltanaka approves once on the `pypi-publish` environment, OIDC publishes to PyPI, and a GitHub Release with auto-generated notes is created. Only the gltanaka approval requires a human.

## Key changes

- **`pyproject.toml`**: `dynamic = ["version"]`, `[tool.setuptools_scm]` with `version_file`, strict `tag_regex`, `local_scheme = "no-local-version"`. **No `fallback_version`** — broken setups fail loud. `[tool.commitizen]` removed; `commitizen` dropped from dev deps.
- **`requirements.txt`**: drop `commitizen`.
- **`.github/workflows/release.yml`**: `fetch-depth: 0` on both jobs (setuptools-scm needs tags). `publish-pypi` gets `contents: write` for `gh release create`. Added idempotent "Create GitHub Release with auto-notes" step using `--verify-tag`.
- **`.github/workflows/unit-tests.yml`**: `fetch-depth: 0` on all three checkout steps.
- **`Makefile`**: replace `bump` + `release` split with a single idempotent `release` target. Handles fresh release, rerun-after-success (no-op), retry-after-push-failure (pushes existing local tag), drift detection (error). Validates `BUMP=patch|minor|major`. Uses strict semver tag discovery (`git tag --list --merged HEAD --sort=-v:refname` + grep regex). `make publish` now requires an exact semver tag at HEAD that exists on origin.
- **`tests/test_version.py`**: rewrite to validate PEP 440 form, reject setuptools-scm fallbacks (`0.0.0`, `0.0.0+unknown`), and assert exact-match-at-tag / `next-patch.devN` prefix-match-between-tags.
- **`.gitignore`**: `pdd/_version.py` (build-time artifact).
- **`README.md`, `pypi_description.rst`**: replace static badges with dynamic PyPI shields (`https://img.shields.io/pypi/v/pdd-cli.svg`). Remove hardcoded version strings.
- **`pdd/pdd_completion.sh`**: remove static `# Version:` comment.
- **`CHANGELOG.md`**: header note pointing to GitHub Releases; historical entries preserved.
- **`docs/ONBOARDING.md`**: replace commitizen workflow section with the new `make release` flow.

## Local validation

- `python -m build` produces `pdd_cli-0.0.242.dev1-py3-none-any.whl` with `Version: 0.0.242.dev1` (no `+gHASH`, confirms `local_scheme` applied).
- `twine check` PASSED.
- 3/3 tests in `tests/test_version.py` pass.
- `check-deps` reports deps in sync (commitizen removed from both pyproject.toml and requirements.txt; counts dropped 10→9 dev, 43→42 total).

## Pre-merge action required (one-time)

Before merging — or at the latest, before the first release on the new flow — back-fill `v0.0.241` as a GitHub Release so that `gh release create --generate-notes` for `v0.0.242` walks back to `v0.0.241` instead of the entire repo history:

```bash
gh release create v0.0.241 \
  --verify-tag \
  --title "v0.0.241" \
  --notes "Historical release; GitHub Releases adopted starting with v0.0.242."
```

## Test plan
- [x] `python -m build` succeeds and produces correctly-versioned wheel
- [x] `twine check dist/*.whl` passes
- [x] `pytest tests/test_version.py` all pass
- [x] Wheel installs cleanly in fresh venv; `pdd.__version__` matches METADATA
- [ ] CI: `unit-tests.yml` jobs pass with `fetch-depth: 0`
- [ ] CI: Cloud Build (`gcbrun`) passes
- [ ] First release on new flow (post-merge): `make release` cuts `v0.0.242`, GHA build+publish+release-create work end-to-end after gltanaka approval

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,6 +23,8 @@ jobs:
       id-token: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - uses: actions/setup-python@v5
         with:
@@ -48,10 +50,12 @@ jobs:
     runs-on: ubuntu-latest
     environment: pypi-publish
     permissions:
-      contents: read
+      contents: write
       id-token: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - uses: actions/setup-python@v5
         with:
@@ -70,3 +74,14 @@ jobs:
         with:
           skip-existing: true
           attestations: true
+
+      - name: Create GitHub Release with auto-notes (idempotent)
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          tag="${{ github.ref_name }}"
+          if gh release view "$tag" >/dev/null 2>&1; then
+            echo "GitHub Release $tag already exists; skipping."
+          else
+            gh release create "$tag" --generate-notes --verify-tag --target main
+          fi
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -37,6 +37,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Set up Python
         uses: actions/setup-python@v5
@@ -86,6 +88,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Set up Python
         uses: actions/setup-python@v5
@@ -117,6 +121,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Set up Python
         uses: actions/setup-python@v5
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 *.vsix
+
+# setuptools-scm generated version file
+pdd/_version.py
+
 .tmp/
 .agentic_prompt*
 .agentic_output_*.txt
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+> Historical record up to v0.0.241. See [GitHub Releases](https://github.com/promptdriven/pdd/releases) for current release notes.
+
 ## v0.0.241 (2026-05-17)
 
 ### Fix
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -148,23 +148,25 @@ Artifacts appear in `dist/`.
 - Participate in meetups and related community events when possible.
 
 ## Publishing a Release (Maintainers)
-1. Ensure build artifacts are up‑to‑date (see Building the Package).
-2. Install Twine if needed:
-   ```bash
-   pip install twine
-   ```
-3. Upload to PyPI with credentials:
-   ```bash
-   twine upload dist/*
-   ```
+
+The version is derived from `v*` git tags via setuptools-scm — no static version in `pyproject.toml`, no bump commit. To cut a release, on a clean `main`:
+
+```bash
+make release              # patch bump (default)
+make release BUMP=minor   # minor bump
+make release BUMP=major   # major bump
+```
+
+`make release` fetches tags, computes the next `vX.Y.Z`, tags HEAD, and pushes the tag. GitHub Actions then builds the wheel, waits for the `gltanaka` approval on the `pypi-publish` environment, publishes to PyPI via OIDC, and creates a GitHub Release with auto-generated notes.
+
+`make publish` is a local twine fallback that requires HEAD to be at a released tag that exists on origin.
 
 ## Pull Request Process
 1. Remove stray build/install artifacts before committing.
 2. Update `README.md` if you change the interface (env vars, commands, parameters, examples).
-3. Bump the version in `pyproject.toml` (SemVer) when appropriate; reflect changes in `README.md` as needed.
-4. Include tests in `tests/` that demonstrate the issue/feature and verify the behavior.
-5. Aim to include dev‑unit elements (prompt/code/example/tests) when applicable.
-6. **Make sure CI passes (unit tests).** If your change affects modules that have integration/LLM tests, run those locally for the impacted areas before marking your PR ready. You do not need to run the full integration suite -- maintainers will do that before merging.
-7. Seek two maintainer reviews prior to merge, or request a reviewer to merge if you lack permissions.
+3. Include tests in `tests/` that demonstrate the issue/feature and verify the behavior.
+4. Aim to include dev‑unit elements (prompt/code/example/tests) when applicable.
+5. **Make sure CI passes (unit tests).** If your change affects modules that have integration/LLM tests, run those locally for the impacted areas before marking your PR ready. You do not need to run the full integration suite -- maintainers will do that before merging.
+6. Seek two maintainer reviews prior to merge, or request a reviewer to merge if you lack permissions.
 
 Thank you for your contributions!
--- a/Makefile
+++ b/Makefile
@@ -58,8 +58,8 @@ help:
 	@echo "  make publish                 - Build & upload current version to PyPI"
 	@echo "  make publish-public          - Copy artifacts to public repo only"
 	@echo "  make check-deps              - Check pyproject.toml and requirements.txt are in sync"
-	@echo "  make bump                    - Bump version with commitizen on the current feature branch (commit only, no tag, no push)"
-	@echo "  make release                 - On main: tag HEAD with v\$$VERSION and push (Actions publishes to PyPI via OIDC after gltanaka approval)"
+	@echo "  make release                 - On main: tag HEAD with next vN.N.N and push (BUMP=patch|minor|major; default patch)"
+	@echo "                                  Actions publishes to PyPI via OIDC after gltanaka approval"
 	@echo "  make staging                 - Copy files to staging"
 	@echo "  make production              - Copy files from staging to pdd"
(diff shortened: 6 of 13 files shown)


gltanaka:
gcbrun

gltanaka:
## Review: DO NOT MERGE

Two blockers and one warning need to be resolved before this is safe to ship.

---

### BLOCKER 1 — `Makefile` (publish guard, ~line 619): dirty-worktree publish allowed

`make publish` only checks whether the HEAD commit is tagged with a matching remote release tag. It does not require a **clean worktree**. A developer who has uncommitted edits in `pdd/` or `pyproject.toml` can run `make publish` and `python -m build` will happily bake those dirty files into the uploaded wheel. This silently ships code that is not in git.

**Required fix:** Add `git diff --quiet HEAD || (echo 'ERROR: dirty worktree – commit or stash first' && exit 1)` before the build step inside `publish`.

---

### BLOCKER 2 — `Makefile` (release retry path, ~line 734): tag pushed but no GitHub Release → exits 0 with "nothing to do"

The retry-after-push-failure case (tag is on the remote, GitHub Release does not yet exist) should recreate the release. Instead the current logic falls through to an "already done" branch and exits 0 without creating the release or triggering PyPI publish. On the next `make release` the tag-exists check fires first and the whole run skips out again — leaving the release permanently stuck with no GitHub Release and no PyPI upload unless the developer manually intervenes.

**Required fix:** In the branch that detects `git ls-remote --tags origin vX.X.X` succeeds but `gh release view vX.X.X` fails, call `gh release create` (and subsequently trigger publish) rather than exiting 0.

---

### WARNING — `Makefile` (`publish-public` target, ~line 831): version derived from wrong git context

`publish-public` runs `git describe` inside the **public clone** after committing changes there. If the public clone's tags are stale (common after a fresh sync), `git describe` returns a `dev`-suffixed version string and `setuptools-scm` builds a `.devN` wheel instead of the release version. The public PyPI upload then has the wrong version.

**Required fix:** Pin `CURR_VER` from the **private** repo's tag (already known at call-time) and pass it explicitly into the public-update flow rather than re-deriving it from the public clone.

---

### MINOR — `CONTRIBUTING.md` (~lines 150, 164): stale release instructions

The contributing docs still describe the old manual Twine / static `pyproject.toml` version flow (`pip install twine`, edit `version = "..."`, etc.). These are now wrong and will mislead contributors.

**Required fix:** Update to document `make release` and the new setuptools-scm model.

---

Everything else checked out:
- `scripts/check_deps.py` passes (requirements.txt matches pyproject.toml).
- setuptools-scm config (`version_file`, `tag_regex`, `local_scheme`) looks correct.
- All workflows that install or build the package have `fetch-depth: 0`.
- `tests/test_version.py` tests are non-tautological and would catch real regressions.
- No remaining commitizen / `make bump` references in active code paths.
- `--verify-tag` behavior in `gh release create` is correct.

gltanaka:
## Second Review — Commit b68c0d4b5

### Blocker 1 (clean-tree check): RESOLVED

The clean-tree guard is the **first** thing that runs inside the `publish` target's `set -e` block (Makefile ~L620-626). It runs `git status --porcelain`, prints a diagnostic explaining *why* (build reads the working tree, not the commit), dumps `git status --short`, and calls `exit 1` before `HEAD_SHA` is even resolved. Every downstream step — `make build`, `make upload-pypi` — is behind this guard. ✓

**Bypass concern:** A developer could still call `make build` or `make upload-pypi` directly and skip the guard; those sub-targets have no dirty-check themselves. But `make publish` is the documented publish surface, and sub-target bypass is the user's own risk. Acceptable.

### Blocker 2 (stuck release recovery): PARTIALLY RESOLVED

The "tag already on origin at HEAD" path now does meaningful recovery instead of silently exiting 0. Walking through each sub-question:

**(a) Detects missing GitHub Release?** Yes. `gh release view "$$EXISTING_TAG" >/dev/null 2>&1` is used as a boolean probe; a non-zero exit means the release is absent. ✓

**(b) Creates it idempotently?** Mostly. `gh release create "$$EXISTING_TAG" --generate-notes --verify-tag --target main` is called when the release is missing. This call will error if the release *already exists* on a concurrent rerun; the `|| echo "(could not create ...)"` fallback swallows that error and continues — that's fine for the idempotency case. One caveat: if `gh` is not authenticated, the error is also swallowed and printed as a soft warning rather than hard failure. Users might not notice the release never got created. This is a **residual weakness** but not a full blocker given the diagnostic text is clear. ✓/~

**(c) Surfaces GHA workflow status?** Yes. `gh run list --workflow release.yml --event push --limit 1` is queried and the `status:conclusion` pair is printed with an actionable message for each case (success, failed, pending/waiting). ✓

**(d) Exits non-zero if recovery fails?** **No — this is the remaining concern.** After the recovery block (whether the release was created or not, whether the workflow failed or not), the target hits `exit 0` (explicit `exit 0` at the end of the `EXISTING_TAG` branch, Makefile ~L771). If `gh release create` failed silently (auth issue, permissions), or if the last workflow run `completed:failure`, `make release` still exits 0. A CI wrapper or a human re-running would get a green exit code even when the release is in a bad state.

The three actionable workflow-failure cases:
- `completed:failure` → prints a message to re-trigger, **but exits 0**
- gh not installed → prints a suggestion, exits 0 (acceptable — gh is optional)
- gh not authenticated → `gh release create` fails, error is swallowed, exits 0 (less acceptable)

The original BLOCKER 2 was about exiting 0 when the release is stuck. For the "workflow failed" sub-case that is still true. This is a **residual weakness** but substantially mitigated: a user will at minimum see the diagnostic text.

### Warning (publish-public version drift): ACKNOWLEDGED / ADDRESSED

The previous review partially mis-framed this: `publish-public` syncs files only and does not publish to PyPI. The fix commit improves the tagging logic in `publish-public` anyway (Makefile ~L860-866): instead of reading a static `version = "..."` field from `pyproject.toml`, it now uses `git describe --tags --abbrev=0 --match='v*'` to find the version, and adds a guard that skips tagging entirely if no valid `vN.N.N` tag is found. This prevents a `.devN` string from being used as a tag and silently corrupting the public repo's tag history. ✓

### Minor (CONTRIBUTING.md): RESOLVED

The Twine-based "Publishing a Release" section is fully replaced with `make release` instructions (CONTRIBUTING.md ~L151-163). The `BUMP=minor/major` variants are documented. The "bump the version in pyproject.toml" step is removed from the PR process. The description accurately matches the Makefile behavior. ✓

### New Issues Found

**1. `make release` exits 0 on workflow failure (residual from Blocker 2)**
Makefile, the `completed:*` case in the recovery block. When `LATEST_RUN_STATUS` is `completed:failure`, the target prints a message but falls through to `exit 0`. If this target is used in a CI/CD script or by an operator expecting non-zero on failure, the stuck state is invisible. Suggested fix: `exit 1` in the `completed:*` case (not `completed:success`).

**2. `gh run list --event push` may report the wrong run**
The workflow is triggered by a tag push (event `push` with a `v*` ref filter), not a branch push. `gh run list --event push --limit 1` returns the *most recent* push-triggered run of `release.yml`, which could be for a *different* tag if another release was attempted recently. There is no filter on the specific tag/ref. This is a diagnostic-accuracy issue, not a correctness issue (the release flow itself is unaffected), but the status message could mislead.

**3. `gh release create --target main` hardcodes branch name**
If the repo's default branch is ever renamed (unlikely but possible), `--target main` would create the release pointing at a non-existent branch. Minor.

**4. No new tests**
The fix does not add any test coverage for the recovery path. This is consistent with the rest of the Makefile test surface (tested manually), so not a blocker, but worth noting.

### Verdict: MERGE (with awareness of residuals)

All original blockers are substantively resolved. The remaining issues are (1) a diagnostic-exit-code weakness in the recovery path that does not affect correctness of the actual release, and (2) a minor diagnostic-accuracy issue with the workflow run lookup. Neither causes incorrect behavior or data loss. The CONTRIBUTING.md is accurate. The clean-tree guard fully closes Blocker 1. This is mergeable.

gltanaka:
**Round 3 / Final — MERGE**

All three flagged issues from round 2 are resolved. Findings per the new commit `5015de0c9`:

**Q1 — RECOVERY_FAILED tracker (exit codes)**

| Scenario | Expected | Result |
|---|---|---|
| Release missing, `gh release create` succeeds | exit 0 | PASS |
| Release missing, `gh release create` fails | exit 1 | PASS |
| Release exists, workflow `completed:success` | exit 0 | PASS |
| Release exists, workflow `completed:failure` | exit 1 | PASS |
| Release exists, workflow `queued`/`in_progress` | exit 0 | PASS |
| Release exists, no run found for this SHA | exit 0 | PASS |

The `RECOVERY_FAILED=0` tracker is initialized before the branch, set to 1 on `gh release create` failure or on `completed:*` (non-success) conclusion, and consumed at the final `exit` — all paths are correct.

**Q2 — `headSha` scoping**

- jq syntax is valid. ✓
- Filter is `.[] | select(.headSha=="$HEAD_SHA")` before `head -1`, so stale runs from prior release attempts on different SHAs are excluded. ✓
- `$HEAD_SHA` is inside a double-quoted shell argument with properly escaped inner quotes, so the variable expands correctly. ✓

**Q3 — `--target main` hardcoded**

PASS. The `release` target depends on `check-release-branch`, which verifies HEAD is on `main` and origin is in sync before this code path is ever reached. The hardcoding is safe and accurate.

**Q4 — New bugs**

None found. No logic errors, off-by-one conditions, variable scoping issues, or unintended side effects introduced by this commit.

---

Both substantive issues from round 2 are correctly addressed. Ready to merge.

PR #1083, opened 2026-05-19
fix(tests): pin HOME/CODEX_HOME so tests can't pollute developer's real home

gltanaka:
## Summary

Stops the long-standing developer-machine bug where running pdd's pytest suite could silently overwrite `~/.bashrc`, `~/.pdd/jwt_cache`, or `~/.codex/auth.json` — breaking the local shell, JWT cache, or Codex CLI. Mirrors the parallel fix [promptdriven/pdd_cloud#1486](https://github.com/promptdriven/pdd_cloud/pull/1486) which closed the same class of bug on the pdd_cloud side.

## Risks in this repo

Two distinct write paths:

1. **Module-level `Path.home()` constants** — `pdd/auth_service.py:20` and `pdd/get_jwt_token.py:74` both define `JWT_CACHE_FILE = Path.home() / ".pdd" / "jwt_cache"` at **import time**. Without HOME pinning *before* `from pdd.llm_invoke import …` runs in the conftest, those constants capture the developer's real `~/.pdd/jwt_cache`. (Verified empirically: `pdd.auth_service.JWT_CACHE_FILE` was resolving to `/Users/gregtanaka/.pdd/jwt_cache` in the test process on pre-fix code.)

2. **Function-level writes in `pdd/cli_detector.py:_save_api_key`** (lines 463–491):
   - `Path.home() / ".pdd"` → `mkdir`
   - `~/.pdd/api-env.{shell}` → `write_text`
   - `~/.bashrc` / `~/.zshrc` / `~/.config/fish/config.fish` → `open(..., "a")`

   Today all callers that hit this in tests pre-patch `Path.home()` to `tmp_path` — but any future test (or a `uses_real_cli_detector`-marked test that exercises real CLI bootstrap) would pollute the developer's real shell rc.

## Fix — two layers in `tests/conftest.py`

1. **Module-level** (runs before any `pdd.*` import): pins `HOME` to a `tempfile.mkdtemp` sandbox, with `atexit` cleanup. Also pins `CODEX_HOME` under that sandbox HOME — neutralizes any inherited direnv-set `CODEX_HOME="\$PWD/.codex"` from the developer's parent shell (a real concern in pdd_cloud-adjacent workflows). Both must happen **before** `from pdd.llm_invoke import InsufficientCreditsError` so the module-level `JWT_CACHE_FILE` constants resolve inside the sandbox.

2. **Autouse per-test fixture `_enforce_isolated_home`**: re-asserts both env vars via `monkeypatch.setenv` so a test that mutates `HOME` mid-run doesn't leak the change into the next test.

## New regression tests

`tests/test_home_isolation.py` (5 tests). On pre-fix code:
- `test_pytest_home_is_isolated_sandbox` — **FAIL** (HOME is `/Users/gregtanaka`)
- `test_codex_home_env_is_pinned_under_sandbox_home` — **FAIL** (CODEX_HOME is project-local)
- `test_path_home_resolves_to_sandbox` — **FAIL** (Path.home is real)
- `test_jwt_cache_file_module_constants_resolve_under_sandbox` — **FAIL** (constants captured `/Users/gregtanaka/.pdd/jwt_cache`)
- `test_save_api_key_writes_under_sandbox_only` — **PASS** (uses its own monkeypatch)

After the fix: **5/5 pass**.

## Test plan

- [x] `tests/test_home_isolation.py` — 5/5 pass
- [x] `tests/test_cli_detector.py + tests/test_agentic_common.py + tests/test_setup_tool.py` — **369 passed, 0 failures**
- [x] End-to-end: real `~/.codex/auth.json` md5 unchanged through the entire change-and-test cycle; `~/.bashrc` never created in `$HOME`; `~/.pdd/jwt_cache` unchanged

## Risks / out-of-scope

- No production source changes; only `tests/conftest.py` + new test file.
- The `dotenv.load_dotenv()` call uses default `override=False`, so a `.env` file won't clobber the pinned HOME/CODEX_HOME values.
- pdd doesn't currently read `CODEX_HOME` itself; pinning it is cheap defense in depth for hypothetical future codex-integration code.


gltanaka:
Do not merge as-is. The implementation appears to isolate HOME correctly, but the new regression tests add an environment-dependent failure.

`tests/test_home_isolation.py` hard-codes `SANDBOX_ROOTS = ("/tmp/", "/var/folders/", "/private/var/folders/")`. `tests/conftest.py` creates the fake HOME with `tempfile.mkdtemp()` and no explicit `dir`, so it correctly honors `TMPDIR` / `TEMP` / `TMP`. In any developer or CI environment with a custom temp directory outside those roots, the tests fail even though HOME is sandboxed.

Reproducer from this branch:

```bash
mkdir -p .tmp/custom-tmp
TMPDIR=/Users/gregtanaka/Documents/pdd_cloud/pdd/staging/public/pdd/.tmp/custom-tmp conda run -n pdd pytest tests/test_home_isolation.py -q
```

Result: 4 failures in `test_pytest_home_is_isolated_sandbox`, `test_codex_home_env_is_pinned_under_sandbox_home`, `test_path_home_resolves_to_sandbox`, and `test_jwt_cache_file_module_constants_resolve_under_sandbox` because the sandbox is under the custom TMPDIR rather than `/tmp` or `/var/folders`.

Required change: remove the hard-coded temp-root whitelist. Assert against the actual sandbox invariant instead, e.g. expose the conftest fake-home path via a fixture and assert:

- `Path(os.environ["HOME"]) == pytest_fake_home`
- `Path.home() == pytest_fake_home`
- `Path(os.environ["CODEX_HOME"]) == pytest_fake_home / ".codex"`
- `auth_service.JWT_CACHE_FILE == pytest_fake_home / ".pdd" / "jwt_cache"` and same for `get_jwt_token`

That keeps the important safety checks while allowing valid temp directories selected by Python / the OS. After that, rerun the new test plus the credential/HOME-heavy subset.

gltanaka:
Review applied in 51a618da7. One legit blocker, fully addressed:

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | blocker | Hard-coded `SANDBOX_ROOTS = ("/tmp/", "/var/folders/", "/private/var/folders/")` whitelist false-fails when `TMPDIR` points outside those roots — `tempfile.mkdtemp()` correctly honors `TMPDIR` but the assertions rejected the legitimate sandbox path. | Dropped the whitelist entirely. Added a session-scoped `sandbox_home` fixture in `tests/conftest.py` that returns the actual `tempfile.mkdtemp()` path. Tests now assert exact equality against that fixture (and `<fixture>/.codex`, `<fixture>/.pdd/jwt_cache`). Strictly stronger invariant — catches misconfiguration where HOME is pinned to *a different* sandbox, not just "any non-/tmp path". |

(Note: the fixture is named `sandbox_home` rather than the reviewer's suggested `pytest_fake_home` — pytest rejects `pytest_*` as a reserved hook-namespace prefix and errors out with `unknown hook 'pytest_fake_home'`.)

Verified both default and the reviewer's exact custom-TMPDIR repro:

```
$ pytest tests/test_home_isolation.py -q                                                     # default
5 passed in 0.13s
$ TMPDIR=.../.tmp/custom-tmp pytest tests/test_home_isolation.py -q                          # repro
5 passed in 0.06s
```

/gcbrun

PR #1085, opened 2026-05-19
Fix: preserve --protect-tests during agentic fallback

### LINKED ISSUE #771: pdd fix: --protect-tests flag is dropped when falling back to agentic mode
The `--protect-tests` flag is silently dropped when `pdd fix` falls back to agentic mode.

Users pass `--protect-tests` to tell PDD "my tests are correct; don't let the LLM weaken them to make the code pass." This works correctly in the standard `pdd fix --loop` retry phase. If that phase fails and PDD hands the task to the agentic fallback, the flag is silently set to `false` in the instructions the fallback receives.

The agent is then told it may edit test files, which the user does not want.

### How to recreate

I ran `pdd fix` with the flag on a module with failing tests:

```bash
pdd fix --loop --protect-tests \
  --verification-program tests/test_auto_include.py \
  prompts/auto_include_python.prompt \
  pdd/auto_include.py \
  tests/test_auto_include.py \
  test_errors.log
```

PDD's standard loop honored `--protect-tests`. After 5 failed iterations, it fell back to the agentic mode. The fallback read an instruction file (`agentic_fix_instructions.txt`) that PDD generated, and handed it to Gemini.

That instruction file contains this line:

```
## Test Protection Mode: false
```

Even though I had passed `--protect-tests`. The flag did not propagate from the CLI into the fallback's instruction template.

### Proposed fix

When PDD constructs the agentic fallback instructions, use the `protect_tests` boolean. The template should read:

```
## Test Protection Mode: {protect_tests}
```

...populated from the CLI flag. And the constraints section could add:

```
If Test Protection Mode is true, you MUST NOT modify any file under `tests/`.
If your analysis suggests the test file itself is wrong, stop and report this
in your final message — do not edit the test.
```

imradawoodani:
Fixes #771

### What changed

- Added a regression test in `tests/test_fix_error_loop.py` to ensure
  `protect_tests=True` is forwarded into the agentic fallback helper.
- Updated `pdd/pdd/fix_error_loop.py` so `_safe_run_agentic_fix()` accepts
  `protect_tests` and every fallback path passes it through to `run_agentic_fix()`.

### Why

When `pdd fix --loop --protect-tests` falls back to agentic mode, the
flag was previously dropped and the fallback agent could modify tests.
This fix preserves the user’s intent and prevents test files from being
edited during agentic fallback.

### Testing

- `export PDD_PATH=$(pwd)`
- `pytest -q tests/test_fix_error_loop.py -k protect_tests`
- `pytest -q tests/test_fix_error_loop.py`


gltanaka:
target 5/21

PR #1078, opened 2026-05-18
fix(checkup): SHA-backed trust boundary for review-loop fixer claims (#1062)

Serhan-Asad:
## Summary

Resolves #1062: SHA-backed trust boundary for review-loop fixer claims.

The review loop's fixer could report `success` and the final report would render `verification=verified` even when the push produced no diff or the remote PR head advanced past the verifier-reviewed SHA. This PR introduces three gates so the loop never claims a finding is fixed without a real, verified, current-head push.

1. **Pushed SHA as the trust token.** `_commit_and_push_if_changed` returns a 3-tuple `(pushed, message, pushed_sha)`. `None` means no commit landed on the PR head — the verifier is not invoked and the finding stays open.
2. **Per-fix verification state.** `FixResult` now carries `push_status`, `pushed_head_sha`, `verified_head_sha`, `verification_status`. The rendered report shows the actual SHA chain and labels rows `attempted` until verification truly succeeded.
3. **Stale-SHA gate before final report.** A new `_fetch_remote_pr_head_sha` helper reads the live remote PR head SHA via `git ls-remote` (with the same token-fallback as `_fetch_pr_head_for_rebase`). If the live SHA does not match the verifier-reviewed SHA, `fresh_final_status` is downgraded to `"stale"` (now a `HARD_NOT_CLEAN_STATES` member), verified fixes flip to `verification_status="stale"`, and any "fixed" findings are reopened.

## Tests

- 3 new regression tests in `tests/test_checkup_review_loop.py::TestReviewLoopShaVerification`
  - `test_fixer_success_without_diff_keeps_finding_open` (no-diff push)
  - `test_remote_head_advanced_after_verify_marks_stale` (stale SHA after verify)
  - `test_happy_path_pushes_verifies_same_sha_marks_fixed` (happy path)
- 2 new direct-call tests in `TestCommitAndPushIfChanged` (no-diff and no-eligible-changes paths return `None` for the SHA)
- All existing review-loop tests updated for the new 3-tuple contract and still pass (101 tests in `test_checkup_review_loop.py`, 8 in `test_issue_1021_reproduction.py`)
- Prompt updated to match the new contract so the next `pdd sync` does not revert the fix

## Test plan

- [x] All 3 new regression tests fail on `main` and pass on this branch
- [x] All existing tests in `tests/test_checkup_review_loop.py` still pass (101 passed, 1 skipped)
- [x] All tests in `tests/test_checkup_pr_mode.py` still pass (33 passed)
[...]

Issue #1063, opened 2026-05-18
bug: checkup review loop can commit generated-code fixes without prompt-source guard


Serhan-Asad:
## Problem

`pdd checkup --review-loop` can let a fixer patch a PDD-generated code file without also updating the prompt source that owns that code.

That is unsafe for this repo because generated modules are not source-of-truth files. The prompt is the source contract; the code is regenerated from it. If the bot commits a code-only fix, a later `pdd sync <module>` can overwrite the fix or reintroduce the old behavior.

## Historical Repro Evidence

PR #1055 exposed this failure mode.

Commit `bf57242d9a18072f3d444f91d9a834ba067f15ca` changed only:

```text
M pdd/agentic_update.py
```

It changed `_compute_include_allowlist` from the prompt-specified contract:

```python
_compute_include_allowlist(prompt_path: Path, *, max_depth: int = 3) -> Set[Path]
```

to generated code without the `max_depth` parameter and with unbounded traversal.

At that commit, `pdd/prompts/agentic_update_python.prompt` still required:

```text
_compute_include_allowlist(prompt_path: Path, *, max_depth: int = 3) -> set[Path]
...
cap recursion at max_depth levels
```

So the review-loop fixer produced prompt/code drift.

## Current Status / Important Correction

The latest PR #1055 head has since updated `pdd/prompts/agentic_update_python.prompt` to match the no-depth-cap direction, so the specific `agentic_update` drift is no longer present at the current PR head.

This issue is still valid because the workflow allowed the bad intermediate bot commit and there is no deterministic guard preventing the same class of generated-code-only fix from happening again.

## Why This Is A Real Workflow Bug

The current review-loop flow relies mainly on LLM review instructions, not an invariant:

- `_run_fix()` lets the fixer edit files and then reports `changed_files` from git status.
- The loop then calls `_commit_and_push_if_changed(...)` and pushes whatever eligible tracked changes exist.
- The reviewer prompt asks reviewers to check prompt/source-of-truth drift, but that is advisory and can be missed.
- The fixer prompt does not require: "if you change a generated module registered in `architecture.json`, update the owning prompt or refuse."
- There is no pre-commit check that maps changed generated code paths back to their prompt entries in `architecture.json`.

For PDD itself, that missing guard is the bug.

## Expected Behavior

Before committing/pushing fixer changes, `checkup_review_loop` should detect generated-code-only changes for prompt-owned modules.

[...]

PR #1095, opened 2026-05-19
fix(checkup): enforce deterministic repo gates before clean verdict (#1092)

### LINKED ISSUE #1092: pdd checkup --pr should enforce repo-agnostic deterministic gates before clean verdict
## Problem

`pdd checkup --pr --review-loop` can still accept an LLM reviewer/verifier `clean` result as the final readiness signal even when a deterministic local check would fail in the PR worktree.

That is a product bug. PDD is meant to review arbitrary user repositories, not only this repo. If a repo has a reliable local format, whitespace, syntax, compile, lint, type, or targeted test gate, a clean LLM verdict must not override that gate.

Concrete failure class: a PR is reported clean while a configured format check such as `prettier --check`, `git diff --check`, `python -m py_compile`, or an equivalent repo-local deterministic check fails. The loop should surface the failed gate as blocking evidence, let the fixer address it when the failure is code-owned, then rerun the gate before accepting the PR.

## Product requirement

The gate layer must be repository-agnostic and safe for untrusted PRs.

Do not design this around PDD-specific files, PDD's Makefile, Conda, `architecture.json`, or our own test suite. The implementation should work for a small JavaScript app, a Python package, a mixed repo, or a repo with only generic Git checks.

PR contents are untrusted. Do not blindly execute scripts, Make targets, CI jobs, package-manager lifecycle hooks, deploy commands, install commands, or arbitrary commands found in PR-modified files. A malicious fork PR can change `package.json`, `Makefile`, or config files. Default gates must be built from fixed, conservative commands and narrowly parsed trusted config, not from unchecked PR-authored command strings.

## Proposed direction

Implement this in upstream `pdd checkup --pr --review-loop` as a generic deterministic-gate layer:

1. Discover a conservative gate set from the PR worktree and changed-file scope, with no PDD-specific assumptions.
2. Always include safe universal checks when applicable, especially PR-range whitespace/conflict-marker checks such as `git diff --check <merge-base>...HEAD` rather than only checking the clean working tree.
3. Add ecosystem-specific gates only when they are local, deterministic, bounded, and non-mutating. Examples: changed-file Python syntax compile, direct local Prettier check, direct local TypeScript typecheck, direct local lint/type tools when clearly configured.
4. Prefer direct tool invocation over package-manager scripts. If package scripts are supported at all, they must not execute lifecycle hooks or shell chains, and they must be parsed/validated so a PR cannot smuggle `curl`, `rm`, deploys, installs, background jobs, or command substitution into the gate.
5. Do not run generic Makefile targets, arbitrary CI workflows, pre-commit hooks, package install commands, cloud jobs, e2e suites, or deploy-like commands by default. Those require an explicit trusted allowlist/config design, preferably read from base-trusted configuration or operator input.
6. Run gates inside the loop-owned PR worktree before every path that would mark the PR clean: initial reviewer clean, review-only clean, fallback-reviewer clean, post-fix verifier clean, and any pending-findings-empty shortcut.
7. Convert repo-owned gate failures into synthetic blocking findings with command, exit code, source, and bounded output excerpt. Feed those findings to the normal fixer path and rerun gates after fixes.
8. Distinguish repo failures from gate-runner infrastructure failures. Both should block a clean verdict, but missing tools, timeouts, permission errors, and runner crashes should be reported as unverified/diagnostic gate failures, not blindly assigned to the fixer as if application code can always fix them.
9. Persist gate artifacts under the existing checkup artifact directory and include a concise final-report section plus machine-readable state for downstream verdict adapters.
10. Scrub secrets before truncating, persisting, logging, or rendering any subprocess output or exception text. Gate output can contain env vars, tokens, URLs, or auth headers.

## Non-goals / caution

This should not try to reproduce full CI in v1. Full CI can be slow, flaky, credentialed, external-service-dependent, or side-effectful. The goal is a small, trustworthy local hard stop that catches deterministic failures the LLM can miss.

This should not become a PDD-only quality gate. PDD-specific sync/prompt/architecture checks can remain separate or be added as optional repo-specific gates, but the default product behavior must work on arbitrary repositories.

## Acceptance criteria

- A failing discovered deterministic gate prevents a clean checkup verdict even if the LLM reviewer/verifier returns `clean`.
- Gates run before every clean-exit path in the review loop.
- Gate discovery and execution are generic: tests use synthetic non-PDD repos, at minimum one Python repo and one JavaScript/TypeScript-style repo.
- A committed whitespace/conflict-marker failure in the PR diff is caught via a PR-range `git diff --check`, not missed because the worktree is clean.
- A malicious or PR-modified `package.json` script with shell chaining, lifecycle hooks, network calls, install/deploy behavior, or command substitution is not executed by default.
- Repo-owned gate failures become actionable blocking findings and are rerun after the fixer changes the worktree.
- Runner/infrastructure failures block clean but are clearly labeled as gate-runner diagnostics/unverified status.
- Gate command output and exception text are bounded, persisted for audit, rendered in the final report, and secret-scrubbed before any public/logging surface.
- The feature has an opt-out only for diagnostics/emergency use; the default path must be fail-closed for discovered gate failures.

Serhan-Asad:
## Summary

- Adds `pdd/checkup_gates.py` with conservative allowlisted gate discovery (`git diff --check`, `prettier --check` from `package.json`, `python -m py_compile` for changed `.py` files, optional `ruff`/`black`/`mypy` when configured in `pyproject.toml`, optional `npx tsc --noEmit` when `tsconfig.json` + `typescript` dep are present).
- Wires `_enforce_gates_before_clean` into **every** clean-exit site in `pdd/checkup_review_loop.py` (round-start "no actionable findings", "pending findings filtered to empty", post-verify clean, review-only clean, fallback-reviewer clean) so a failing local gate cannot be overridden by an LLM "clean" verdict.
- Failed gates convert into synthetic blocker `ReviewFinding` rows tagged `reviewer="gate:<name>"` and `area="deterministic-gate"`; they flow through the normal `_fix_prompt` path so the fixer addresses the local failure and gates re-run on the next round until they pass.
- Subprocess hygiene: list-form argv (no `shell=True`), bounded per-gate wall-clock timeout, ~10KB-per-stream output truncation, `_scrub_secrets` applied before persistence, env scrubbed of `PDD_*` with `CI=1`/`NO_COLOR=1`/`FORCE_COLOR=0`. The runner never raises — timeouts, missing binaries, and OS errors are captured as `GateResult(exit_code=None, error=...)` rows.
- Artifacts: `round-{R}-{mode}-gates.json` + per-gate text under `.pdd/checkup-review-loop/issue-{N}-pr-{M}/` (already gitignored via `.pdd/*` so `_commit_and_push_if_changed` does not stage them; PR #1078 owns that helper and stays untouched).
- Final report gains a `### Deterministic Gates` section grouped by round (per-gate name, source, exit code, duration, failure tail). `final-state.json` carries a top-level `gates` field with the full audit trail (passing + failing gates).
- New CLI flags: `--no-gates`, `--gate-timeout SECONDS`, repeatable `--gate-allow CMD`. Defaults preserve legacy behaviour on repos without discoverable gates.
- Source-of-truth contract: `pdd/prompts/checkup_gates_python.prompt`, `pdd/prompts/checkup_review_loop_python.prompt`, `pdd/prompts/commands/checkup_python.prompt`, `pdd/prompts/agentic_checkup_python.prompt`, completion specs (bash/zsh/fish), `context/checkup_review_loop_example.py`, and `architecture.json` are all updated so the next `pdd sync` does not revert the code (#1063 trap).

## Design-doc deviations

- Design called for the helper to guard "site 6 — top-of-loop early exit before line ~894". That line is actually the not-clean failure path; there is no sixth site. Final count of guarded clean-exit sites is **five** (matching codex's pre-review note).
- Design said "verify that `_commit_and_push_if_changed` filters `.pdd/checkup-review-loop/**`". The repo's existing `.gitignore` covers `.pdd/*` so artifacts are never seen as untracked by `_git_untracked_files`. No change to `_commit_and_push_if_changed` (PR #1078's surface) was required.
- Targeted pytest gate, `.pre-commit-config.yaml` hooks, and arbitrary `Makefile` targets are explicitly out of scope (codex flagged them as too risky); v2 follow-up.

## Test plan

- [x] Run `pytest -vv tests/test_checkup_gates.py` — 25 new tests cover discovery (allowlist, package.json scripts, ruff/black/mypy, eslint guard, list-form), the runner (exit-code, truncation, timeout, missing binary, secret scrub, env hygiene, cwd), and the findings adapter (failed gate → blocker, passed gates skipped, runner-side error → finding, stable dedup key across rounds).
- [x] Run `pytest -vv tests/test_checkup_review_loop.py::TestReviewLoopDeterministicGates` — 10 new integration tests pin the per-site behaviour: failing prettier gate blocks the clean verdict, fixer receives the gate finding, gates rerun and the loop reaches clean once they pass, verify-clean+failing-gate stays open, review-only reports findings, fallback-reviewer-clean does not ship, `--no-gates` preserves legacy behaviour, gate artifacts are persisted, dedup key is stable, the `### Deterministic Gates` section + `final-state.json` gates field render.
- [x] Run `pytest -vv tests/test_checkup_gates.py tests/test_checkup_review_loop.py tests/test_checkup_pr_mode.py tests/test_issue_1021_reproduction.py` — 194 passed, 1 skipped (no regressions).
- [ ] Live PR review against a repo with a `prettier --check` failure to confirm the end-to-end veto in production.

Closes #1092


Serhan-Asad:
## Independent verification

Tested in an isolated worktree off PR head (`011f7fd24`) against a baseline worktree at the merge-base on `main` to attribute any failures. Full venv install, no shared state.

### Phase summary

| Phase | Result |
|---|---|
| Static (`py_compile` + `pylint -E`) on new modules | OK · 10.00/10 |
| Focused unit (`test_checkup_gates` + `TestReviewLoopDeterministicGates` + `test_checkup_pr_mode` + `test_issue_1021_reproduction`) | **244/244 pass** in ~30s |
| Full `pytest tests/` (9,225 tests) | **9092 pass / 62 fail / 71 skip** — every sampled failure pre-exists on the baseline; zero failures touch checkup_gates / checkup_review_loop / checkup_pr_mode / issue_1021 |
| Adversarial fixtures (6 hostile-repo scenarios) | **6/6 pass** |
| Live E2E (full review loop + mocked LLM 'clean' + failing gate, report dumped) | **4/4 pass** — report renders the `### Deterministic Gates` section, the `gate:prettier-check` blocker, and `reviewer-status: codex=findings` (not clean) |
| Fail-closed paths (timeout · missing binary · corrupt `package.json` · unreachable base ref · read-only artifacts dir) | **5/5 pass** — every error emits a synthetic `gate:*` blocker, no silent fall-through |
| Regression (`--no-gates` legacy parity · gates-on with no discoverable gates · review-only + failing gate) | **4/4 pass** — `--no-gates` is bit-for-bit legacy |
| Perf (1 / 10 / 50-file diff) | 0.555s / 0.960s / 2.172s — ~1464× faster than the worst-case `60s × n_gates` bound, negligible vs MODULE_TIMEOUT |

### Adversarial scenarios specifically exercised

1. Python syntax-broken file in PR → py-compile blocker
2. JS unformatted file (verifies prettier script-gate is intentionally skipped on JS PRs since configs can `require()` arbitrary local JS)
3. **`.npmrc script-shell=./evil-sh` RCE primitive** → entire npm family correctly skipped; the planted `evil-sh` was never spawned
4. **`PATH=.:$PATH` + worktree-shipped `./mypy` shim** → PATH sanitization stripped worktree-resolving entries; resolved to the system tool, not the shim
5. **pyproject mypy plugin matches worktree package name** → mypy gate skipped (would have loaded PR-controlled plugin code under editable install)
6. **Whitespace error in PR-range diff (worktree currently clean)** → `git diff --check <base>...<head>` caught it (exit 2) — the literal acceptance case

### Notes

- The unchecked test-plan box (live PR with seeded prettier failure) is covered by the Phase 4 driver above — same code path, real subprocess execution for gates, mocked LLM standing in for the 'clean' verdict.
- Default flips to gates-on; existing runs that were quietly clean may newly block. This is the intended behavior.
- Stage-2 cloud rollout needs a separate smoke once `pdd-cli` publishes to PyPI (`backend/functions/requirements.txt` pins `pdd-cli>=0.0.242`).

LGTM to merge.

Serhan-Asad:
## Cloud Batch — 100% pass

Ran the full 77-task Cloud Batch suite against PR head (`011f7fd24`) from an isolated worktree:

**Run 1** (`pdd-test-run-20260525-122644-011f7fd24`): 73 passed, 2 failed, 2 errors. All 4 failures shared one root cause — JWT-exchange auth quota on the staging cloud:
- task 9 (`pytest chunk_9`): `test_context_generator.py` × 3 tests, 300s timeout each — cloud LLM auth fell back to device-code which times out
- task 13 (`pytest chunk_13`): `test_real_fix_command`, same 300s device-code timeout
- task 68/69 (`cloud_regression` cases 1/2): direct `QUOTA_EXCEEDED` on JWT exchange pre-flight

None of these tasks touch the modules in this PR (`pdd/checkup_gates.py`, `pdd/checkup_review_loop.py`, `pdd/commands/checkup.py`, related prompts/architecture).

**Run 2** (`pdd-test-run-20260525-125121-011f7fd24`, 14:20 wallclock): **77 passed, 0 failed, 0 errors.** Transient quota cleared, confirming the run-1 failures were infrastructure flakes, not PR-introduced regressions.

Combined with the earlier local verification (244 focused unit tests + 6 adversarial fixtures + live review-loop E2E + 5 fail-closed scenarios + 4 regression scenarios), Cloud Batch is the missing piece. Done.

gltanaka (changes requested):
This feature is needed: issue #1092 is a real product gap, and `main` does not currently have a deterministic gate layer. But I would not merge this PR as-is.

Required change: the gate layer must not run bare `git` from inside the untrusted PR worktree with the inherited `PATH`.

The new changed-file discovery added for gates calls `subprocess.run(["git", ...], cwd=worktree, ...)` in `_pr_changed_files_all` (`pdd/checkup_review_loop.py`, around the new `git rev-parse` / `git diff --name-only` calls). That runs before `run_gates()` builds the sanitized gate subprocess env. If an operator/CI environment has `PATH=.:$PATH` or another relative/worktree-resolving PATH entry, a PR can add an executable `./git` and it will be executed during gate enforcement, before the sanitized runner ever starts.

Minimal repro against this PR head:

```bash
tmp=$(mktemp -d)
printf '#!/bin/sh\necho shim >> "$PWD/marker"\nexit 1\n' > "$tmp/git"
chmod +x "$tmp/git"
PATH=.:$PATH PYTHONPATH=/tmp/pdd-pr1095 python - <<'PY' "$tmp"
import sys
from pathlib import Path
from pdd.checkup_review_loop import _pr_changed_files_all
root = Path(sys.argv[1])
_pr_changed_files_all(root, {"base_ref": "main"})
print((root / "marker").read_text() if (root / "marker").exists() else "NO_MARKER")
PY
```

This prints multiple `shim` lines, proving the PR-controlled `./git` ran. That violates the same trust-boundary invariant the PR already added for npm/mypy/ruff/black path resolution.

Please fix this by resolving a trusted `git` binary outside the PR worktree, and use that absolute path plus a sanitized `PATH` for every gate-layer git subprocess, including `_pr_changed_files_all`, `_resolve_pr_base_spec`, `_is_git_worktree`, `_git_diff_check_gate`, and the PR-base refresh path if it is part of the gates-on flow. Add a regression test with `PATH=.:$PATH` and a worktree-local `git` shim that fails if the marker file is created.

gltanaka (approved):
Re-reviewed PR head `8be4662e27ee0a8e3bc732e020461b295f0ed600`. The requested trusted-git fix is in place: the prior `PATH=.:$PATH` + worktree-local `./git` shim repro now prints `NO_MARKER`, and the new focused trusted-git regressions pass locally. Also ran `pytest -q tests/test_checkup_gates.py`, `pytest -q tests/test_checkup_review_loop.py::TestReviewLoopDeterministicGates tests/test_checkup_review_loop.py::TestPrChangedFilesScannerBaseLocalRef`, and `python -m pdd checkup --validate-arch-includes`; all passed.

LGTM to merge.

PR #1098, opened 2026-05-19
Revert PR #1091

Serhan-Asad:
Reverts merge commit 2fbe2bb6f for PR #1091.\n\nThis restores the tree to the pre-merge parent f449cc0d0 while keeping history linear and reviewable. Verified with:\n\n```\ngit diff --exit-code f449cc0d0 HEAD\n```


Serhan-Asad (commented):
Reviewed the revert. The diff cleanly backs out merge commit 2fbe2bb6f from PR #1091, and the resulting tree matches pre-merge parent f449cc0d0 ().

PR #1156, opened 2026-05-24
fix(opus-4.7): flip CSV to adaptive + remove PDD_FORCE gate on github_copilot

gltanaka:
## Summary

- Flip the Anthropic claude-opus-4-7 CSV row to reasoning_type=adaptive (Anthropic enforced the new shape ~2026-05-23 17:25 UTC; PR #1047 added the code path but deferred the CSV flip "out of scope" — this closes that).
- Teach the catalog generator about adaptive so re-generation doesn't revert the flip.
- Drop the PDD_FORCE gate on the github_copilot token-file check so non-interactive contexts (Cloud Run, library use) don't hang on litellm device-flow OAuth when no Copilot token is present.

See commit body for the full incident timeline and rationale.

## Test plan

- [ ] `pytest tests/test_llm_invoke.py -k github_copilot -v` (three new tests pass)
- [ ] `pytest tests/test_generate_model_catalog.py -k reasoning_type -v` (three new generator tests pass)
- [ ] `pytest tests/test_llm_invoke.py -v` (no regression in existing tests)
- [ ] `/gcbrun` cloud-test passes
- [ ] After merge + 0.0.249 release + downstream redeploy: verify PROD fixcode returns success on POST with model=claude-opus-4-7 and no `Please visit github.com/login/device` warnings appear in Cloud Run logs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1158, opened 2026-05-24
fix(llm_invoke): fast-fail on permanent invalid_request_error instead of cascading

gltanaka:
## Summary

When a candidate model returns a 400 with a permanent parameter rejection (e.g. ``"thinking.type.enabled" is not supported for this model``), every other candidate that proxies the same model class rejects the same shape identically. Instead of walking through Vertex/Bedrock/Azure/OpenRouter/Perplexity/github_copilot and getting the same 400 each time (then hanging on github_copilot device-flow), fast-fail and surface the root error.

Adds ``_is_permanent_invalid_request_error(exc)`` classifier (conservatively allow-listed phrases — ``"is not supported for this model"``, ``"unsupported parameter"``, ``"thinking.type"``, ``"output_config.effort"``). The candidate-iteration loop in ``llm_invoke()`` consults it and re-raises on True.

``ContextWindowExceededError`` (subclass of BadRequest) is explicitly retryable — a model with a larger context may succeed. Everything else falls through to existing retry logic.

## Background — what motivated this

Anthropic enforced ``thinking.type.adaptive`` for Opus 4.7 on 2026-05-23 ~17:25 UTC. The legacy ``thinking.type.enabled`` shape started 400-ing. The candidate cascade walked through every Anthropic-relay provider and got identical 400s, then hit ``github_copilot/*`` rows which triggered litellm's device-flow OAuth — hung 10+ minutes per fix request in PROD. Cascading on a permanent parameter error is wasted work; fast-failing surfaces the root cause immediately. Companion to PR #1156 (which fixed the specific Opus 4.7 CSV row).

## Test plan

- [ ] ``pytest tests/test_llm_invoke.py -k permanent_invalid_request -v`` (new classifier tests pass)
- [ ] ``pytest tests/test_llm_invoke.py -v`` (no regression in existing tests)
- [ ] ``/gcbrun`` cloud-test passes
- [ ] Spot-check: with a deliberately-malformed parameter, ``llm_invoke()`` raises the original ``BadRequestError`` instead of returning ``all_candidate_models_failed``


gltanaka:
## Round 1 Review — Request Changes

Two blocking issues before merge.

---

**1. `pdd/llm_invoke.py:2060` — `"unsupported_parameter"` (underscored) is dead code with no known emitter**

The allow-list includes both `"unsupported parameter"` (space) and `"unsupported_parameter"` (underscore). Neither litellm source nor openai source emits the underscored form. A scan of litellm's installed source confirms zero matches for the token `unsupported_parameter` in any `.py` file. It is harmless (allow-list is conservative-only), but it signals incomplete vetting of the phrase list.
**Required action:** Either cite a real provider/error-code reference that emits `unsupported_parameter` exactly as a substring of a 400 message body, or remove the underscored form. If no citation exists, remove it.

---

**2. `tests/test_llm_invoke.py` — No loop-level integration test; a low-cost harness exists**

The six new tests cover `_is_permanent_invalid_request_error()` in isolation. They do not test that the `if _is_permanent_invalid_request_error(e): raise` guard at line 4646 actually fires correctly inside the candidate-iteration loop — that is, that `llm_invoke()` raises `BadRequestError` (not `RuntimeError("All candidate models failed")`) when the first candidate returns a permanent 400. A regression could silently revert to the old cascade behavior with all classifier tests still green.

A low-cost harness exists: mock `litellm.completion` to raise `litellm.BadRequestError` with a permanent message, call `llm_invoke()` with a minimal multi-candidate CSV stub, and assert the raised exception is `BadRequestError` (not `RuntimeError`). The existing test file already has mock seams for `litellm.completion` (e.g. `side_effect` patterns in nearby tests). This is not gnarly.
**Required action:** Add one integration-level test that calls `llm_invoke()` (with `litellm.completion` monkeypatched to raise a permanent `BadRequestError`) and asserts the exception propagates as `BadRequestError` rather than being swallowed into `all_candidate_models_failed`.

---

**Non-blocking findings (no changes required)**

- **Context-window safety**: confirmed. `ContextWindowExceededError` is caught by the earlier dedicated `except litellm.ContextWindowExceededError` handler at line 4563 and never reaches the broad `(openai.BadRequestError, ..., Exception)` block. The `isinstance(exc, ctx_err)` guard inside the classifier is belt-and-suspenders; correct as-is.
- **Re-raise correctness**: confirmed. The `for model_info in candidate_models` loop at line 3296 is not wrapped in any outer `try/except`. The bare `raise` at line 4651 propagates cleanly to all callers. All callers already handle exceptions (all wrap `llm_invoke()` in `except Exception`), so no backward-compat breakage.
- **MRO chain**: confirmed. `litellm.BadRequestError → openai.BadRequestError → openai.APIStatusError → Exception`. The existing `except (openai.BadRequestError, ..., Exception)` handler catches it, and `isinstance(exc, litellm.BadRequestError)` passes inside the classifier.
- **Logging**: `[FAST-FAIL]` log omits the original error message text. The preceding `[ERROR]` line already emits it, so context is recoverable by reading two log lines together. Acceptable.
- **Temperature-retry interaction**: if the error contains both `"thinking"` and `"temperature"`, the temperature-retry `continue` fires first (before fast-fail). Fast-fail only fires after `temp_adjustment_done=True`. This ordering is correct.
- **Caller backward compat**: all callers (`fix_errors_from_unit_tests.py`, `fix_code_module_errors.py`, `agentic_common.py`, etc.) wrap `llm_invoke()` in `except Exception` already. The new raise is a different exception type but flows through the same path.
- **PR #1156 reference**: confirmed open (`promptdriven/pdd#1156`, `fix(opus-4.7): flip CSV to adaptive + remove PDD_FORCE gate`). Reference is accurate.
- **Other call sites**: only one candidate-iteration loop in `llm_invoke.py` catches `BadRequestError`. No other cascade loops need the same guard.

gltanaka:
Both Round 1 blockers addressed in `ccc532c0e`:

**1. `"unsupported_parameter"` (underscored) removed from allow-list** — verified codex's claim independently: zero hits in installed litellm, openai SDK only mentions "unsupported parameters" (plural with space, in docstrings — not in error messages). Speculative marker dropped. Existing `"unsupported parameter"` (space form) covers the real cases. Updated the corresponding allow-list-coverage test to drop the underscored phrase too.

**2. Integration test added** — `test_llm_invoke_fast_fails_on_permanent_invalid_request_error`. Monkeypatches `litellm.completion` to raise a permanent `BadRequestError`, calls `llm_invoke()` against the existing 3-candidate `mock_load_models` fixture, asserts (a) the original `BadRequestError` propagates out of `llm_invoke()` rather than the cascade returning `all_candidate_models_failed`, and (b) `mock_completion.call_count == 1` — proves the fast-fail prevented cascading through the other 2 candidates. The "gnarly fixtures" pushback in my original report turned out to be wrong; the existing `test_llm_invoke_valid_input` pattern handles all the DataFrame/credentials/cache plumbing cleanly.

Dropped from the list: nothing — both items were legitimate.

Test totals: 7 classifier-related tests pass (6 unit + 1 new integration), full `test_llm_invoke.py` 299→300 passed, no regression.

Ready for Round 2 review.

gltanaka:
**LGTM — both Round 1 blockers resolved cleanly.**

Verified all five Round 2 checklist items against the source at `ccc532c0e`:

- `permanent_markers` in `_is_permanent_invalid_request_error` now contains exactly the four intended phrases (`"is not supported for this model"`, `"unsupported parameter"`, `"thinking.type"`, `"output_config.effort"`); the underscored `"unsupported_parameter"` form is gone.
- The new `test_llm_invoke_fast_fails_on_permanent_invalid_request_error` integration test correctly uses `side_effect` (not `return_value`) so `litellm.completion` actually raises; asserts the original `litellm.BadRequestError` propagates; and asserts `mock_completion.call_count == 1` against the 3-candidate `mock_load_models` fixture — proving the fast-fail prevents cascade.
- Drift commit `18f3114fb` between rounds is metadata-only (`.pdd/meta/` updates), no functional regression.

Ready for merge.

(Posting on behalf of codex review subagent, which approved Round 2 but couldn't reach api.github.com from its sandbox.)

PR #1193, opened 2026-05-26
fix(opus-4-7): patch LiteLLM _is_claude_opus_4_5 so Vertex/Bedrock/Azure use adaptive thinking

gltanaka:
## Summary

- Patch `AnthropicConfig._is_claude_opus_4_5` at `pdd.llm_invoke` import to also match `opus-4-7`, so every Anthropic-relay provider routes Opus 4.7 to the new adaptive thinking shape.
- Closes the deferred audit from PR #1156 (CSV flip was direct-Anthropic only; Vertex/Bedrock/Azure left at the legacy shape).
- Patch is idempotent so `importlib.reload(pdd.llm_invoke)` doesn't recurse.

## Repro (before fix)

```
LiteLLM completion() model= claude-opus-4-7; provider = vertex_ai
Vertex_aiException BadRequestError - "thinking.type.enabled" is not supported for this model.
Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.
[FAST-FAIL] Permanent invalid_request_error from vertex_ai/claude-opus-4-7
```

Failing pdd-cli LLM smoke today (pdd_cloud `test-pdd-cli-llm-smoke-cloud`) with `vertex_ai/claude-opus-4-7` selected by the candidate cascade.

Root cause: LiteLLM's `AnthropicConfig._is_claude_opus_4_5` hardcodes `"opus-4-5"`. `VertexAIAnthropicConfig` inherits it (verified via MRO), so every Anthropic-relay provider sends `thinking.type.enabled` for Opus 4.7. Provider responds with 400 and the FAST-FAIL classifier (correctly) refuses to cascade through more relays that would all 400 the same way.

## Test plan

- [x] new regression tests (2): `_is_claude_opus_4_5` matches opus-4-7 + opus-4-5 across name prefixes; not matching unrelated models; patch propagates to `VertexAIAnthropicConfig`.
- [x] full `test_llm_invoke.py` (305 tests) — passes including the reload-fixture tests that originally tripped the recursion.
- [x] adjacent llm_invoke suites (`csv_model_registration`, `nested_schema`, `retry_cost`, `vertex_retry`) — 331 total green.
- [ ] downstream: pdd_cloud `make test-pdd-cli-llm-smoke-cloud` rerun against this branch (see linked issue / will rerun once merged).

## Risks / out-of-scope

- The CSV rows for `vertex_ai/claude-opus-4-7`, `azure_ai/claude-opus-4-7`, and `AWS Bedrock/anthropic.claude-opus-4-7` still carry `reasoning_type=effort`. That stays the right choice while LiteLLM owns the translation — the patch just teaches LiteLLM that opus-4-7 deserves the adaptive shape. A future LiteLLM release with native opus-4-7 matching makes this patch a no-op and the `try:` becomes removable.
- No change to direct-Anthropic Opus 4.7 (CSV is already `adaptive`, code path already correct) — the patch is a no-op for that row.


gltanaka:
## Codex review round 1 — resolution table

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | blocker | `map_openai_params` overwrites `thinking` with the enabled shape even when `_is_claude_opus_4_5` matches; predicate patch alone is insufficient. | **Fixed in ebeeea98.** Added a layered patch that wraps `map_openai_params` and replaces the stale `thinking.type.enabled` with `thinking.type.adaptive` whenever both `output_config` and the predicate match. Verified empirically via direct call — final shape is now `{thinking: {type: adaptive}, output_config: {effort: X}}` for Vertex/Bedrock Opus 4.7. Caller's richer adaptive payload (e.g. `display: summarized`) is preserved. |
| 2 | major | CSV rows for Vertex/Bedrock/Azure not flipped to `adaptive`; patch description over-claimed coverage. | **Description narrowed.** Vertex/Bedrock CSV rows stay at `effort` — correct, since pdd emits `reasoning_effort` and the new `map_openai_params` wrap fixes the downstream LiteLLM bug. Azure flagged as out of scope (see #3). |
| 3 | major | Azure AI uses `AzureAIStudioConfig` (OpenAI-based), not `AnthropicConfig`; the monkey-patch does not reach the Azure adapter. | **Acknowledged.** Updated the patch docstring to explicitly call out that Azure is not covered. Azure Opus 4.7 needs a separate code-path fix (CSV is currently `budget` so pdd sends `thinking.type.enabled` directly without going through `_map_reasoning_effort`). Tracked as a follow-up. |
| 4 | major | `pdd_cloud` (`backend/functions/llm_invoke.py` + `generate_code.py`) wraps `_is_claude_opus_4_5` without a sentinel — could double-wrap. | **Out of scope (cross-repo).** This PR is in `promptdriven/pdd`; the wrappers in `pdd_cloud` are independent. The smoke test (`test-pdd-cli-llm-smoke-cloud`) installs pdd-cli from this repo, not from pdd_cloud — so there is no cross-impact for the failing test. Will open a follow-up PR in pdd_cloud to mirror the sentinel. |
| 5 | minor | Tests only assert the predicate result, not the actual transformed kwargs. | **Fixed in ebeeea98.** Added 4 transform-level tests covering: vertex-opus-4-7 effort-only → adaptive, bedrock-opus-4-7 effort-only → adaptive, direct-opus-4-7 adaptive-preserve (with `display` field), and sonnet-4-6 unchanged regression-guard. |

All 335 llm_invoke tests pass (was 331 + 4 new transform-level + 2 predicate already counted). Re-dispatching codex for round 2.

gltanaka:
## Codex review round 2 — resolution table

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | major | Round-2 wrap only covers `AnthropicConfig` (and its subclasses Vertex + Bedrock-invoke). Bedrock **Converse** uses `AmazonConverseConfig`, which doesn't inherit from `AnthropicConfig` — its own `map_openai_params` calls `AnthropicConfig._map_reasoning_effort` directly with no Opus 4.5/4.7 branch and always emits `thinking.type.enabled` for Anthropic Converse models including opus-4-7. | **Fixed in d0f9d60a.** Added a third sentinel-guarded wrap targeting `AmazonConverseConfig.map_openai_params`. Mirrors the round-2 wrap (post-process the result; if model name contains opus-4-7 and result carries the legacy enabled shape, swap to adaptive, preserving caller's richer payload). Gated on substring directly since Converse has no `_is_claude_opus_4_5` helper. |
| 2 | minor | Round-2 Bedrock tests instantiate generic `AnthropicConfig` with an `anthropic.*` model name; they don't actually exercise Bedrock's own adapter classes, so they wouldn't catch a Converse regression. | **Fixed in d0f9d60a.** Added 3 tests that instantiate the actual Bedrock classes: `AmazonAnthropicClaudeConfig` (invoke-style; covered via inheritance from `AnthropicConfig`), `AmazonConverseConfig` (covered via the new wrap), and an unrelated-models regression-guard for Converse opus-4-5 / sonnet-4-6. |

Sentinel/reload behavior verified clean in round 2 — kept the same pattern for the new converse wrap (`_pdd_opus_4_7_converse_patched`).

All 338 llm_invoke tests pass (was 335 + 3 new Bedrock-adapter tests). Re-dispatching codex for round 3.

gltanaka:
## Codex review round 3 — resolution table

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | major | Converse wrap drops `reasoning_effort` on the floor when rewriting `thinking` to bare `{type: adaptive}`. AWS Bedrock Converse for Claude flattens adaptive thinking into a single key (no `output_config` sibling), so effort must travel inside the thinking dict. | **Fixed in 43753f17.** Updated the Converse wrap to (a) preserve the caller's adaptive payload — `display`, `effort`, etc. — if they supplied one, (b) otherwise rebuild `{type: adaptive}` and copy `reasoning_effort` into the `effort` field. `time_to_effort_level()` now controls thinking depth on Converse Opus 4.7 the same way it does on direct/Vertex via `output_config`. Tests updated to pin the `effort` field; added a second Converse test for caller-supplied-payload preservation. |
| 2 | major | `DatabricksConfig.map_openai_params` is yet another LiteLLM adapter that bypasses `AnthropicConfig` and would emit `thinking.type.enabled` for `databricks/claude-opus-4-7`. | **Out of scope.** Confirmed via `grep -i databricks pdd/data/llm_model.csv` — the stock pdd-cli ships **no** Databricks Claude rows. To hit `DatabricksConfig` with Opus 4.7, a user would have to register their own CSV row pointing at `databricks/claude-opus-4-7`. The failing smoke test (`test-pdd-cli-llm-smoke-cloud`) routes through the stock Vertex AI row — Databricks is not on the failing path. Documented as a follow-up in the commit message; the same wrap pattern from round 3 applies cleanly if/when Databricks Claude is added to the stock catalog. |

All 339 llm_invoke tests pass (was 338 + 1 new + 1 updated). Re-dispatching codex for round 4.

Note: per `docs/contributors/pr-loop-process.md`, codex review caps at 4-5 rounds before escalating to human reviewer. This is round 4. If codex still REQUEST_CHANGES on a non-blocking finding, I will escalate.

Issue #1006, opened 2026-05-14
Enforce metadata finalization in CI auto-heal / preflight drift-heal paths


Serhan-Asad:
## Problem
CI auto-heal and preflight drift-heal can invoke `pdd update <code>` without a guaranteed metadata-finalization step. A successful prompt update in CI may push code with stale `.pdd/meta/` state, which downstream commands then interpret as partially synced.

## Files
- `pdd/ci_drift_heal.py`
- `pdd/prompts/ci_drift_heal_python.prompt`
- `.github/workflows/auto-heal.yml`
- `.pdd/meta/ci_drift_heal_python.json`
- `tests/test_ci_drift_heal.py` (regression test)

## Desired Behavior
- CI auto-heal and preflight drift-heal must invoke a metadata-finalization step after a successful prompt/code update, or fail loudly if finalization cannot run.
- A successful auto-heal commit must include up-to-date `.pdd/meta/` changes alongside the prompt/code changes — never one without the other.
- If finalization fails, the workflow must mark the job as failed with a clear error rather than commit partial state.

## Acceptance Criteria
- After a successful auto-heal run, `.pdd/meta/` for affected modules is committed in the same PR/commit as the prompt/code change.
- A finalization failure causes the workflow step to fail (non-zero exit) with an explicit message.
- Regression test in `tests/test_ci_drift_heal.py` covers the success path (metadata committed) and the failure path (workflow fails loudly).

## Notes
Reuse the same fingerprint finalization API used by `pdd update <code>` and `pdd auto-deps`. Keep behavior language-agnostic.

---
## Split Contract
**Command sequence:** change → sync
**Allowed write set:**
- `pdd/ci_drift_heal.py`
- `pdd/prompts/ci_drift_heal_python.prompt`
- `.github/workflows/auto-heal.yml`
- `tests/test_ci_drift_heal.py`
**Acceptance criteria:**
- Successful CI auto-heal commits up-to-date `.pdd/meta/` alongside the prompt/code change.
- Finalization failure fails the workflow step with an explicit error (no silent partial commit).
- Regression test in `tests/test_ci_drift_heal.py` covers success-with-metadata and finalization-failure paths.
**Independently mergeable:** True
**Scope rule:** Do not expand beyond this contract or implement sibling sub-issue work. If the contract is insufficient, report the gap instead.

---
**PDD Command Hint:** This is a bug fix. Use `bug → fix` (create failing tests, then fix the code).

---
[...]

PR #1217, opened 2026-05-26
Fix #1198: warn on unknown keys in .pddrc with full schema validation

### LINKED ISSUE #1198: .pddrc silently ignores unknown keys (environmental variables); no schema validation
**Summary**

PDD reads .pddrc, silently drops unrecognized top-level keys, and falls back to built-in defaults without warning. There is no pdd config validate command and no startup warning. Users who hand-write a config based on stale documentation get no signal that their configuration is being ignored.

**Reproduction**
1. Create a .pddrc using incorrect keys 
2. Run pdd sync <module>.
3. Observe that output paths in .pddrc are not honored, and no warning is emitted.
4. Run pdd setup, it produces a config with the actual current schema

**Expected**

Either (a) warn on unknown keys at config load time, (b) fail loudly with a schema error and a pointer to pdd setup, or (c) provide a pdd config validate subcommand.

**Impact**

Roughly an hour of debugging per affected user. New users following older docs will silently get default behavior and assume their config "works."

**Suggested fix**

Add a strict-mode JSON-schema validator over .pddrc at load time. Emit WARNING: .pddrc contains unknown key 'X' at path 'Y', ignored. Run 'pdd setup' to regenerate. Optionally add pdd config validate for explicit checks.

sohni-tagirisa:
Closes #1198.

## Problem

`.pddrc` accepts unknown keys silently. Typos, stale documentation, and leftover keys from older schemas are dropped without warning, and PDD falls back to built-in defaults. Per the issue, this costs users ~1 hour of debugging per affected case.

## Fix

Add schema-based validation in `_load_pddrc_config` that emits a `UserWarning` for any unknown key at the root, context, or defaults levels. Warning format matches the issue's specification: WARNING: .pddrc contains unknown key 'X' at path 'Y', ignored. Run 'pdd setup' to regenerate.

Unknown keys are reported but not rejected, so stale configs continue to load — avoids breaking users on upgrade. Could be promoted to errors in a future release.

## Schema

Known keys by nesting level:

- **Root**: `version`, `contexts`
- **Context**: `paths`, `defaults`
- **Defaults**: `generate_output_path`, `test_output_path`, `example_output_path`, `prompts_dir`, `default_language`, `target_coverage`, `strength`, `temperature`, `budget`, `max_attempts`, `outputs`

Schema verified by tracing `pdd/construct_paths.py`'s `_resolve_config_hierarchy` to identify keys that are actually consumed.

## Tests

7 new tests in `tests/test_construct_paths.py`:

- Unknown root key emits warning
- Unknown context key emits warning
- Unknown defaults key emits warning
- Clean config emits no warnings
- Warning message format matches issue specification
- Validator warns on `auto_deps_csv_path` (regression test)
- Validator warns on `prompt_path` (regression test)

## Findings surfaced by the validator

Running the validator against the codebase caught four real issues:

**1. Test fixtures with incorrectly-nested keys**

Two tests in `TestConstructPathsResolutionModeParameter` wrote `.pddrc` files with path keys directly under the context block (e.g., `contexts.default.generate_output_path`) rather than inside `defaults`. Dead config — the resolver ignores those keys at the context level. Fixed in this PR.

**2. `auto_deps_csv_path` prescribed by templates but not consumed**

PDD's template at `pdd/templates/generic/generate_pddrc_YAML.prompt` prescribed `auto_deps_csv_path` in every context across 9 occurrences. The repo's own production `.pddrc` and `examples/prompts_linter/.pddrc` (5 contexts each) followed this convention, and the example README instructed users to add the key. But `_resolve_config_hierarchy` never reads it — `auto_deps_main` falls back to the default `"project_dependencies.csv"`.

Per PR review by @gltanaka: removed from the validator's allow-list, from the template (9 occurrences), from the prompt's schema doc, and from `examples/prompts_linter/.pddrc` and `README.md` (10 occurrences total). Following up with a separate issue for the actual wiring work (~6 touchpoints in path-resolution machinery).

**3. `match` fixture in `test_find_prompt_file.py:424`**

Validator surfaced `contexts.backend-utils.match` as an unknown key. Investigation showed `match` is not a real PDD schema key (the test's assertion only verifies `prompts_dir`). Removed from the fixture.

**4. `prompt_path` documented but not consumed**

Per PR second review by @gltanaka: `prompt_path` was in `_PDDRC_DEFAULTS_KEYS` because `construct_paths_python.prompt` documented it as an alias for `prompts_dir`, but `_resolve_config_hierarchy` never implemented the documented resolution chain. Same class of silent-ignore as `auto_deps_csv_path`.

Took the 'remove' option: stripped `prompt_path` from `_PDDRC_DEFAULTS_KEYS`, removed the alias claim from `construct_paths_python.prompt` (the prompt was the source-of-truth bug, not the code), and added a regression test. `PDD_PROMPT_PATH` env var is unaffected — only the `.pddrc` key alias was removed.

## CI

Local filtered suite passes with the same failure count as `main` (verified with `git stash` + branch checkout). No new `unknown key` warnings emitted by the validator anywhere in the suite. Loading the repo's checked-in `.pddrc` and `examples/prompts_linter/.pddrc` both produce zero warnings.

## Files changed

- `pdd/construct_paths.py` — validator function, schema constants, call site
- `pdd/prompts/construct_paths_python.prompt` — validation requirement documented, removed `prompt_path` alias claim
- `pdd/templates/generic/generate_pddrc_YAML.prompt` — removed `auto_deps_csv_path` prescription
- `examples/prompts_linter/.pddrc` — removed `auto_deps_csv_path`
- `examples/prompts_linter/README.md` — removed `auto_deps_csv_path` instructions
- `tests/test_construct_paths.py` — 7 new tests + 2 fixture corrections
- `tests/test_find_prompt_file.py` — 1 fixture correction (match key)


gltanaka (changes requested):
Requesting changes rather than merging as-is.

The validator itself is needed for #1198; unknown `.pddrc` keys currently fail silently. The blocker is that this implementation whitelists and documents `auto_deps_csv_path` as a valid defaults key (`pdd/construct_paths.py:132-145`, `pdd/prompts/construct_paths_python.prompt:51-54`) while the runtime still does not consume that key from `.pddrc`.

Current flow:
- `_resolve_config_hierarchy` only resolves the existing path/model keys and does not include `auto_deps_csv_path` (`pdd/construct_paths.py:530-542`).
- `generate_output_paths.COMMAND_OUTPUT_KEYS['auto-deps']` only includes `output`, so no `csv` output path is produced (`pdd/generate_output_paths.py:18-33`).
- `auto_deps_main` then reads `output_file_paths.get("csv", "project_dependencies.csv")`, so the `.pddrc` key remains silently ignored (`pdd/auto_deps_main.py:56-73`).

1. Either wire `auto_deps_csv_path` through config resolution/output path generation and add regression coverage proving `.pddrc` affects the auto-deps CSV path, or remove it from the allow-list/template so users get the unknown-key warning instead of another silent ignore.
2. Clean the new trailing whitespace reported by `git diff --check` at `pdd/construct_paths.py:206`.

Tests I ran: `python -m pytest tests/test_construct_paths.py -q` passed; `git diff --check origin/main...HEAD` failed on the trailing whitespace above.

gltanaka:
Additional required cleanup from the broader non-integration run: the new validator emits a warning for `contexts.backend-utils.match` from `tests/test_find_prompt_file.py:424`. If `match` is an unsupported/stale .pddrc shape, please fix that fixture; if it is intended to be supported, add it to the schema and cover the resolver behavior. Either way, the validator change should not leave the suite with a new unknown-key warning.

gltanaka (changes requested):
Thanks for addressing the earlier review. The original blockers are fixed: `auto_deps_csv_path` was removed from `_PDDRC_DEFAULTS_KEYS`, the generator template no longer emits it, the `match` fixture was cleaned up, and `git diff --check` is now clean.

I still would not merge this current head because the validator now exposes checked-in stale config/docs and another allow-listed silent ignore:

1. `examples/prompts_linter/.pddrc` still contains `auto_deps_csv_path` in five contexts, and `examples/prompts_linter/README.md` still instructs users to add the same key. Loading the repo's checked-in `.pddrc` files on this branch emits five new warnings from that example config. Please remove/update those entries too, otherwise a public example in the repo immediately warns under the new validator.

2. `prompt_path` is still included in `_PDDRC_DEFAULTS_KEYS`, but `_resolve_config_hierarchy` does not resolve it; it only resolves `prompts_dir` and the other existing keys. I verified with a `.pddrc` containing `defaults.prompt_path: configured_prompts`: `_load_pddrc_config` emits no warning, but `resolve_effective_config()` returns only `default_language` and `_matched_context`. This is the same class of silent-ignore hole that was fixed for `auto_deps_csv_path`. Please either wire `prompt_path` through the resolver as the documented alias for `prompts_dir` with regression coverage, or remove it from the allow-list/schema docs until it is actually consumed.

Focused verification I ran:
- `git diff --check origin/main...HEAD`: passed
- `python -m pytest tests/test_construct_paths.py -q`: 111 passed
- `python -m pytest tests/test_find_prompt_file.py::TestPddrcPrefixForNewModules::test_new_module_respects_pddrc_prompts_dir_prefix -q`: 1 passed
- loading all checked-in `.pddrc` files still emits `auto_deps_csv_path` warnings from `examples/prompts_linter/.pddrc`

sohni-tagirisa:
Pushed updates addressing both points from your second review.

**1. `examples/prompts_linter` cleanup** (`a9429e12`) — removed all 5 occurrences of `auto_deps_csv_path` from `examples/prompts_linter/.pddrc` and all 5 corresponding lines from the README.

**2. `prompt_path`** (`197ac26f`, `ec741626`) — took the 'remove' option. Stripped `prompt_path` from `_PDDRC_DEFAULTS_KEYS` and from `construct_paths_python.prompt`'s alias documentation (the prompt was the source-of-truth bug — it claimed an aliasing system that was never implemented). Added a regression test asserting the validator warns on the key. `PDD_PROMPT_PATH` env var is unaffected.

Verified locally with your exact checks:

- `git diff --check origin/main...HEAD` → clean
- `python -m pytest tests/test_construct_paths.py -q` → 114 passed
- `python -m pytest tests/test_find_prompt_file.py::TestPddrcPrefixForNewModules::test_new_module_respects_pddrc_prompts_dir_prefix -q` → 1 passed
- Loading `.pddrc` and `examples/prompts_linter/.pddrc` — both produce **zero warnings**

PR description updated. Ready for re-review.

gltanaka (changes requested):
Thanks for the update. The `.pddrc` validator issues from the previous review look addressed now: `examples/prompts_linter` no longer carries `auto_deps_csv_path`, `prompt_path` is no longer documented as supported without wiring, and the focused tests pass locally.

I still would not merge this head because unrelated generated/local artifacts are now included in the PR:

- `pdd-env/` commits a local virtualenv, including activation scripts, console entrypoints, Python symlinks, and `pdd-env/pyvenv.cfg` with contributor-local paths such as `/Users/sohnitagirisa/pdd/pdd-2/pdd-env`.
- `git diff --check origin/main...HEAD` fails on the committed virtualenv files, starting with trailing whitespace throughout `pdd-env/bin/Activate.ps1`.
- `error_log.txt` is added and appears to be a local generated test/debug artifact.
- `.gitignore` removes the committed `.pdd/meta/...` exceptions; that looks unrelated to issue #1198 and should not be included unless there is a deliberate repo-level reason.

1. Remove `pdd-env/` from the PR entirely and keep local virtualenvs ignored.
2. Remove `error_log.txt` unless this PR intentionally needs it and the purpose is documented.
3. Revert the unrelated `.gitignore` deletion, or split/explain it separately if it is intentional.
4. Keep the PR scoped to the `.pddrc` validation, template/example cleanup, and regression tests.

Local checks run on head `34034fb4`:

- `python -m pytest tests/test_construct_paths.py -q` -> `114 passed`
- `python -m pytest tests/test_find_prompt_file.py::TestPddrcPrefixForNewModules::test_new_module_respects_pddrc_prompts_dir_prefix -q` -> `1 passed`
- `git diff --check origin/main...HEAD` -> fails because of the committed `pdd-env/` files

gltanaka (changes requested):
The previous artifact blockers are mostly cleaned up now: `pdd-env/` and `error_log.txt` are no longer committed as files, `git diff --check origin/main...HEAD` is clean, CI is green, and the focused `.pddrc` validation tests pass locally.

I still would not merge this head because `.gitignore` still contains unrelated and potentially harmful deletions. The PR removes existing ignore/allowlist rules and comments such as:

- `pdd/_version.py`, the setuptools-scm generated version file ignore
- `.pdd/meta/*.json`, which changes the ignore behavior for generated `.pdd/meta` JSON files
- several explicit `.pdd/meta/...` allowlist entries
- the Issue #1006 run-report comment and run-report exceptions
- `.loop/`

Those changes are not part of issue #1198 and could change what local/generated files are picked up by git. Adding ignores for the accidental local artifacts (`pdd-env/`, `error_log.txt`) is fine, but the existing `.gitignore` behavior should be preserved.

Required change before merge:

1. Restore the unrelated `.gitignore` deletions and keep the PR scoped to the `.pddrc` validation/template/example/test changes. If you want to keep the new `pdd-env/` and `error_log.txt` ignore entries, add them without removing the existing rules.

Local checks run on head `89828b0`:

- `git diff --check origin/main...HEAD` -> clean
- `python -m pytest tests/test_construct_paths.py -q` -> `114 passed`
- `python -m pytest tests/test_find_prompt_file.py::TestPddrcPrefixForNewModules::test_new_module_respects_pddrc_prompts_dir_prefix -q` -> `1 passed`
- loading checked-in `.pddrc` files via `_load_pddrc_config` -> no warnings

gltanaka (approved):
The latest update resolves my remaining blockers.

What I verified on head `f548a79`:

- The committed `pdd-env/` and `error_log.txt` artifacts are gone.
- `.gitignore` now only appends ignores for `pdd-env/` and `error_log.txt`; the existing `.pdd/meta` fingerprint/run-report rules are preserved.
- `git diff --check origin/main...HEAD` is clean.
- `python -m pytest tests/test_construct_paths.py -q` -> `120 passed`.
- `python -m pytest tests/test_find_prompt_file.py::TestPddrcPrefixForNewModules::test_new_module_respects_pddrc_prompts_dir_prefix -q` -> `1 passed`.
- Loading checked-in `.pddrc` files through `_load_pddrc_config` produced no unknown-key warnings.

The `.pddrc` warning behavior looks scoped and useful now, and the previous PR hygiene issues are fixed. I think this is reasonable to merge once branch protection is satisfied.

gltanaka (approved):
Re-approved after updating the branch with main. Verified updated head 594f702 locally: diff remains scoped, git diff --check is clean, tests/test_construct_paths.py passes, and the targeted prompt-dir regression passes.

PR #1227, opened 2026-05-27
perf(cloud-batch): split heavy test files at class granularity + faster poll

gltanaka:
## Summary

Two changes to `ci/cloud-batch/` that together cut Cloud Batch wall time:

1. **Per-class/per-test splitting** in `balance-chunks.py` for files exceeding `--heavy-threshold` (default 300s). Files emit per-unit IDs (`file::ClassName` or `file::test_func`) instead of the whole path; the greedy bin packer treats each unit independently. New `seed` subcommand AST-scans heavy files and distributes their existing file-level total across classes/module-level tests, so **the very first run** after this lands gets the benefit.

2. **`submit.sh POLL_INTERVAL` 10s → 5s.** Polling dominates the tail once both batch jobs are near-complete.

## Predicted impact

From a simulation over the checked-in `test-durations.json` (32 chunks, greedy packer):

| metric | baseline (file-only) | with splitting |
|---|---|---|
| pytest chunk max | **558s** | **450s** (-19%) |
| pytest chunk min | 445s | 450s |
| imbalance ratio | 1.25x | **1.00x** |
| balance units | 359 | 1145 |

Most recent observed `cloud-test` wall on `main`: 560s, with `chunk_21` at 737s (1.6x). With per-class data accumulating from real junitxml records, real-world tail should compress toward the 450s floor.

## Repro (predicted improvement)

```
python3 ci/cloud-batch/balance-chunks.py seed --test-dir tests/ \
    --durations ci/cloud-batch/test-durations.json
# (already applied in this PR — 792 per-unit entries seeded across 6 heavy files)

python3 ci/cloud-batch/balance-chunks.py assign --chunk-index 0 --num-chunks 32 \
    --durations ci/cloud-batch/test-durations.json --test-dir tests/
# emits a mix of tests/foo.py and tests/foo.py::TestBar
```

## Test plan

- [x] new `tests/test_balance_chunks.py`: 12 unit tests covering `_split_unit_id` (module-level / class / nested class / parametrize / mismatch), `_build_units` (light / heavy with subunits / heavy without subunits / unknown file), `_greedy_pack_units` (no duplicate coverage of split files), and `cmd_seed` (per-class distribution + threshold gating). All pass.
- [x] `pytest --collect-only --strict-markers --strict-config tests/ -k __nonexistent__` clean — same preflight `entrypoint.sh` runs (9761 tests deselected, no collection errors).
- [x] Verified: across all 32 chunks at num-chunks=32, all 1145 emitted units are unique — no heavy file appears as both `F` and `F::X` in the same packing.
- [ ] Cloud Batch end-to-end run (will trigger via `/gcbrun` after PR opens).

## Backward compatibility

When `test-durations.json` contains only file-level entries (the prior format), `_build_units` emits files whole — identical to legacy behavior. Pinned by `test_build_units_heavy_file_without_subunits_emits_whole`. The seeded data file in this PR adds per-unit entries for 6 files only (the heavies); all 359 existing file-level entries are preserved unchanged.

## Risks / out-of-scope

- **Class splitting changes pytest module-import behavior** for split files only: a single chunk that previously had file F now has only F's TestA; F's TestB lives in another chunk. Each chunk now imports F once instead of zero/one. Acceptable cost (~import time of F).
- **Files using cross-class module-level state** (e.g., a module-scoped fixture two classes both rely on) will see that fixture instantiated in both chunks. Per-class fixtures unaffected.
- The `seed` step's per-class weights assume uniform per-test cost within a file. Real durations land on the next `record` run, which corrects any drift.
- Out of scope: changes to llm-smoke (`pdd_cloud`), image cache strategy, additional chunk count tuning.


gltanaka:
LGTM. All verification checks pass: 12/12 unit tests green, no double-coverage across all 32 chunk assignments, seed drift within 0.4s across all heavy files. The `_build_units` split gate is correct — a file is emitted either whole or as sub-units, never both. `cmd_record` prunes stale sub-unit keys via the `fresh_files` filter. Nested class node ID syntax (`::` not `.`) matches pytest node ID convention. `_filepath_from_classname` handles subdirectory paths correctly. `POLL_INTERVAL=5` poses no rate-limit risk at this call frequency. Approving.

gltanaka:
Benchmark + follow-up fix in `6f26a9a19`:

**Wall-clock result (`make test-pdd-cli-release` from pdd_cloud, against this PR's worktree):**
| | baseline | this PR | delta |
|---|---|---|---|
| cloud-test (cloud-batch wall) | 560s | **385s** | **-31%** |
| pytest chunk max (worst tail) | 737s (chunk_21) | needs re-bench after `6f26a9a19` | — |
| cloud-test exit | 77/77 PASS | 75/77 (2 latent flakes — fixed in `6f26a9a19`) | — |

**Latent test leak surfaced by the rebalance:** The 2 cloud-batch failures (`test_preprocess.py::test_unresolved_include_*`) trace back to `tests/test_quiet_flag.py`: the `pdd --quiet` CLI group callback writes `os.environ["PDD_QUIET"] = "1"` and never unwinds it. The prior file-only packing happened to keep these tests in different chunks; the heavy-file splitter co-located them, exposing the leak. Defended via the existing `isolate_cloud_only_overrides` autouse fixture in `tests/conftest.py` (`6f26a9a19`).

Repro of the leak (without the conftest fix):
```
pytest tests/test_quiet_flag.py tests/test_preprocess.py -k unresolved_include  # 3 fail
```
With the conftest fix:
```
pytest tests/test_quiet_flag.py tests/test_preprocess.py -k unresolved_include  # 3 pass
```

Re-benching against `6f26a9a19` to confirm green end-to-end.

gltanaka:
Gated splitter behind opt-in env var (`b5210bd5d`).

**Default mode** (no env var) emits files whole, identical to legacy packing — zero risk to existing cloud-batch runs. **Opt-in** via `PDD_BALANCE_SPLIT_HEAVY=1` activates the per-class splitter once the latent ordering flakes the rebalance exposes are addressed (one PDD_QUIET leak fixed here; others discovered would land as separate PRs).

| commit | what |
|---|---|
| `791689a43` | core splitter + POLL_INTERVAL 10→5 + seed CLI |
| `6f26a9a19` | conftest `PDD_QUIET` cleanup (real env leak, defensive in both modes) |
| `b5210bd5d` | opt-in gating + 2 gating tests (14/14 balance tests pass) |

**Verified speedup (opt-in mode, round 3 bench):** cloud-test wall **560s → 345s (-38%)**, chunk imbalance 1.6x → 1.2x.

Re-benching in default mode to confirm parity with baseline.

PR #1242, opened 2026-05-27
fix(sync): allow no-self prompt contracts (#1237)

### LINKED ISSUE #1237: Fix fix_error_loop prompt-contract preflight before sync
## Problem

`pdd sync fix_error_loop` cannot start for the prompt-only change in #1230 because dry-run validation fails during prompt-contract preflight, before generation or tests run.

The prompt `pdd/prompts/fix_error_loop_python.prompt` declares a prompt-local public interface for existing symbols already implemented in `pdd/fix_error_loop.py`:

- `cloud_fix_errors`
- `fix_error_loop`

However, the prompt does not include any existing source context from `pdd/fix_error_loop.py`. The strict prompt-contract preflight rejects that shape for a changed prompt, because the LLM would be asked to preserve existing public symbols without seeing their current implementation.

This blocks #1230 entirely: the progress-sensitive circuit-breaker change cannot be synced until this prompt-contract issue is fixed.

## Evidence

Observed in the #1230 sync job `qja8amrSnaUWz17S1FG1` after the prompt-only PR #1233 was created.

The same error was posted repeatedly at 2026-05-27 18:53 UTC:

```text
Dry-run validation failed:
fix_error_loop: prompt contract preflight failed:
- /tmp/pdd_job_qja8amrSnaUWz17S1FG1_xzspd3j8/pdd/prompts/fix_error_loop_python.prompt: prompt-local interface declares 2 public symbols already present in /tmp/pdd_job_qja8amrSnaUWz17S1FG1_xzspd3j8/pdd/fix_error_loop.py but includes no existing module source context: missing cloud_fix_errors, fix_error_loop. Use a full <include> of the existing module or select every declared existing public symbol.
```

I also reproduced the validator locally on `origin/change/issue-1230` by calling `_prompt_contract_errors_for_module("fix_error_loop", root, root)`. It returns the same missing-context error. `rg` confirms `fix_error_loop_python.prompt` has no self-include for `pdd/fix_error_loop.py`.

## Root Cause

The strict preflight path intentionally treats changed prompts differently from legacy untouched prompts. For an existing module, if the prompt itself declares public symbols that already exist in the output file, the prompt must include source context for those symbols.

Relevant behavior:

- `pdd/agentic_sync.py::_prompt_contract_strict_self_context_required()` enables strict source-context checks when the prompt changed relative to the base branch.
- `pdd/architecture_include_validation.py::validate_prompt_contract_context()` then rejects prompt-local interfaces that declare existing public symbols without a matching self-include.
- `pdd/prompts/fix_error_loop_python.prompt` declares `cloud_fix_errors` and `fix_error_loop`, but includes only dependency/example context, not `pdd/fix_error_loop.py` itself.

So this is a real prompt-contract metadata/context issue, not a model failure and not a transient CI problem.

## Required Change

Update only the prompt contract/context for `fix_error_loop` so strict preflight has source context for every declared existing public symbol.

Preferred fix:

```xml
<include select="def:cloud_fix_errors">pdd/fix_error_loop.py</include>
<include select="def:fix_error_loop">pdd/fix_error_loop.py</include>
```

Place these in the existing dependency/context section of `pdd/prompts/fix_error_loop_python.prompt` using the repo's established include style.

A full self-include is also acceptable if that is the local convention:

```xml
<include>pdd/fix_error_loop.py</include>
```

Do not change `pdd/fix_error_loop.py` behavior in this issue. This issue is only about making the prompt contract valid so sync can proceed.

## Reproduction

From a branch where `pdd/prompts/fix_error_loop_python.prompt` is changed relative to `main`, run:

```bash
pdd sync fix_error_loop
```

Expected current result:

```text
Dry-run validation failed:
fix_error_loop: prompt contract preflight failed:
... missing cloud_fix_errors, fix_error_loop ...
```

A deterministic local check is:

```bash
python - <<'PY'
from pathlib import Path
from pdd.agentic_sync import _prompt_contract_errors_for_module
root = Path.cwd()
print('\n'.join(_prompt_contract_errors_for_module('fix_error_loop', root, root)) or 'NO_ERRORS')
PY
```

## Acceptance Criteria

- `pdd sync fix_error_loop` no longer fails prompt-contract preflight for `fix_error_loop_python.prompt`.
- `_prompt_contract_errors_for_module("fix_error_loop", root, root)` returns no errors on the changed prompt branch.
- The prompt includes source context for both existing public symbols: `cloud_fix_errors` and `fix_error_loop`.
- No behavioral change is made to the progress-sensitive circuit breaker in this issue.
- Existing prompt-contract validation tests still pass.
- #1230 can proceed past dry-run validation and reach the generation/execution stage.

## PDD Command Hint

This is a prompt-contract infrastructure fix. Use:

```text
change → sync
```

## Split Contract

Command sequence: change → sync

Allowed write set:

- `pdd/prompts/fix_error_loop_python.prompt`
- `architecture.json` only if metadata sync updates the existing `fix_error_loop` entry

Acceptance criteria:

- Strict prompt-contract preflight passes for `fix_error_loop`
- Existing public symbols are resolved through self-included source context
- Sync reaches generation/execution instead of stopping at dry-run validation
- No unrelated code, test, or behavior changes

Independently mergeable: True

Scope rule: Do not implement the progress-sensitive circuit breaker from #1230 here. This issue only removes the preflight blocker that prevents #1230 from syncing.

Parent: #1230
Related parent: #1203

Serhan-Asad:
## Summary
- Fixes the #1237 dry-run blocker in the prompt-contract validator instead of adding `pdd/fix_error_loop.py` to `fix_error_loop_python.prompt`.
- Prompts with no output self-include remain valid, matching existing repo convention.
- Partial output self-includes still must cover every declared existing public symbol.

## Root cause
The observed error in #1237 was real, but the issue's requested prompt change was the wrong direction. Most prompt-local interface prompts in this repo do not self-include their generated Python module. The strict changed-prompt path was overreaching by treating no self-include as a hard failure.

The validator should only enforce completeness when a prompt already includes a selected slice of its own output module. That catches the dangerous case without forcing generated module source into prompts that never used that pattern.

## Changes
- `validate_prompt_contract_context(...)` now returns cleanly when there is no output self-include, even in strict mode.
- Existing partial self-include validation remains intact.
- `fix_error_loop_python.prompt` no longer includes `pdd/fix_error_loop.py`.
- Prompt specs for `agentic_sync` and `architecture_include_validation` were updated to match the corrected contract.
- Regression/e2e tests cover the #1237 sync path and the partial-self-include failure case.

## Validation
- Prompt-contract preflight for `fix_error_loop`: `NO_ERRORS`
- `collect_architecture_include_validation_warnings(Path.cwd())`: `0`
- Strict validator scan across repo prompts: `0` failures
- `pytest tests/test_fix_error_loop_prompt_contract.py tests/test_fix_error_loop_sync_e2e.py tests/test_architecture_include_validation.py tests/test_prompt_contract_validation.py tests/test_agentic_sync.py` - 213 passed
- `git diff --check` - passed

Closes https://github.com/promptdriven/pdd/issues/1237


Serhan-Asad:
Ping: pushed final cleanup in d2c1b114d (removed the brittle anti-self-include assertion while keeping the contract/e2e coverage). Local targeted tests pass, and all GitHub checks are green.

Serhan-Asad:
Clarifying the tradeoff here: the original strict rule was trying to prevent a real failure mode where a prompt declares existing public symbols, the LLM only sees part of the module, and generation accidentally drops exports it did not see.

But requiring every changed prompt-local interface to self-include its generated .py file is too broad for this repo. Many first-party prompts declare their interface without self-including their own output module, so that rule turns normal prompt edits into dry-run blockers.

The better split is:
- Preflight should reject present-but-partial self-includes that omit declared existing symbols. If a prompt chooses to include a slice of its own output, that slice must cover the declared existing public surface.
- Prompts with no self-include should remain valid, matching repo convention.
- Deleted exports should be caught by the post-generation public-surface / architecture conformance checks, which are the right layer for validating generated code behavior.

So this PR keeps the useful preflight protection without forcing uncommon self-includes everywhere.

PR #1257, opened 2026-05-27
fix(checkup): #1433 — close 4 of 13 acceptance criteria + 5 codex review findings (partial)

Serhan-Asad:
## Summary

Addresses gltanaka/pdd#1433 (4 review-loop reliability gaps that prevent producing a mergeable PR from a `pdd checkup --pr --review-loop` run).

### Bug #1 — Pre-flight merge conflict detection
`_detect_pr_base_merge_conflict` runs `git merge-tree --write-tree` against the freshly-fetched base branch before the first round. If the PR already has conflicts GitHub would flag, the loop surfaces a `ReviewFinding` blocker instead of reporting clean. Fail-open: git ≥ 2.38 required; any failure skips the check silently.

### Bug #2 — Deterministic gate: `ruff format --check`
`discover_gates` now emits a separate `ruff-format` gate **independently** of the lint gate. Opt-in via `[tool.ruff.format]` in pyproject.toml OR CI-signal scan across `.pre-commit-config.yaml` / cloudbuild / GitHub Actions workflows. 12 codex review rounds hardened this:
- pass-6: CI signal scan instead of `[tool.ruff.format]`-only
- pass-7: `id: ruff-format` canonical hook id + per-file exclude_paths (not whole-branch suppression)
- pass-8: independent of `[tool.ruff]` block + `Set` import fix
- pass-9: YAML comment stripping (no false positives from `# TODO: enable ruff format`), per-file pre-vs-post signal comparison for touched CI configs (unchanged signals count; PR-introduced signals don't), pre-commit `files:`/`exclude:` scope filtering
- pass-10: pyproject `[tool.ruff.format]` bypasses pre-commit scope filter; all hooks unioned (not just first); PR-controlled scope-narrowing detected via pre-vs-post comparison
- pass-11: workflow signal uses full-file scan (not pre-commit scope); exec-context-aware scan skips YAML step name fields (`name: ruff format migration`); `default_stages` propagated into hook dict comparison
- pass-12: list-item marker stripping before YAML-KV check; `default_stages` merged into per-hook dict for full comparison; architecture.json conflict resolved

### Bug #3 — Fixer-commit-push gap
`_commit_and_push_if_changed` now pushes even when the worktree is clean but `pre_fix_sha != HEAD` — the autoheal already committed locally. Rebase helper uses `pre_fix_sha` as upstream instead of `HEAD~1` to handle multi-commit fixer outputs. Guard baseline pinned to `pre_fix_sha` so prompt-source and architecture-registry guards evaluate against the pre-fixer state.

### Bug #4 — Reviewer companion scope
`_collect_companion_source_of_truth_files` maps every code file in the PR diff to its owning prompt + architecture entry. Skips only modules where BOTH the prompt AND the architecture entry appear in the diff. Truncation at 200 entries with a structured `{__truncated__}` marker and a JSON sidecar artifact for full coverage.

## Test coverage
- 413 unit tests pass (413 passed, 1 skipped) across `tests/test_checkup_gates.py` + `tests/test_checkup_review_loop.py`
- All 4 bugs have regression tests; each codex finding has a targeted test
- Full CI suite: `pytest tests/ -m "not integration and not e2e and not real" --timeout=60`

## Codex review passes
12 rounds of codex code review addressed: Set import, CodeQL clear-text-logging, independent ruff-format gate, CI signal false positives (comments), per-file signal comparison, pre-commit scope filtering, guard baseline pinning, sidecar artifact path, per-entry architecture comparison, pyproject scope bypass, multi-hook union, exec-context-aware workflow scan, default_stages propagation, merge conflict resolution.


gltanaka (changes requested):
This fixes real review-loop gaps from gltanaka/pdd#1433, but I do not think it should merge as-is. Required change: make the new pre-commit based ruff-format signal stage-aware. Right now pdd/checkup_gates.py treats any id: ruff-format text in .pre-commit-config.yaml as an opt-in, and the scope parser drops the merged stages value before filtering. An unchanged hook with stages: [manual] still emits a blocking ruff-format gate for ordinary changed Python files. The same problem applies when default_stages: [manual] is inherited by a hook with no explicit stages. That is a false blocker and breaks the stated goal of only adding ruff format --check when there is a real project CI or default pre-commit signal. Please update the pre-commit signal and scope parsing to ignore manual-only/non-default-stage ruff-format hooks, including inherited default_stages, and add regression coverage for both explicit stages: [manual] and top-level default_stages: [manual]. Workflow and CloudBuild signals should still be able to opt in independently.

Serhan-Asad:
@gltanaka addressed in 9ed8965a9 — the pre-commit ruff-format signal is now stage-aware.

- New `_hook_runs_automatically` / `_pre_commit_has_active_ruff_format`: an `id: ruff-format` hook counts as an opt-in **only when it runs automatically**. A manual-only hook — explicit `stages: [manual]` **or** an inherited top-level `default_stages: [manual]` — no longer emits a gate for ordinary changed Python.
- Applied in **both** paths: the untouched-file signal (`_ruff_format_signaled_by_ci`) and the touched-file path (`_file_has_unchanged_ruff_format_signal`). The scope parser also skips manual-only hooks so their `files:`/`exclude:` no longer narrow/widen the gate (the merged `stages` value is honoured, not dropped).
- Workflow and CloudBuild signals stay independent — either can opt in on its own.
- Regression tests added for all three cases you named: explicit `stages: [manual]`, inherited `default_stages: [manual]`, and workflow-opts-in-despite-a-manual-pre-commit-hook.

Also removed the now-unused `_RUFF_FORMAT_CI_PATTERN` and updated the prompt + `architecture.json` to document the stage-aware behavior. All checks green.

gltanaka (approved):
The requested stage-aware pre-commit ruff-format handling is fixed. I rechecked the previous false-blocker path locally: explicit manual-only hooks and inherited default_stages manual no longer emit ruff-format, while workflow opt-in still does. Focused local suite passes: 420 passed, 1 skipped.

Issue #825, opened 2026-05-05
Tooling: Add pdd gate command to enforce PDD evidence policies in CI


gltanaka:
## Problem

If PDD prompts, stories, tests, and evidence are the source program, CI needs a compiler-like gate that evaluates PDD evidence rather than relying on manual inspection of generated code. Today there is no single command that can enforce policies such as “stories passed,” “generated outputs are fresh,” or “no unchecked critical contract rules.”

This should build on evidence manifests and contract coverage rather than duplicating their logic.

## Proposal

Add:

```bash
pdd gate
pdd gate --policy .pdd/policy.yml
pdd gate --json
```

Example policy:

```yaml
require:
  stories_pass: true
  verify_pass: true
  unit_tests_pass: true
  generated_outputs_fresh: true
  no_unchecked_critical_rules: true

allow:
  waivers: true
  story_only_rules: false
  skipped_verify: false

limits:
  max_cost_usd: 20
  max_nondeterministic_context_items: 0
```

Example failure output:

```text
PDD gate failed

- R6 Privacy is unchecked in prompts/refund_payment_python.prompt
  Add a story, generate a test, or create a waiver.

- src/refund_payment.py changed after last evidence manifest
  Run: pdd --force sync refund_payment --evidence

- story__refund_idempotency.md failed validation
  Run: pdd fix user_stories/story__refund_idempotency.md
```

## Dependencies

This issue depends on or should gracefully integrate with:

- Evidence manifests for latest run/devunit state.
- Contract coverage reporting for rule statuses.
- Story validation from `pdd detect --stories`.
- Waivers once a waiver model exists.

## Acceptance criteria

- Reads evidence manifests from `.pdd/evidence/`.
- Fails when generated output hashes do not match latest evidence.
- Fails when changed prompts have no story validation, if policy requires it.
- Fails when critical `MUST` or `MUST NOT` rules are unchecked and not waived.
- Fails when `--skip-verify` or `--skip-tests` was used against policy.
- Emits human-readable and JSON output.
- Provides suggested commands to fix each failure.
- Does not require contracts for legacy modules unless policy requires them.

## Verification

- Add policy fixture tests for pass/fail cases.
- Add stale-output fixture where code hash differs from latest manifest.
- Add unchecked critical rule fixture.
- Run `pdd gate --json` and validate machine-readable output.

DianaTao:
Diana-June 1

PR #1261, opened 2026-05-27
feat(checkup): add pdd checkup drift for regeneration stability (#831)

### LINKED ISSUE #831: Conformance: Add pdd drift command for regeneration stability checks
## Problem

The prompting guide recommends an optional drift check: regenerate a module without changing the prompt, especially after model upgrades or before major releases, and tighten the prompt if output differs behaviorally or fails checks.

The repo also has `pdd/ci_drift_heal.py`, which detects and heals prompt/example drift in CI. That is useful, but it is not the same as a user-facing conformance command that repeatedly regenerates from the same prompt/context to evaluate behavioral stability.

## Proposal

Add:

```bash
pdd drift refund_payment
pdd drift refund_payment --runs 3
pdd drift refund_payment --model <model>
pdd drift refund_payment --from-evidence .pdd/evidence/devunits/refund_payment.latest.json
pdd drift refund_payment --json
```

The command should:

1. Reconstruct or snapshot prompt context.
2. Regenerate code one or more times.
3. Run tests, stories, verify, and policy checks where configured.
4. Compare output shape and behavior.
5. Report whether the dev unit is stable.

Example output:

```text
PDD Drift Report: refund_payment

Runs: 3
Tests: passed 3/3
Stories: passed 3/3
Verify: passed 3/3
Policy: passed 3/3

Output drift:
- public API: unchanged
- implementation structure: changed
- behavior: unchanged

Status: stable
```

## Relationship to existing work

- Do not duplicate `pdd/ci_drift_heal.py`; this command is about regeneration stability, not auto-healing changed prompt/code pairs.
- Should integrate with evidence manifests and context snapshots when available.

## Acceptance criteria

- Supports one-run and multi-run drift checks.
- Can run from latest evidence manifest.
- Separates harmless implementation drift from behavioral drift.
- Fails if tests, stories, verify, or policy checks fail.
- Emits JSON for CI.
- Has cost controls and clear dry-run/safety behavior.

## Verification

- Add fixtures where regenerated output differs syntactically but tests pass.
- Add fixtures where public API changes and command fails.
- Add fixtures where tests/stories fail in one run.
- Run `pdd drift --json` and validate stable/unstable statuses.

DianaTao:
## Summary

Adds **`pdd checkup drift`**, a user-facing command to check whether a dev unit stays stable when code is regenerated from the same prompt—especially after model upgrades or before releases.

This is **not** `pdd/ci_drift_heal` (CI prompt/example auto-heal) and **not** `pdd contracts drift` (contract/code conformance). It targets **regeneration stability**: repeated `pdd generate` into isolated temp paths, compare each candidate to the **pre-run baseline** public API/hash, and rerun configured behavioral checks per candidate.

Command path is **`pdd checkup drift`** (under checkup, alongside `lint` / `contract` / `coverage`). There is **no** top-level `pdd drift`.

Closes #831 (with documented deviations below).

## Motivation

The prompting guide recommends an optional drift check: regenerate without changing the prompt and tighten the prompt if output differs behaviorally or checks fail. The repo had no dedicated CLI for that workflow.

## What's included

| Area | Change |
|------|--------|
| **CLI** | `pdd checkup drift <devunit>` with `--runs`, `--model`, `--from-evidence`, `--code-file`, `--dry-run`, `--json`, `--max-cost` |
| **Core** | `pdd/drift_main.py` — non-mutating regen, baseline API comparison, per-candidate checks, cost budget, `DriftReport` |
| **Routing** | `pdd/commands/checkup.py` dispatches `target == "drift"` |
| **Evidence** | `pdd/evidence_store.py` — manifest loading / path resolution (new on `main`) |
| **Tests** | `tests/test_drift_main.py` (11 tests): baseline unchanged, isolated paths, API drift, live stories/verify/policy wiring, default cost cap, integration path with only regen mocked |
| **Docs** | `docs/drift.md` |

## Usage

```bash
pdd checkup drift refund_payment
pdd checkup drift refund_payment --runs 3
pdd checkup drift refund_payment --model <model>
pdd checkup drift refund_payment --from-evidence .pdd/evidence/devunits/refund_payment.latest.json
pdd checkup drift refund_payment --dry-run
pdd checkup drift refund_payment --json
pdd checkup drift refund_payment --max-cost 5.0   # optional; default $20 for non-dry-run
```

Example report:

```text
PDD Drift Report: refund_payment

Runs: 3
Tests: passed 3/3
Stories: passed 3/3
Verify: passed 3/3
Policy: passed 3/3
Max cost: $20.0000
Total cost: $1.2345

Output drift:
- public API: unchanged
- implementation structure: changed
- behavior: unchanged

Status: stable
```

### Non-mutating regeneration

- Each trial writes to `temp/candidates/run-N/<rel-path>` — **not** the worktree code file.
- After all runs, drift **byte-compares** the baseline file to the pre-run snapshot and raises if the worktree was mutated.
- Regression tests: `test_drift_does_not_mutate_baseline_code_file`, `test_drift_regenerates_into_isolated_paths`, `test_drift_integration_real_eval_baseline_unchanged`.

### Baseline-aware API comparison

- Captures `baseline_public_api` and hash **before** regeneration.
- `--runs 1` still fails when the regenerated candidate’s public API differs from baseline (not only run-to-run).
- Test: `test_drift_public_api_change_fails_against_baseline`.

### Live behavioral checks (per candidate)

Evidence `validation` keys only **enable** checks; pass/fail comes from **live** reruns:

| Check | When enabled | What runs |
|-------|----------------|-----------|
| **Tests** | Discovered `tests/test_<module>.py` | Pytest on sandboxed **candidate** copy |
| **Stories** | `validation.detect_stories` configured | `run_user_story_tests` on the dev unit **prompt** (prompt↔story scope; not regenerated-code semantics) |
| **Verify** | `validation.verify` configured | `pdd verify` on temp copies of prompt + **candidate** + program |
| **Policy** | Policy file or evidence manifest present | `run_gate_policy` when `gate_main` is installed; **fail closed** if configured but gate unavailable |

### Cost controls

- Default max budget **`DEFAULT_MAX_COST_USD` ($20)** for non-dry-run runs (override with `--max-cost`).
- Stops the regen/eval loop when cumulative parsed cost exceeds the cap.
- Tests: `test_drift_max_cost_stops_regeneration`, `test_drift_applies_default_max_cost_for_non_dry_run`.

### Stable vs unstable

- **Stable:** baseline public API unchanged on all completed runs **and** all enabled behavioral checks pass **and** cost budget not exceeded.
- **Implementation drift** (hash changes) is reported separately and does not alone fail stability.
- Exit `0` / `1` and `--json` for CI (`cost_budget_exceeded`, `policy_check_unavailable`, etc.).

## Behavioral check scope (important — do not over-claim)

| Check | Limitation |
|-------|------------|
| **Stories** | Live rerun, but **prompt-level** consistency — not a direct semantic re-check of regenerated code |
| **Policy** | Requires `pdd checkup gate` in the installed package; projects with evidence/policy but no gate will report **unstable** until #825 lands on `main` |
| **Tests** | Heuristic discovery (`tests/test_<module>.py` only); no tests found ⇒ tests treated as passed |
| **Public API** | Shallow AST top-level `def`/`class` names — not full signature/export contracts |

See `docs/drift.md` for full detail.

## Issue #831 alignment

| Criterion | Status |
|-----------|--------|
| One-run and multi-run checks | Done (`--runs`, regen per run into temp) |
| Run from latest evidence manifest | Done (`--from-evidence` + default `.pdd/evidence/devunits/<devunit>.latest.json`) |
| Separate impl vs behavioral drift | Done |
| Fail on tests / stories / verify / policy when configured | Done (live reruns; stories prompt-scoped — see above) |
| JSON for CI | Done (`--json`, exit 0/1) |
| Cost controls | Done (default `$20`, `--max-cost`) |
| Non-mutating / safety | Done (temp candidates + post-run baseline byte guard) |
| Don't duplicate `ci_drift_heal` | Done |

**Intentional deviations from issue text:**

- Command is **`pdd checkup drift`**, not top-level `pdd drift`.
- Stories validate prompt↔story fit, not regenerated-code behavior alone.
- Policy depends on gate module (#825) when policy/evidence is configured.

## Out of scope (follow-ups)

- Top-level `pdd drift` alias
- Code-scoped story validation per regenerated candidate
- End-to-end test with **real** `pdd generate` writing out-of-tree `--output` (today: temp path + baseline byte guard)
- Full context-snapshot replay API
- Broader pytest discovery / manifest-driven test paths
- Tighter policy gating (e.g. only when `validation` or `policy.yml` explicitly present, not any evidence manifest)

## Test plan

- [x] `pytest -q tests/test_drift_main.py` (14 tests)
- [x] `pytest -q tests/commands/test_drift_cli.py` (5 CLI smoke tests)
- [x] `pdd checkup drift --help` (`test_checkup_drift_help`)
- [x] `pdd checkup drift <devunit> --dry-run --json` (`test_checkup_drift_dry_run_json`)
- [x] `pdd checkup drift <devunit> --runs 2` worktree unchanged (`test_checkup_drift_runs_two_leaves_worktree_unchanged`)
- [x] `pdd checkup drift --from-evidence ...` (`test_checkup_drift_from_evidence_json`)
- [x] No top-level `pdd drift` (`test_no_top_level_pdd_drift_command`)


gltanaka (changes requested):
Thanks for taking this on. I think the feature itself is useful: `ci_drift_heal` is CI auto-heal for prompt/example drift, while #831 is asking for a user-facing regeneration-stability check. I do not think this should merge as-is, though.

1. Do not mutate the real generated source while checking drift. `_regenerate_code()` currently runs `pdd generate ... --output <code_path> --force` against the resolved code file (`pdd/drift_main.py:228`-`pdd/drift_main.py:236`), and `run_drift()` calls that inside the runs loop (`pdd/drift_main.py:300`-`pdd/drift_main.py:307`). That leaves the worktree on the last regenerated candidate, even when the command is only supposed to check stability. Please generate each trial into an isolated temp/sandbox path, or restore the original file on every exit path, and add a regression test that the original code file is unchanged after non-dry-run drift checks.

2. Compare regenerated output against a baseline, not only against other regenerated runs. The current decision logic only compares `apis` collected from the run snapshots (`pdd/drift_main.py:341`-`pdd/drift_main.py:350`). With the default `--runs 1`, a regenerated file can rename/remove the public API and still be reported as `stable` because there is no baseline to compare against. Please snapshot the current artifact before regeneration, compare each generated candidate to that baseline public API, and add the acceptance test from #831 where a public API change fails.

3. The behavioral gates are not actually rerun for regenerated candidates. `detect_stories` and `verify` are read from stale manifest validation (`pdd/drift_main.py:314`-`pdd/drift_main.py:320`), so a newly regenerated candidate can break stories/verify while the report still shows them passing. Either rerun the configured tests/stories/verify/policy per candidate, or narrow the PR/issue closure so it does not claim the #831 acceptance criteria. The same applies to the missing cost controls: #831 explicitly requires cost controls for repeated LLM regeneration, but this PR does not add a max-cost/max-budget guard.

Given those gaps, my recommendation is not to merge this yet. The command shape is a reasonable direction, but it needs to be non-mutating, baseline-aware, and honest/live about behavioral checks before it can close #831.

gltanaka (changes requested):
Thanks for the update. The previous concerns around non-mutating generation, baseline API comparison, and a max-cost knob are much better addressed now. I still do not think this should merge yet because two behavior paths make the command report the wrong result.

1. Candidate pytest runs are still testing the baseline package, not the regenerated candidate. `_run_pytest_for_candidate()` copies the candidate into `sandbox_root` and prepends that path to `PYTHONPATH`, but it still runs pytest with `cwd=project_root` (`pdd/drift_main.py:307`-`pdd/drift_main.py:317`). For normal packages, the current working directory/project package wins before or instead of the sandbox copy, and the sandbox also lacks package files such as `__init__.py`. I reproduced this locally: a broken candidate passes when the baseline passes, and a fixed candidate fails when the baseline fails. That means `behavior_unchanged` can be based on tests for the wrong code. Please add a regression test where baseline and candidate have opposite behavior, and make pytest import the candidate artifact reliably (for example by running from an isolated overlay that contains the package/module layout needed by imports, or otherwise removing the baseline package from import precedence).

2. `--from-evidence` is effectively unstable on current `main` even with no policy configured. `_policy_configured()` returns true for any manifest (`pdd/drift_main.py:339`-`pdd/drift_main.py:346`), while `pdd.gate_main` is not present on current `main`, so `_run_policy_check()` fails closed for every evidence-backed drift run (`pdd/drift_main.py:361`-`pdd/drift_main.py:371`). A manifest with `detect_stories`, `verify`, and `unit_tests` all `not_available` still returns `policy_check_unavailable=True` and `status='unstable'`. Since running from latest evidence is an acceptance criterion for #831, evidence alone should not imply a configured policy gate. Please only enable policy when a policy file or explicit policy/gate configuration exists, and add a test that a stable dev unit loaded from an ordinary evidence manifest can pass on current main.

Given these two issues, my recommendation is still not to merge this PR yet. Once candidate tests actually exercise the candidate and ordinary evidence manifests do not force an unavailable policy gate, this should be much closer to mergeable.

gltanaka (changes requested):
Thanks for the latest update. The ordinary evidence-manifest policy behavior is fixed, and the new tests cover the previous baseline-vs-candidate import-precedence issue. I still do not think this should merge yet because the pytest sandbox now has the opposite import problem: it does not include project-local dependencies.

Required change before merge:

`_build_pytest_overlay()` copies only the candidate module plus the matching test files into the overlay (`pdd/drift_main.py:307`-`pdd/drift_main.py:328`), and `_run_pytest_for_candidate()` then runs pytest with `cwd=overlay_root` and `PYTHONPATH=overlay_root` (`pdd/drift_main.py:355`-`pdd/drift_main.py:365`). That means any candidate that imports another project module, or any test that relies on local `conftest.py` / test helpers, fails even when the regenerated candidate is behaviorally stable. I reproduced this with a stable `pdd/refund_payment.py` candidate that imports `from pdd.helper import adjust`; the drift pytest path returned `False` solely because `pdd.helper` was not present in the overlay.

Please add a regression test for a candidate with a project-local import (and ideally copied test support files), then make the sandbox preserve enough of the project/package tree for tests to exercise the candidate in realistic conditions. A full temp copy/symlink overlay with the candidate replacing the baseline module would be a safer shape than a candidate-only package stub.

Minor follow-up: the command help still says policy is required when `policy/evidence` is configured, but the implementation and docs now correctly say ordinary evidence alone does not enable policy. Please align that wording too.

Recommendation remains: do not merge until the candidate pytest path can run stable candidates with local project imports.

gltanaka (approved):
Re-reviewed the latest head (`e6afe74b1`). The previous blockers are addressed:

- Drift regeneration is non-mutating and baseline-aware.
- Ordinary evidence manifests no longer force an unavailable policy gate.
- Candidate pytest runs now use an overlay that preserves package/source roots, conftest, and helpers; my local reproduction with a candidate importing a sibling project module now passes.
- Added CLI smoke coverage for the manual test plan.

I ran:

- `python -m pytest -q tests/test_drift_main.py tests/commands/test_drift_cli.py` → 21 passed
- `python -m pytest -q tests/commands/test_checkup.py tests/commands/test_checkup_contracts.py tests/commands/test_coverage.py tests/test_prompt_lint.py` → 68 passed
- `python -m pdd checkup drift --help`
- `python -m pdd drift --help` (still correctly unregistered)

Non-blocking cleanup: `docs/drift.md` still has one bullet saying policy runs when `policy/evidence` is configured, while the detailed table below correctly says ordinary evidence alone does not enable policy. Worth aligning, but not a merge blocker.

DianaTao:
human runable verify test are in https://github.com/DianaTao/pdd/tree/feat/issue-831-drift-manual-demo

PR #1268, opened 2026-05-28
fix(ci): skip tests/fixtures in cloud-batch chunking + isolate PDD_QUIET leak

gltanaka:
## Summary

Fixes the two failures the cloud-batch release suite (`make test-pdd-cli-release`) hit on `main` — both in the cloud-batch chunked runner, neither reproducible under a plain `pytest tests/`.

**1. Broken fixture crashed chunk collection.** Chunk file discovery (`balance-chunks.py`, and the `entrypoint.sh` fallbacks) hands pytest **explicit file paths**, which bypass `collect_ignore_glob`. The exclusion only skipped `tests/fixtures/one_session_eval/`, so the intentionally broken-syntax `tests/fixtures/coverage_contracts/fake_tests/test_receipt_failed.py` got packed into a chunk and aborted collection with a `SyntaxError`. Now the entire `tests/fixtures/` subtree is excluded from discovery, matching `tests/fixtures/conftest.py`'s `collect_ignore_glob = ["*"]` intent. `tests/test_739_fixtures.py` (a real top-level test, no `fixtures/` path component) is unaffected.

**2. `PDD_QUIET` leaked across tests.** In-process `CliRunner` tests that pass `--quiet` trigger the CLI group callback's `os.environ["PDD_QUIET"] = "1"` write (`pdd/core/cli.py`), which is never undone. Once cloud-batch packed such a test (`test_e2e_issue_1205_example_output_extension.py`) ahead of `test_preprocess.py` in the same chunk, quiet mode silenced the Rich console warning that 4 `test_unresolved_include_*` tests assert on `capsys.out`. Added an autouse `_isolate_pdd_quiet` fixture that snapshots/restores `PDD_QUIET` per test — same pattern as the existing `_enforce_isolated_home` / `preserve_cwd` isolation fixtures. Fixes the whole class (9 test files do in-process `--quiet` CLI calls).

## Test plan

- [x] `make test-pdd-cli-release` (cloud-batch, 77 tasks): **77 passed, 0 failed** (was 75/77).
- [x] Leaker file + all 4 `test_unresolved_include_*` in one process: 9 passed (failed before the fix).
- [x] `test_preprocess.py` + `test_quiet_flag.py`: 138 passed (quiet-mode tests not regressed).
- [x] `balance-chunks.py` discovery excludes all `tests/fixtures/*`, keeps `test_739_fixtures.py`.
- [x] `bash -n entrypoint.sh` + `py_compile balance-chunks.py` clean.

### DIFF
--- a/ci/cloud-batch/balance-chunks.py
+++ b/ci/cloud-batch/balance-chunks.py
@@ -37,14 +37,16 @@ HEAVY_FILE_THRESHOLD_DEFAULT = 300.0
 
 
 def _should_include_test_file(path: Path) -> bool:
-    """Exclude fixture-only test trees that are not part of the runnable suite."""
-    parts = path.parts
-    if "fixtures" in parts:
-        fixtures_idx = parts.index("fixtures")
-        fixture_subtree = parts[fixtures_idx + 1 :]
-        if fixture_subtree[:1] == ("one_session_eval",):
-            return False
-    return True
+    """Exclude fixture-only test trees that are not part of the runnable suite.
+
+    Everything under a ``fixtures/`` directory is scenario data for higher-level
+    tests (one-session eval, coverage contracts, etc.) and is excluded from
+    pytest collection by ``tests/fixtures/conftest.py``. Such files must never be
+    packed into a chunk: an explicit path on the pytest command line bypasses
+    ``collect_ignore_glob``, and some fixtures are intentionally broken (e.g.
+    syntax-error files used to exercise failed-state handling).
+    """
+    return "fixtures" not in path.parts
 
 
 def discover_test_files(test_dir: str) -> list[str]:
--- a/ci/cloud-batch/entrypoint.sh
+++ b/ci/cloud-batch/entrypoint.sh
@@ -383,7 +383,7 @@ if [ "${TASK_INDEX}" -ge "${PYTEST_START}" ] && [ "${TASK_INDEX}" -le "${PYTEST_
             mapfile -t CHUNK_TESTS < "${ASSIGN_OUTPUT}"
         else
             echo "=== balance-chunks.py failed, falling back to alphabetical split ==="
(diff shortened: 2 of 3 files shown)


gltanaka:
LGTM. I verified all six checks against the local worktree at `/private/tmp/pdd-cloud-batch-fix-review`:

**CHECK 1 — Over-exclusion:** `find tests -name 'test_*.py' | grep fixtures` returns only files under `tests/fixtures/coverage_contracts/fake_tests/` and `tests/fixtures/one_session_eval/` — all intentional non-runnable fixture data. `tests/test_739_fixtures.py` is a top-level file; `fixtures` does not appear in its `path.parts`, so it is correctly kept.

**CHECK 2 — entrypoint.sh consistency:** All three `find` invocations that discover test files include `! -path 'tests/fixtures/*'`. No discovery path was missed.

**CHECK 3 — Dead code / unit tests:** Neither `_should_include_test_file` nor any removed helper appears in `ci/cloud-batch/` or `tests/`. The refactored `discover_test_files` in `balance-chunks.py` now uses the simpler `"fixtures" not in path.parts` guard and is tested by `tests/test_balance_chunks.py`, which passes against the new implementation.

**CHECK 4 — `_isolate_pdd_quiet` correctness:** The fixture captures `os.environ.get("PDD_QUIET")` before `yield` and after the test either calls `os.environ.pop("PDD_QUIET", None)` (if unset pre-test) or restores the saved value. No test uses `monkeypatch.setenv` or `monkeypatch.delenv` on `PDD_QUIET`, so there is no teardown ordering conflict. No test asserts that `PDD_QUIET` persists across tests.

**CHECK 5 — `tests/fixtures/conftest.py`:** The file already carries `collect_ignore_glob` entries covering the broken-syntax `fake_tests/` file. The CI-layer exclusion is belt-and-suspenders and correct.

**CHECK 6 — `balance-chunks.py` exclusion:** `"fixtures" not in path.parts` correctly targets only paths where `fixtures` is a directory component — the top-level `tests/test_739_fixtures.py` has parts `('tests', 'test_739_fixtures.py')` and is not excluded.

PR #1275, opened 2026-05-28
Add opt-in Claude Code interactive mode

Serhan-Asad:
## Summary
- add opt-in `PDD_CLAUDE_CODE_MODE=interactive` for the Anthropic agentic provider
- run interactive Claude Code through a POSIX PTY with a temporary `pdd_reply` MCP tool instead of `claude -p --output-format json`
- keep the default Anthropic `claude -p - --output-format json` path unchanged when the flag is unset
- allow short non-empty zero-cost replies only for the interactive Anthropic mode, while preserving existing false-positive guards for normal `claude -p`
- document the new env var and update the agentic prompt contract

Related: promptdriven/pdd_cloud#1780

## Validation
- `conda run -n pdd python -m pytest -q tests/test_agentic_common.py`
- `conda run -n pdd python -m pytest -q tests/test_agentic_common.py tests/test_agentic_common_issue_813_anthropic_api_key_oauth_shadow.py`
- `git diff --check`
- `python -m py_compile pdd/agentic_common.py`
- live no-edit smoke with `PDD_AGENTIC_PROVIDER=anthropic PDD_CLAUDE_CODE_MODE=interactive CLAUDE_MODEL=haiku`: returned `PDD_INTERACTIVE_OK`
- live edit smoke with the same mode: edited a temp file and returned `PDD_EDIT_OK`

## Notes
- `make test` was not completed because the broad target enters real LLM/e2e pytest work and stalled inside a generated `pdd generate` subprocess.
- `conda run -n pdd pylint pdd/agentic_common.py` still fails on existing module lint debt, plus complexity warnings from this bridge; not used as the gate for this change.


Serhan-Asad:
## Independent verification — full local testing ✅

Verified on an isolated worktree at the PR head (`ac4969cc0`) in a dedicated conda env, against real Claude Code (Max OAuth, claude.ai/firstParty), without touching any other working tree.

### Static + unit
- `git diff --check` clean; `py_compile` OK.
- Confirmed the Issue #813 `ANTHROPIC_API_KEY` strip runs **before** the interactive early-return, so interactive mode correctly inherits an OAuth-only env (this is what keeps it on the subscription channel).
- `tests/test_agentic_common.py`: **367 passed** (incl. the new interactive / MCP-server / trust-prompt / zero-cost-gate tests).
- Broader agentic regression sweep (~900 tests): pass. One failure (`test_agentic_fix.py::TestCwdHandling::test_run_agentic_fix_respects_cwd_parameter`) is **environmental** (`PDD_PATH` unset) and reproduces identically on the merge-base — **not a regression**.
- Full-suite `--collect-only`: **10,101 tests collected, zero import/syntax errors** → no cross-module breakage.

### Live end-to-end (real Claude, `CLAUDE_MODEL=haiku`, brand-new untrusted dirs)
| Scenario | Result | Cost |
|---|---|---|
| no-edit, interactive | returned `PDD_INTERACTIVE_OK` | **$0.00** |
| edit, interactive | returned `PDD_EDIT_OK`, wrote the target file | **$0.00** |
| no-edit, flag **unset** (default `claude -p`) | returned `PDD_DEFAULT_OK` | **$0.097** |

The **$0.00 vs $0.097** split is the key evidence: interactive runs the subscription channel (no usage reporting → `cost=0.0`), the default path runs the cost-reported `-p` API tier. Both no-edit runs were in **fresh, never-trusted directories**, so the workspace-trust auto-confirm path was genuinely exercised end-to-end.

### Trust-prompt handling (the load-bearing live path)
- On the installed CLI (2.1.154), the workspace-trust prompt **is shown even with `--dangerously-skip-permissions`** → the detection code is load-bearing, not dead.
- The real prompt wording contains all seven detector tokens (`quick / safety / check / trust / folder / enter / confirm`), so `_claude_interactive_needs_trust_confirmation` fires correctly and the bridge sends Enter once.

### Error / edge paths
- PTY timeout → graceful failure at the deadline, no hang.
- Malformed reply JSON, `job_id` mismatch, non-POSIX platform → all return clean, actionable failures.

### Zero-cost false-positive gate
`MIN_VALID_OUTPUT_LENGTH = 50`. A 9-char zero-cost interactive reply is logged with `false_positive=False` (audit log) — i.e. genuinely in the demotable range yet correctly preserved — corroborated by the `"OK"` (2-char) unit test. The default `claude -p` false-positive guards are unchanged.

### Confidence boundary (intentionally not overclaimed)
- **Mechanism:** high confidence — bypasses `-p`, PTY+MCP reply works, zero-cost short replies survive, default path is byte-identical when the flag is unset.
- **Scope:** this PR covers `run_agentic_task` only. `pdd_cloud`'s `extensions/github_pdd_app/src/services/llm_caller.py` (`call_claude_agentic*`) still uses `claude -p` and remains exposed — that needs a separate follow-up (SHA bump + staging-only flag + cloud-worker smoke).

**Recommendation: safe to merge.** It's opt-in, leaves the default path unchanged, and the interactive path is fully exercised live.

Serhan-Asad:
## Cost tracking verification ✅

Updated the PR with session-transcript cost estimation. Here's the full verification:

### What was added

- **`--session-id <uuid>`** passed to the claude CLI so the exact JSONL transcript can be located after the PTY exits
- **`_find_claude_interactive_session_file`** polls `~/.claude/projects/**/<session-id>.jsonl` with a 2-second deadline after PTY exit
- **`_estimate_claude_interactive_session_cost`** reads the transcript, **deduplicates by `requestId`** (Claude writes the same usage totals on the thinking row, tool-use row, and final text row for each API call — counting all three would triple-charge), aggregates per-model token counts, applies `_calculate_anthropic_cost` (same CSV-backed formula as `claude -p`)
- Falls back to `0.0` gracefully if the file is absent (sandboxed env, transcripts disabled)
- Prompt spec updated to document `--session-id` requirement, JSONL dedup strategy, and cost-estimation contract

### Unit tests

369 passed (`tests/test_agentic_common.py` + `tests/test_agentic_common_issue_813_anthropic_api_key_oauth_shadow.py`), including two new tests:
- `test_estimate_claude_interactive_session_cost_dedupes_request_ids` — verifies requestId dedup and correct per-model token math
- `test_run_claude_interactive_with_mcp_uses_session_usage_cost` — confirms the estimate propagates through `_run_claude_interactive_with_mcp`

### Live smoke (real Claude, Max OAuth, haiku)

```
PDD_CLAUDE_CODE_MODE=interactive PDD_AGENTIC_PROVIDER=anthropic CLAUDE_MODEL=claude-haiku-4-5
```

| Before this commit | After this commit |
|---|---|
| `cost=$0.000000` | `cost=$0.009642` |

Cost is now read from the session transcript instead of always returning 0.0. The $0.00 vs non-zero split confirms the JSONL parsing and `_calculate_anthropic_cost` are wired correctly end-to-end.

### Previous verification (from earlier comment)

All prior checks still hold: 367 unit tests pass, live PTY+MCP works in fresh untrusted dirs (`$0.00` subscription channel vs `$0.097` for default `claude -p` path confirming channel separation), trust-prompt auto-confirm fires, error paths (timeout, malformed reply, job_id mismatch) all fail gracefully.

PR #1279, opened 2026-05-28
feat(models): add Claude Opus 4.8 to the model catalog

### LINKED ISSUE #1278: Add Claude Opus 4.8 to the model catalog
## Summary

Claude Opus 4.8 was released (2026-05-28). Add it to the model catalog as the top Anthropic model. Opus 4.8 is **adaptive-thinking-only**, the same contract as 4.7: `thinking={"type":"adaptive"}` + the effort parameter; the legacy `thinking={"type":"enabled","budget_tokens":N}` shape returns a 400.

## Changes

- `pdd/data/llm_model.csv` — direct-Anthropic `claude-opus-4-8` row (`reasoning_type=adaptive`, `max_reasoning_tokens=16000`, $5 in / $25 out matching 4.7). Anthropic-direct only — Bedrock / Vertex / Azure / OpenRouter don't expose 4.8 at launch and would create dead routes.
- `pdd/llm_invoke.py` — extend `_model_disallows_temperature()` to opus-4-8.
- `pdd/generate_model_catalog.py` — add opus-4-8 to `_ADAPTIVE_ANTHROPIC_MODELS` (so a catalog regen emits `adaptive`, not `budget`) and seed `STATIC_ELO_FALLBACK` (1575, provisional) so the row survives `ELO_CUTOFF` until live arena data exists.
- `README.md` + `pdd/prompts/llm_invoke_python.prompt` — document the `adaptive` reasoning_type (previously only `none`/`budget`/`effort` were listed).
- `tests/test_generate_model_catalog.py` — focused coverage: 4-8 adaptive inference, 16000 max-reasoning-tokens, and regen survival.

## Notes

- `coding_arena_elo` 1575 is provisional (`[EST]`), just above 4.7's 1565; update with real coding-arena data later.
- Runtime adaptive serialization already exists (`reasoning_type='adaptive'` branch in `llm_invoke.py`), so the row is functional on this base — this is a catalog + generator + docs change, not a runtime change.

Serhan-Asad:
## Summary

Adds Claude Opus 4.8 (released 2026-05-28) to the model catalog as the top Anthropic model. Opus 4.8 is **adaptive-thinking-only**, the same contract as 4.7 (verified against the Anthropic extended-thinking docs): `thinking={"type":"adaptive"}` + effort; the legacy `budget_tokens` shape returns a 400.

## Changes

- `pdd/data/llm_model.csv` — direct-Anthropic `claude-opus-4-8` row (`adaptive`, `16000`, $5 in / $25 out). Anthropic-direct only; other providers don't expose 4.8 at launch.
- `pdd/llm_invoke.py` — `_model_disallows_temperature()` extended to opus-4-8.
- `pdd/generate_model_catalog.py` — durability across `generate_model_catalog.py` regen:
  - litellm doesn't know `claude-opus-4-8` yet, so the litellm-driven build loop would drop it. Seed it via **`_MANDATORY_MODEL_ROWS`** (the existing "PDD default but absent from litellm" mechanism), which dedupes automatically once litellm registers the model.
  - seed `STATIC_ELO_FALLBACK` (1575, provisional) so the resolved ELO clears `ELO_CUTOFF=1300`.
  - add opus-4-8 to `_ADAPTIVE_ANTHROPIC_MODELS` so the litellm-derived row is classified `adaptive` (not `budget`) once litellm ships it.
- `README.md` + `pdd/prompts/llm_invoke_python.prompt` — document the `adaptive` reasoning_type (previously only `none`/`budget`/`effort`).
- `tests/test_generate_model_catalog.py` — adaptive inference, 16000, static-ELO cutoff, litellm-unaware seeding, and dedup coverage.

## Why these specific changes

The runtime `reasoning_type='adaptive'` branch already exists, so the catalog row functions immediately. The generator changes are the durability piece: without `_MANDATORY_MODEL_ROWS`, a regen today (litellm unaware of 4.8) would drop the hand-added row entirely; without the `_ADAPTIVE_ANTHROPIC_MODELS` + static-ELO entries, a later litellm-driven row would be misclassified `budget` (→ 400 on 4.8) or fall below the ELO cutoff.

## Notes

- `coding_arena_elo` 1575 is provisional (`[EST]`), just above 4.7's 1565; update with real arena data later.
- Pricing $5/$25 = same as 4.7 (verified against Anthropic's published rate).

## Testing

- `pytest tests/test_generate_model_catalog.py` → **31 passed**.
- `pytest tests/test_llm_invoke.py -k "adaptive or temperature or disallows"` → **23 passed**.
- **End-to-end regen verified**: `build_rows(refresh_elo=False)` with litellm confirmed unaware of 4.8 emits `Anthropic,claude-opus-4-8,...,adaptive` at ELO 1575 via the mandatory-row seed.

Closes #1278


Serhan-Asad:
Thanks — all three failure modes were valid and are addressed in the follow-up commit. Reproduced #1 directly before fixing.

**1. Runtime serialized 4.8 as legacy `thinking.type="enabled"` (not callable) — FIXED.**
Root cause: the litellm opus-adaptive monkeypatch in `llm_invoke.py` gates on `_OPUS_ADDITIONAL_ALIASES`, which listed opus-4-7/4-5 but not 4-8. Reproduced your finding:
```
claude-opus-4-7 -> thinking: {'type': 'adaptive'}              output_config: {'effort': 'medium'}
claude-opus-4-8 -> thinking: {'type': 'enabled', 'budget_tokens': 2048}  output_config: None
```
Added the opus-4-8 hyphen/dot aliases. After the fix both 4-7 and 4-8 return `{'type': 'adaptive'}` + `output_config={'effort': ...}`. Added a `map_openai_params` regression test for 4.8 (hyphen/dot/vertex/bedrock naming) mirroring the 4.7 suite — that's the net that would have caught this.

**3. Generator prompt out of sync (budget-only) — FIXED.**
Updated `generate_model_catalog_python.prompt` so `_infer_reasoning_type` documents the direct-Anthropic adaptive allowlist (opus-4-7/4-8 → `adaptive`) and `_infer_max_reasoning_tokens` documents the 16000 rule, so a prompt-driven regen can't revert to budget-only.

**Nits — FIXED.** Added `adaptive` to README:787, the models API schema (`prompts.py`), and the frontend contract (`api.ts`).

**2. Frontend strength slider (`ALL_MODELS`) stale — PARTIALLY addressed, with a note.**
Added `anthropic/claude-opus-4-8` (elo 1575) so the slider can reach 4.8 as the strongest model. But that hardcoded list is *broadly* stale — it currently tops out at gemini-3-pro (1487) and omits the entire current Anthropic top tier (no opus-4-7, sonnet-4-6, etc.). A full resync of `ALL_MODELS` to the catalog is its own change and out of scope here; flagging it for a dedicated follow-up rather than half-syncing it in this PR.

Tests after the fix: `test_llm_invoke.py` opus/adaptive/map suite **29 passed**, `test_generate_model_catalog.py` **31 passed**.

Serhan-Asad:
## Verification & testing summary

Tested in an isolated worktree (PR head merged with current `main`, fresh venv, `litellm==1.82.6` — the repo's upper pin, confirmed **unaware** of `claude-opus-4-8`, so the seed/registration paths are exercised, not dedup).

### Bug found and fixed in this PR
The catalog/kwargs/transform changes were all correct, but `claude-opus-4-8` was **not actually callable**: `_register_csv_models_with_litellm()` derived `litellm_provider` from a `/` split and defaulted **bare** ids to `"openai"`. Since litellm doesn't yet ship 4.8, it got registered as an OpenAI model, so PDD's runtime resolved the wrong provider, failed auth against the Anthropic key, and **silently fell back** to another model.

Evidence (PDD's real path, dummy key):
- `claude-opus-4-8` → `LiteLLM completion() ... provider = openai` → `[AUTH ERROR] ... Trying next model`
- `claude-opus-4-7` (control, known to litellm) → `provider = anthropic` ✓

Fix (commit `97eb81e2`): derive the provider from the id prefix when present, otherwise from the CSV `provider` column (`Anthropic → anthropic`), keeping the `"openai"` fallback for unmapped providers so existing registrations are unchanged. Added a line to the generating prompt so `pdd sync` won't regenerate the old default, plus regression tests.

### What was verified
| Check | Result |
|---|---|
| `tests/test_llm_invoke.py` (full, +5 new regression tests) | **322 passed** |
| `tests/test_generate_model_catalog.py` (full) | **31 passed** |
| Catalog regen durability (litellm-unaware seed path) | 1 `Anthropic claude-opus-4-8` row, ELO 1575, `adaptive`, no dup, clears cutoff |
| Strength routing | 4.8 (1575) is unique global max; `strength=1` flips 4.7→4.8; mid-strengths unchanged (no downgrade) |
| Adaptive kwargs PDD builds for 4.8 | `thinking={type:adaptive,display:summarized}` + effort, **temperature omitted** — byte-identical to 4.7 |
| litellm `map_openai_params` transform for 4.8 | emits adaptive shape (no legacy `enabled` 400-shape) |
| Provider resolution after fix (PDD real path) | `claude-opus-4-8 → provider=anthropic` |
| **Live Anthropic call** through `llm_invoke` | `claude-opus-4-8` → **200**, `pong`, $0.00019; `claude-opus-4-7` control → 200 ✓ |
| Pricing | $5 in / $25 out matches Anthropic's published Opus 4.8 rate |

The live call is the key piece CI can't cover (all LLM calls are mocked), and it’s what surfaced the registration bug — now confirmed working end-to-end after the fix.

gltanaka (changes requested):
I think this PR is needed: Anthropic's current docs list `claude-opus-4-8` as available today, priced at $5/$25 per MTok for regular usage, and adaptive-thinking-only for Opus 4.8/4.7. I also confirmed the current LiteLLM in this repo does not know `claude-opus-4-8`, so the mandatory seed + CSV-provider registration path is necessary.

I would not merge yet. Required changes:

1. `pdd/frontend/lib/modelResolver.ts:10` hardcodes Opus 4.8 with `context_limit: 200000`, but Anthropic's model overview lists Claude Opus 4.8 with a 1M-token context window. `PromptMetricsBar` prioritizes `selectedModel.context_limit` over backend metrics, so the newly strongest slider model will under-report context/output capacity and can show false warnings. Please change this row to the documented 1M context limit, or stop hardcoding this value and derive it from a backend/catalog source.

2. Please reconcile the provider availability assumption. The PR body/comments say direct-Anthropic only because other providers do not expose 4.8 at launch, but Anthropic's current model overview says current Claude models are available through AWS Bedrock and Vertex AI, and lists `anthropic.claude-opus-4-8` for Bedrock plus `claude-opus-4-8` for Vertex AI. Either add verified Bedrock/Vertex rows with correct provider registration and reasoning metadata, or update the PR/comments to explicitly say those routes are intentionally omitted pending validation. The catalog update should not merge with a stale launch assumption.

Local verification I ran: `python -m pytest tests/test_generate_model_catalog.py tests/test_llm_invoke.py -k "opus_48 or provider_for_bare_anthropic_model_maps_to_anthropic or registers_bare_anthropic_model_under_anthropic_provider or set_model_rate_map_populates_provider_map or provider_for_prefixed_model_uses_prefix or provider_for_bare_openai_and_unknown_fallback"` -> 12 passed.

Serhan-Asad:
## 4.8 — gap fully closed (all live calls on `claude-opus-4-8`)

Live Anthropic calls through `llm_invoke` across every parameter path:

| Path | Result |
|---|---|
| Text, effort=low (original) | ✅ `pong`, $0.00019 |
| Text, effort=medium | ✅ `ok`, $0.000175 |
| Text, effort=high | ✅ `ok`, $0.000175 |
| Structured output (pydantic) | ✅ `answer=4` (real parsed object), $0.0033 |
| Batch mode | ✅ `['ok']`, $0.000175 |
| 4-7 control (every test) | ✅ matches |

`pdd generate --local` (full CLI) could not be driven to completion in the headless test env — it stalls on a GitHub device-flow auth prompt, and `claude-opus-4-7` stalls at the identical point, so it's a model-agnostic environment artifact, not a 4.8 issue. The LLM call `generate` makes is the same `llm_invoke` path proven live above.

Serhan-Asad:
Both required changes addressed in the latest commit, plus a callable-on-relays fix surfaced while doing it.

## Change 1 — frontend context window ✅
`modelResolver.ts` Opus 4.8 row `context_limit: 200000 → 1000000` (1M, per Anthropic's overview). It's the strongest slider model, so this removes the ~5× under-report / false output-capacity warning from `PromptMetricsBar`. Other rows stay at 200000 (correct for those models; the broad `ALL_MODELS` resync remains a separate follow-up).

## Change 2 — provider availability reconciled ✅
The "other providers don't expose 4.8 at launch" framing was stale — corrected:
- **Added Bedrock** (`anthropic.claude-opus-4-8`) and **Vertex** (`vertex_ai/claude-opus-4-8`) rows at ELO 1575, `reasoning_type=effort`, mirroring the existing opus-4-7 relay rows, in both `llm_model.csv` and `_MANDATORY_MODEL_ROWS` (the latter so a catalog regen keeps them while litellm is unaware of the 4-8 ids — verified a regen emits exactly 3 opus-4-8 rows, no dups).
- **Azure / Microsoft Foundry + third-party aggregators (Perplexity/OpenRouter/GMI): intentionally deferred pending validation**, documented in a seed comment and called out here for your sign-off. Rationale: Azure routes through `AzureAIStudioConfig` (OpenAI-based), which the Bedrock/Vertex adaptive relay patches do **not** reach; the existing `azure_ai/claude-opus-4-7` row still rides the legacy budget shape, which is unaudited for the adaptive-only 4.7+/4.8 contract (see the note at `llm_invoke.py:92-94`). I didn't blind-add an `azure_ai/claude-opus-4-8` budget row I can't validate against the live relay. Happy to add it if you'd prefer it land here.

## Callable-on-relays fix (found while adding the rows)
Adding the relay rows surfaced that the Bedrock Converse and Vertex transform patches gated on **per-adapter, 4-7-only** alias tuples — so the `effort → adaptive` conversion was being **dropped for 4-8** on those relays (it would have shipped the wrong thinking shape). Fixed at the root: a single `_RELAY_OPUS_ADAPTIVE_ALIASES` (4-7 **and** 4-8, deliberately **not** pre-adaptive 4-5, which ships `reasoning_type=none` on its relay rows) now feeds both relay sites so they can't half-sync again. Also mapped `AWS Bedrock → bedrock` in the CSV-provider registration so the bare-dot Bedrock id resolves to `bedrock` instead of the `openai` default.

## Tests
Real `AmazonConverseConfig` + `VertexAIAnthropicConfig` transform regression tests for 4-8 (red before the alias fix, green after), bedrock dotted-id provider registration, relay seed durability/dedup, and reasoning_type inference. `test_llm_invoke.py` + `test_generate_model_catalog.py` → **360 passed**.

## Frontend null-safety
This change makes `TokenMetrics.context_usage_percent` legitimately nullable (backend returns null for litellm-unknown models). Widened the type to `number | null` and guarded all 5 consumers (`PromptMetricsBar` ×4, `PromptSpace` ×1) with `?? 0` to prevent a `null.toFixed()` whitescreen. Note: I could not run vitest/tsc in this worktree (no `node_modules`), so these TS edits are unrun — they're standard nullish-coalescing + a type widening, consistent across all consumers.

gltanaka (changes requested):
The two previous review items are addressed: the frontend Opus 4.8 row now uses the documented 1M context window, and Bedrock/Vertex rows now exist with relay-specific adaptive transform coverage.

I still would not merge yet. There is a regen blocker around the Azure deferral:

`pdd/generate_model_catalog.py` documents Azure AI / Microsoft Foundry Opus 4.8 as intentionally deferred pending validation, and the committed `pdd/data/llm_model.csv` correctly does not add `azure_ai/claude-opus-4-8`. But the generator does not enforce that deferral. In the updated PR worktree, with a supported LiteLLM install (`litellm==1.80.11` in my env), `litellm.model_cost` already contains `azure_ai/claude-opus-4-8`, so `build_rows(refresh_elo=False)` emits:

```text
Azure AI azure_ai/claude-opus-4-8 1575 budget AZURE_AI_API_KEY
```

That means the next catalog regen can add the exact route this PR says is unaudited, with `reasoning_type=budget` / `max_reasoning_tokens=128000`. This is the same risk the deferral comment is trying to avoid.

Required change: either explicitly skip `azure_ai/claude-opus-4-8` in the generator until the Azure adapter is validated, with a regression test that a fake LiteLLM `azure_ai/claude-opus-4-8` entry is omitted, or validate/fix the Azure path and intentionally add the row with tests. As written, the committed CSV and the generator policy diverge.

Local verification:
- `python -m pytest tests/test_generate_model_catalog.py tests/test_llm_invoke.py -k "opus_48 or bare_bedrock or registers_bare_bedrock or provider_for_bare_bedrock or provider_for_bare_anthropic or register_csv_models or set_model_rate_map_populates_provider_map or map_openai_params_direct_opus_48 or bedrock_converse_adapter_opus_48 or vertex_anthropic_transform_request_keeps_output_config_for_opus_48"` -> 16 passed.
- `build_rows(refresh_elo=False)` -> 4 Opus 4.8 rows, including the unintended Azure AI row above.

Serhan-Asad:
Good catch — the deferral was documented but not enforced, so the generator and the committed CSV could diverge on a LiteLLM build that already ships the Azure id. Fixed by making the generator enforce it.

**Change:** added a `_DEFERRED_MODEL_IDS` guard in `generate_model_catalog.py`, checked first in the `build_rows()` loop, so deferred routes are omitted regardless of whether `litellm.model_cost` knows them. `azure_ai/claude-opus-4-8` is the first entry, with the rationale inline (legacy budget shape via `AzureAIStudioConfig`, unaudited for the adaptive-only contract). The comment says to remove an entry once its path is validated and the row is added with tests.

**Reproduced your scenario** — injecting a fake `azure_ai/claude-opus-4-8` into `litellm.model_cost` and running `build_rows(refresh_elo=False)` now yields exactly 3 Opus 4.8 rows, Azure omitted:
```
AWS Bedrock       anthropic.claude-opus-4-8
Anthropic         claude-opus-4-8
Google Vertex AI  vertex_ai/claude-opus-4-8
```
(was 4 with the unintended Azure row before this commit.)

**Regression test** (`test_azure_opus_48_is_deferred_even_when_litellm_knows_it`): injects a fake LiteLLM `azure_ai/claude-opus-4-8` entry and asserts `build_rows` omits it while keeping the direct-Anthropic 4.8 row. `test_generate_model_catalog.py` → 35 passed.

The Azure deferral itself stands (I can't validate the budget shape against a live Azure relay here); the guard makes that a real, enforced policy rather than a comment. Happy to flip it — drop the guard entry and add the `azure_ai/claude-opus-4-8` row with tests — if/when the Azure adaptive path is validated.

gltanaka (approved):
The last blocker is fixed. I verified that the generator now enforces the Azure 4.8 deferral instead of just documenting it: `build_rows(refresh_elo=False)` emits exactly the direct Anthropic, AWS Bedrock, and Google Vertex AI Opus 4.8 rows, and does not emit `azure_ai/claude-opus-4-8` even though this local LiteLLM registry knows that id.

Local verification on the latest head:
- `python -m pytest tests/test_generate_model_catalog.py -k "azure_opus_48 or opus_48"` -> 10 passed
- `python -m pytest tests/test_llm_invoke.py -k "opus_48 or bare_bedrock or provider_for_bare_bedrock or provider_for_bare_anthropic or register_csv_models or set_model_rate_map_populates_provider_map or map_openai_params_direct_opus_48 or bedrock_converse_adapter_opus_48 or vertex_anthropic_transform_request_keeps_output_config_for_opus_48"` -> 7 passed
- `npm ci && npm run build && npm test -- --run` in `pdd/frontend` -> build passed, 49 frontend tests passed
- GitHub checks are green.

Approved from my side.

Issue #31, opened 2025-09-05
better to not ask interactively for unset env var


rudi-cilibrasi:
I just upgraded to pdd v0.51.0. I copied the llm_models.csv from the package distribution into my ~/.pdd directory. I tried to run generate and found it was hard to use because it is asking me for many different API keys I don't have interactively forcing me to hit return many times. I think it would be better if the pdd utility simply printed a warning without prompting the user to enter the key interactively. Then I wouldn't need to modify the distributed llm_models.csv file each time a new version is released.

jamesdlevine:
I agree, would prefer to suppress the behavior by default, and enable it only with a variable setting or commandline option.   I'd prefer default suppressed behavior to be, simply skip the model if no key is available, and error out completely if no appropriate models can be selected.

gltanaka:
Closing as duplicate of #140 which tracks the same issue (interactive API key prompts blocking automated and manual workflows). Details and context from this issue have been consolidated in #140.

Issue #27, opened 2025-08-27
pdd runs slowly on windows


rudi-cilibrasi:
the first time pdd is run under the cmd shell took about 45 seconds to finish for me

all subsequent times i run pdd it takes about 5 seconds to print the help message

windows 11 home
64 GB RAM
SSD HD

PR #1283, opened 2026-05-29
feat: teach pdd test to generate tests from contract rule IDs

### LINKED ISSUE #821: Context: Teach pdd test to generate tests from contract rule IDs
## Problem

`context/test.prompt` is automatically included during `pdd test`, and tests are permanent assets that accumulate while generated code can be regenerated. The current file has strong project-wide testing guidance, especially around external-boundary mocking and test isolation, but it does not tell test generation how to use named natural-language contract rules.

If prompts contain `<contract_rules>` with stable rule IDs, generated tests should use those IDs as an explicit test planning surface.

## Proposal

Add contract-aware testing guidance to `context/test.prompt`:

```text
When the module prompt contains <contract_rules>:

1. Read every rule ID.
2. Generate at least one behavioral test for each MUST rule.
3. Generate at least one negative test for each MUST NOT rule.
4. Name tests with the rule ID when practical.
   Example: test_R2_rejects_over_refund
5. Prefer observable outcomes over private implementation details.
6. Do not assert private helper names or internal class structure.
7. If a rule cannot be tested with available fixtures, emit a TODO explaining the missing fixture.
8. Preserve existing accumulated tests.
```

Keep the existing external-boundary mocking and test isolation rules intact.

## Codebase pointers

- Test context file: `context/test.prompt`
- Test generation command: `pdd/cmd_test_main.py`
- Agentic test flow: `pdd/agentic_test.py`, `pdd/prompts/agentic_test_*`
- Existing tests around test generation: `tests/test_cmd_test_main.py`

## Acceptance criteria

- `context/test.prompt` explicitly instructs test generation to inspect `<contract_rules>`.
- Generated tests reference contract IDs in names or comments where practical.
- `MUST NOT` rules produce negative tests when fixtures allow it.
- If a rule cannot be tested, generated tests include a clear TODO or skipped-test reason rather than silently ignoring it.
- Existing external service mocking, test pollution, and merge guidance remains present.

## Verification

- Add a fixture prompt with `<contract_rules>` containing at least one `MUST` and one `MUST NOT` rule.
- Run `pdd test` against the fixture and confirm generated test names/comments reference rule IDs.
- Confirm existing test-generation tests still pass.
- Confirm `--merge` behavior preserves accumulated tests.

DianaTao:
## Summary

- Adds contract-aware test planning guidance to `context/test.prompt` so `pdd test` treats stable rule IDs in `contract_rules` as the test-planning surface (MUST → behavioral tests, MUST NOT → negative tests, rule-linked TODO/skip when fixtures are missing).
- Aligns legacy (`generate_test_LLM`, `generate_test_from_example_LLM`) and agentic (`agentic_test_generate_LLM`) templates so all generation paths apply the same rules.
- Documents manual-mode behavior in `README.md` and adds regression tests for context content, prompt propagation, and `--merge` preservation.

Closes #821

## Checkup (#37) — out of scope here

Automated checkup ([DianaTao/pdd#37](https://github.com/DianaTao/pdd/issues/37) → [DianaTao/pdd#40](https://github.com/DianaTao/pdd/pull/40)) ran on this branch and reported **repo-wide** hygiene (deps, imports, frontend modals). Those changes are **not** included in this PR; they should land separately on `main` (not stacked on this feature branch).

## Gaps closed vs DianaTao/pdd#31

- **LLM templates**: PR #31 only updated `agentic_test_generate_LLM`; this PR also wires `generate_test_LLM` and `generate_test_from_example_LLM` to apply `context/test.prompt` contract planning explicitly.
- **README**: Adds user-facing manual-mode note (suggested in issue #27 workflow, missing from #31).
- **Scope**: Omits unrelated `architecture.json` filepath churn from #31 (stale agentic-bug entries, etc.) to keep the diff reviewable.
- **Fixtures**: Uses `tests/fixtures/test_generation/` for regression (same approach as #31’s second commit) instead of an extra `context/contract_rules_test_fixture*` module that sync would not regenerate.

## Test plan

- [x] `PYTHONPATH=. pytest -q tests/test_cmd_test_main.py` (55 passed, 1 skipped)
- [x] `PYTHONPATH=. pytest -q tests/commands/test_contract_rule_test_smoke.py` — CLI smoke for `pdd test --manual --merge` on a `contract_rules` fixture (rule IDs in prompt + merged test file)


gltanaka (changes requested):

This is needed in principle: #821 is still valid and main does not have contract-rule test-planning guidance. But this PR does not yet deliver the core legacy `pdd test` path because `generate_test_LLM.prompt`'s `context/test.prompt` include is not expanded in the final LLM prompt.

Repro from this PR branch with the worktree pinned as `PDD_PATH`:

```bash
PDD_PATH="$PWD/pdd" PDD_QUIET=1 python - <<'PY'
from pdd.load_prompt_template import load_prompt_template
from pdd.preprocess import preprocess

s = preprocess(
    load_prompt_template("generate_test_LLM"),
    recursive=False,
    double_curly_brackets=False,
)
print("CONTRACT RULE TEST PLANNING" in s)
print("<include optional>context/test.prompt</include>" in s)
PY
```

Expected: the contract planning section is present and the include tag is gone. Actual on this branch: `False` then `True`.

The apparent cause is that the literal inline text ``<include>s module ...`` earlier in `generate_test_LLM.prompt` is matched as an include opening and spans through the real `</include>` for `context/test.prompt`; because that false match intersects an inline-code span, the preprocessor leaves the whole match untouched. The new tests only assert prompt source text or patch `generate_test`, so they do not catch the final preprocessed prompt that the LLM receives.

Please fix the include expansion for `generate_test_LLM.prompt` (for example by escaping/removing the literal false `<include>` opener or by hardening the preprocessor), then add a regression that runs the actual template through `preprocess(load_prompt_template("generate_test_LLM"), ...)` with `PDD_PATH` pinned to the test repo and asserts the final prompt contains `CONTRACT RULE TEST PLANNING` / the `MUST NOT` negative-test guidance. After that, this should be mergeable.

pdd-cli (approved):
The fixes for the unexpanded `<include>s module` tag in `generate_test_LLM.prompt` and the new regression tests look great. CI is green.

pdd-cli:
Hey @DianaTao! I took a look. It looks like you perfectly addressed `gltanaka`'s feedback by rewording the `<include>s module` literal in the prompt and adding the regression test (`test_generate_test_llm_preprocess.py`) to verify it expands correctly. CI is totally green across the board ✅. Looks ready to me!

gltanaka (approved):
The requested fix is in place. I verified that both legacy test-generation templates now expand context/test.prompt in the final preprocessed prompt, including the contract-rule and MUST NOT guidance, and the include tag no longer survives preprocessing. Targeted pytest and pylint pass locally.

DianaTao:
manual test in https://github.com/DianaTao/pdd/tree/docs/issue-821-manual-test-plan

Issue #1254, opened 2026-05-27
pdd sync fails in --force mode when ANTHROPIC_API_KEY is unavailable despite alternate model labels


sohni-tagirisa:
Summary

pdd sync fails hard when ANTHROPIC_API_KEY is unavailable, even after switching model-selection labels away from Anthropic-based models.

In issue #996, the workflow was reconfigured from:

pdd-sonnet
→ pdd-opus
→ pdd-gemini-flash

However, the sync execution continued attempting to invoke Anthropic providers and exited with:

ERROR - API key 'ANTHROPIC_API_KEY' not set. In --force mode, skipping interactive prompt.

This suggests provider/model resolution is either:

cached incorrectly,
falling back to Anthropic unexpectedly,
or ignoring the active model-selection labels during sync execution.
Observed Behavior

In issue #996:

pdd change initially failed due to provider rate limits
pdd sync then failed repeatedly with missing ANTHROPIC_API_KEY
changing labels from:
pdd-sonnet
to pdd-opus
to pdd-gemini-flash
did not prevent Anthropic invocation attempts

Relevant failure output:

ERROR - API key 'ANTHROPIC_API_KEY' not set. In --force mode, skipping interactive prompt.

Despite:

Anthropic labels being removed
Gemini Flash label being added
workflow relaunch occurring afterward
Expected Behavior

When:

Anthropic-specific labels are removed
another provider label is active (pdd-gemini-flash)
and sync jobs are requeued

the orchestrator should:

resolve to the active provider/model correctly
avoid Anthropic API validation entirely
fail only if the selected provider credentials are unavailable
Potential Root Cause Areas
Provider selection cached between workflow retries
Sync workflow defaulting to Anthropic regardless of active labels
Label parsing occurring only at initial queue time
Model override not propagating into downstream sync phases
Fallback-provider ordering ignoring explicit label selection
Reproduction Steps
Start issue with pdd-sonnet
Trigger workflow
Remove Anthropic label
Add pdd-gemini-flash
Requeue pdd-sync
Observe sync still attempts Anthropic authentication
Notes

This issue blocks recovery workflows where users attempt to switch providers after rate limits or credential failures occur mid-execution.

The logs indicate the workflow acknowledges the new model label in GitHub comments, but runtime provider selection still attempts Anthropic authentication internally.

Issue #1135, opened 2026-05-21
pdd sync Opus run fails commands module because inner llm_invoke requires ANTHROPIC_API_KEY


Serhan-Asad:
## Problem

The follow-up `pdd sync` job for #1128 failed after the original change job successfully created PR #1131.

The sync job succeeded for 7 modules, but the `commands` module failed because an inner `llm_invoke` path required `ANTHROPIC_API_KEY`, which was not present in the subprocess environment.

## Evidence

- Issue: #1128
- PR artifact: #1131
- Sync Job ID: `mk7r5m1dhv1k5ceD8yHX`
- Cloud Run execution: `pdd-executor-job-opus-km4d6`
- Final GitHub comment: https://github.com/promptdriven/pdd/issues/1128#issuecomment-4513398715
- Progress comment: https://github.com/promptdriven/pdd/issues/1128#issuecomment-4513319508

Final module result from Cloud Run logs:

```text
Status: Failed
Message: Failed: ['commands']. Succeeded: ['jobs', 'models', 'track_cost', 'cost_budget_watcher', 'budget_settings', 'slash_command_parser', 'budget_comments'].
commands: │ Total time: 232.02s | Total cost: $0.8077 | Overall status: Failed │
ERROR: API key 'ANTHROPIC_API_KEY' not set. In --force mode, skipping interactive prompt.
Cost: $50.0924
Model: anthropic
```

The attribution payload inside the failed module showed:

```text
PDD_AGENTIC_PROVIDER=anthropic
PDD_MODEL_DEFAULT=vertex_ai/gemini-3.5-flash
CLAUDE_CODE_USE_VERTEX=0
GOOGLE_CLOUD_PROJECT=prompt-driven-development
CLOUD_RUN_EXECUTION=pdd-executor-job-opus-km4d6
```

The outer executor selected Opus/OAuth credentials earlier in the job:

```text
Model override: claude-opus-4-7 (family=claude_cli)
Claude CLI model override: claude-opus-4-7
Trying credential: claude-oauth-v9
```

## Expected behavior

An Opus GitHub App sync run should not fail an inner module because direct `ANTHROPIC_API_KEY` is absent if the executor is intentionally using Claude OAuth credentials. The credential/model path used by `pdd sync` module generation should stay compatible with the selected executor credential path, or fail early with a clearer unsupported-credential error before spending work on other modules.

## Actual behavior

[...]

PR #1115, opened 2026-05-20
fix(llm_invoke): lock provider when PDD_MODEL_DEFAULT has routing prefix (#1113)

Serhan-Asad:
## What this PR does

Locks `llm_invoke` model selection to the provider implied by `PDD_MODEL_DEFAULT`'s routing prefix (`vertex_ai/`, `gemini/`, `anthropic/`, `azure_ai/`). Cross-provider fallback is now opt-in via `PDD_CROSS_PROVIDER_FALLBACK`, instead of being the silent default.

Closes #1113.

## Why

Cloud Run sync configured for Vertex (`PDD_AGENTIC_PROVIDER=google` + `PDD_MODEL_DEFAULT=vertex_ai/gemini-3.5-flash` + `GOOGLE_GENAI_USE_VERTEXAI=true`) was failing when the configured model wasn't in the bundled `llm_model.csv`. Selection drifted into higher-ELO Anthropic / Fireworks rows, then died on missing `ANTHROPIC_API_KEY` / `FIREWORKS_API_KEY` in the Google-only env.

`PDD_AGENTIC_PROVIDER` constrains the agentic CLI but does not constrain LiteLLM selection inside `llm_invoke`. This PR treats the routing prefix on `PDD_MODEL_DEFAULT` as the authoritative provider boundary for `llm_invoke`.

## Behavior changes

- Provider-prefixed `PDD_MODEL_DEFAULT` keeps base lookup, alias resolution, surrogate fallback, and strength interpolation inside the implied provider.
- Cross-provider fallback after a transient failure (rate limit, timeout/connection error, 429, 5xx) now requires `PDD_CROSS_PROVIDER_FALLBACK=1|true|yes|on|enabled|transient`.
- Cross-provider fallback skips providers without configured credentials, and can be restricted further with `PDD_MODEL_FALLBACK_PROVIDERS`.
- Legacy/unprefixed CSV rows can still match through the provider alias set; rows with a different known routing prefix are rejected from the lock.

## Files

- `pdd/llm_invoke.py` — provider-lock logic in `_select_model_candidates`, opt-in transient fallback path
- `pdd/prompts/llm_invoke_python.prompt`, `README.md`, `context/llm_invoke_example.py` — contract docs
- `tests/test_llm_invoke.py`, `tests/test_e2e_issue_295_openai_schema.py`, `tests/test_e2e_issue_296_custom_csv.py`, `tests/test_e2e_openai_required_array.py` — provider-lock regressions
[...]

PR #1295, opened 2026-05-29
feat(llm_invoke): Codex/ChatGPT subscription as a first-class provider (#1269)

### LINKED ISSUE #1294: feat(llm_invoke): Codex/ChatGPT subscription fallback when ANTHROPIC_API_KEY is missing or failing (#1269; fixes #1254/#1135)
## Summary

Make a **ChatGPT/Codex subscription** a first-class fallback for PDD's **inner `llm_invoke`** path so that when `ANTHROPIC_API_KEY` is **missing** *or* **failing** (rate-limit / auth error), `pdd sync` (and every `llm_invoke` caller) can transparently fall back to a flat-rate ChatGPT subscription instead of crashing.

This is the feature requested in #1269 ("use codex subscription, reference openclaw") and it fixes the failure class in #1254 / #1135 ("`pdd sync` fails in `--force` mode when `ANTHROPIC_API_KEY` is unavailable" / "inner `llm_invoke` requires `ANTHROPIC_API_KEY`").

## Background — what already exists vs the gap

- **Agentic path (already done):** the coding-agent path already supports Codex via `codex exec` (see `pdd/agentic_common.py` `_has_codex_auth_file`, provider order `openai→codex`). PDD Cloud already stages a `codex-auth` secret and runs a provider waterfall (claude→codex→gemini) for agentic steps.
- **Inner `llm_invoke` path (the gap):** `pdd/llm_invoke.py` calls `litellm.completion()` for non-agentic prompt/classification/structured calls. It has **zero** Codex reachability — no `chatgpt/` CSV row, no codex auth bridge. When `ANTHROPIC_API_KEY` is absent in `--force` mode it logs `API key 'ANTHROPIC_API_KEY' not set. In --force mode, skipping interactive prompt.` and the run fails (confirmed in staging Cloud Run logs, 2026-05-29).

## Mechanism (validated end-to-end on a real ChatGPT subscription)

- PDD's pinned litellm (`1.82.6`) already ships `LlmProviders.CHATGPT` → `https://chatgpt.com/backend-api/codex` and knows `chatgpt/gpt-5.3-codex`, `chatgpt/gpt-5.4`, etc.
- `codex login` (ChatGPT sign-in) caches `~/.codex/auth.json` with tokens nested under `tokens:{access_token, refresh_token, id_token, account_id}`. litellm's authenticator expects these **top-level** in `CHATGPT_TOKEN_DIR` → a small auth bridge reshapes it.
- **litellm bug:** `litellm.completion()` on `chatgpt/*` returns **empty output** on 1.82.6 (upstream BerriAI/litellm#25429; still repro on 1.85.1). The Codex backend streams content via `response.output_item.done` SSE events with an empty `response.completed.output`. Fix = vendored 12-line patch from litellm PR #27562 (aggregate the items). Proven to restore real content.
- **Backend ignores `response_format`/`json_schema`** → must prompt-coerce JSON. Proven: "return ONLY a JSON object …" yields clean parseable JSON.

## Scope of THIS issue/PR — PDD CLI (the engine) only

1. **`pdd/data/llm_model.csv`** — add a Codex subscription row, empty `api_key` (device-flow, like the existing `github_copilot/` rows), `structured_output=False`, `reasoning_type=none` (avoid 400s on the fallback path):
   `OpenAI ChatGPT,chatgpt/gpt-5.3-codex,0.0,0.0,1407,,,0,False,none,`
2. **Vendor litellm PR #27562** — version-gated, fail-safe monkey-patch applied once at `llm_invoke` import (never raises at import; no-op if the class/method is absent).
3. **Codex auth bridge** — reshape `~/.codex/auth.json` (or `CODEX_HOME/auth.json`) nested `tokens` → top-level into a `CHATGPT_TOKEN_DIR`; reuse the existing codex-auth detection. Only when the user hasn't set `CHATGPT_TOKEN_DIR`.
4. **Force-mode credential check** — extend the `github_copilot/` block in `_ensure_api_key` so `chatgpt/` rows are skipped cleanly in `PDD_FORCE` mode when no codex auth is present (prevents litellm hanging on interactive device login in headless runs).
5. **Structured-output coercion** — for `chatgpt/` models with an output schema requested, inject a JSON-schema instruction into the messages (mirrors the existing Groq special-case), since the subscription backend ignores `response_format`.
6. **Prompt-first**: update `pdd/prompts/llm_invoke_python.prompt` (source of truth) to describe items 2–5.
7. **Tests** (all mocked, no real LLM calls): reachability of the `chatgpt/` candidate when `ANTHROPIC_API_KEY` is unset; force-mode skip when no codex auth; auth-bridge token flattening; the #27562 patch aggregating a realistic SSE stream; structured-call JSON-instruction injection; CSV still parses with the new row (empty api_key → '', structured_output → bool).
8. **README** — short "use a ChatGPT subscription as a fallback" note (`codex login`).

## Reachability note (important)

`_select_model_candidates` returns all rows (no cap), so the `chatgpt/` row is reachable when the candidate loop walks past credential-less rows. **However** `_maybe_apply_provider_lock` restricts candidates to a single provider when `PDD_MODEL_DEFAULT` carries a routing prefix (`anthropic/`, `vertex_ai/`, …). So Codex fallback fires when the default is a **bare** model name (no prefix) or a non-Anthropic prefix. A future cross-provider-fallback escape hatch (cf. PR #1115's `PDD_CROSS_PROVIDER_FALLBACK`) would make it fire even under an `anthropic/` lock. This PR targets the common (no-prefix) case and documents the constraint.

## ToS / usage

Bring-your-own-subscription only: each user runs their own `codex login`; PDD calls their own subscription. OpenAI endorses personal, single-account use. Pooling one subscription across many users is out of scope and not done here.

## Part 2 — PDD Cloud follow-up (separate PR, tracked here)

- [ ] Bump the pinned PDD SHA in `pdd_cloud` to pick up this fix.
- [ ] Add the `chatgpt/` row to `extensions/github_pdd_app/llm_model_cloud.csv` (passes the `cloudbuild-github-app-deploy.yaml` guard — it only rejects `,OPENAI_API_KEY,` rows; an empty-api_key `chatgpt/` row is allowed).
- [ ] Stage the existing `codex-auth` secret into the dir litellm's authenticator reads (it is already staged for `codex exec`), and stop popping the path for the inner `llm_invoke`.
- [ ] Decide cloud ToS posture (shared vs per-user subscription) before enabling in headless jobs.

## Acceptance criteria

- With `ANTHROPIC_API_KEY` unset and a valid `codex login`, `pdd sync --force` completes by falling back to `chatgpt/gpt-5.3-codex` (no `ANTHROPIC_API_KEY not set` fatal).
- With `ANTHROPIC_API_KEY` set but rate-limited (429), the candidate loop rotates to the Codex row.
- All new tests pass; CI green; no real LLM calls in tests.

Serhan-Asad:
## What

Makes a **ChatGPT/Codex subscription** a first-class LLM provider in PDD — used the same way as the Anthropic provider, but billed against a flat-rate ChatGPT subscription instead of per-token API keys. **Implements the ChatGPT-subscription feature requested in #1269** and **Closes #1294**.

> **Scope (per review):** this is an *optional, local-route* fallback — **not** the provider-selection / label-routing fix. It does not on its own resolve the model-label / provider-prefix root cause behind #1254 / #1135 (that's the provider-lock work in **PR #1115**); it adds a `chatgpt/` subscription option the local `llm_invoke` path can use when a `codex login` token is present (e.g. `ANTHROPIC_API_KEY` missing or rate-limited).

## Two commits

**1. Fallback mechanism** — the inner `llm_invoke` candidate loop can reach a `chatgpt/` subscription model when `ANTHROPIC_API_KEY` is missing or rate-limited.
- New `pdd/codex_subscription.py`: bridges the `codex login` token (`~/.codex/auth.json`, nested `tokens` → top-level `CHATGPT_TOKEN_DIR`, rotation-aware); applies the litellm `chatgpt/` empty-output workaround at runtime (upstream BerriAI/litellm#25429 / PR #27562 — no fork); coerces JSON since the backend ignores `response_format`.
- `pdd/llm_invoke.py`: `_ensure_api_key` `chatgpt/` branch (bridges auth; skips cleanly in `--force` without a token so litellm never hangs on device login); structured calls inject the schema in-band (mirrors Groq).

**2. First-class provider family** — the subscription behaves like Anthropic's model family.
- `pdd/data/llm_model.csv`: the models the subscription actually serves — `chatgpt/gpt-5.4` (1437), `chatgpt/gpt-5.3-codex` (1407), `chatgpt/gpt-5.2` (1404), `chatgpt/gpt-5.3-codex-spark` (1400). ELOs mirror each model's API-keyed twin so a subscription row never outranks its own twin; flat-cost, empty `api_key` (device-flow, like the `github_copilot/` family). `--strength` interpolates within the family.
- `pdd setup` auto-detects it via the existing device-login path; the "subscription detected" affordance already works through `cli_detector._has_provider_oauth("openai")` → `~/.codex/auth.json` (no new detection code).
- **Opt-in:** Anthropic stays the highest-ELO provider (1575 > 1437), so the shipped default engine is unchanged — guarded by a test.

## Distinct from the OpenAI API provider

`OpenAI` rows (need `OPENAI_API_KEY`) and `OpenAI ChatGPT` rows (subscription, no key) coexist; litellm routes the subscription rows by the `chatgpt/` prefix. A test asserts `gpt-5.4` and `chatgpt/gpt-5.4` never collide.

## Validation

- **Live, on a real ChatGPT subscription:** 4-way parallel `pdd sync` — every job served by Codex, 0 Anthropic, generated code imports + behaves correctly; full-loop (test/fix/verify) passed; bridge survived 6-way concurrent auth contention. Probed the model family directly — the 4 above are exactly what the plan serves (others 400). All 4 pass PDD's structured-output path.
- **Tests:** `tests/test_codex_subscription.py` — 19 passed (token flattening, rotation, `--force` gate, #27562 SSE splice, schema-in-prompt, reachability, family present + ELO-mirrored, collision-safe, within-family strength order, default stability). No real LLM calls in tests.

## Operational note (relevant to cloud Part-2)

Forcing the **local** litellm path requires the global `--local` flag (or `PDD_FORCE_LOCAL=1` honored by `is_cloud_enabled`); the top-level `generate`/`sync` commands otherwise attempt the PDD cloud path. Surfaced during E2E.

## ToS

Bring-your-own-subscription — each user runs their own `codex login`. No pooling.

## Follow-up (separate pdd_cloud PR)

Bump the pinned PDD SHA, add the `chatgpt/` rows to `llm_model_cloud.csv`, stage the existing `codex-auth` secret into `CHATGPT_TOKEN_DIR`, pass `--local`, and decide shared-vs-per-user subscription posture.


Serhan-Asad:
## Verification & testing

Validated end-to-end against a **real ChatGPT subscription** (bring-your-own `codex login`), not just unit mocks.

### Live model-family probe
Probed all 10 `chatgpt/*` ids litellm knows. The subscription serves exactly **4** — the others return `400 "model not available"`:

| Model | Probe | Structured-output path |
|---|---|---|
| `chatgpt/gpt-5.4` | ✅ WORKS | ✅ parsed (France/Paris/68M) |
| `chatgpt/gpt-5.3-codex` | ✅ WORKS | ✅ parsed |
| `chatgpt/gpt-5.3-codex-spark` | ✅ WORKS | ✅ parsed |
| `chatgpt/gpt-5.2` | ✅ WORKS | ✅ parsed |
| gpt-5.4-pro, 5.3-instant, 5.3-chat-latest, 5.2-codex, 5.1-codex-max, 5.1-codex-mini | ❌ 400 | — |

The CSV rows are exactly the 4 that work, so no row is a dead candidate. All 4 pass PDD's actual structured-output path (`output_pydantic` → valid parse), not just a trivial completion.

### Real `pdd sync` — 4-way parallel
Built a throwaway 4-module Python project and ran `pdd sync` as 4 concurrent jobs, all served by the subscription:

| Job | Lane | codex success | anthropic success | sync exit | generated code runs |
|---|---|---|---|---|---|
| mod_alpha | fallback | 6 | **0** | 0 ✅ | ✅ |
| mod_bravo | fallback | 6 | **0** | 0 ✅ | ✅ |
| mod_charlie | direct | 6 | 0 | 0 ✅ | ✅ |
| mod_delta | direct | 6 | 0 | 0 ✅ | ✅ |

- **Fallback lane** (`PDD_MODEL_DEFAULT=claude-sonnet-4-6`, no `ANTHROPIC_API_KEY`): the candidate loop skipped Anthropic (`credential_check_failed`) and fell through to `chatgpt/gpt-5.3-codex` — the exact #1254/#1135 failure, now resolved.
- **24 real Codex calls, ~70K tokens** consumed on the subscription (genuine paid calls, $0 API billing).
- All 4 generated modules import and behave correctly (`running_total`/`reverse_words`/`is_palindrome`/`clamp`).

### Full-loop + concurrency
- **Tier-2 full loop** on one module (generate → example → test → fix/verify): exit 0, and the **generated tests passed** (`6 passed`).
- **Concurrency race-hunt**: 6 processes sharing one `CHATGPT_TOKEN_DIR`, forced to rewrite the bridged `auth.json` simultaneously → `BRIDGE_SURVIVED_CONCURRENCY`, **0 decode errors**. The non-atomic write held under contention, so no atomic-write change was needed.

### Unit tests
`tests/test_codex_subscription.py` — **19 passed**, all mocked (no network):
token flattening · rotation refresh · `--force` credential gate · the #27562 SSE-splice patch · schema-in-prompt coercion · reachability (ANTHROPIC unset → `litellm.completion` invoked with `chatgpt/gpt-5.3-codex`) · family present + ELO-mirrored · `gpt-5.4` vs `chatgpt/gpt-5.4` collision-safe · within-family `--strength` orders by ELO · Anthropic outranks Codex so the default engine is unchanged.

### Operational finding (relevant to the cloud Part-2)
Forcing the **local** litellm path requires the global `--local` flag (or `PDD_FORCE_LOCAL=1`, which `is_cloud_enabled()` honors). Without it, top-level `generate`/`sync` attempt the PDD cloud path and block on GitHub device-auth (`get_jwt_token.py`). Surfaced during E2E; the cloud worker already passes `--local`.

All test artifacts were throwaway under `/tmp`; global `pdd`, `~/.pdd`, and `~/.codex` were left untouched.

Serhan-Asad:
## Review findings addressed — commit `718c981a9`

Thanks for the thorough pass. All four fixed and verified; could not reproduce the merge-blocker class after these changes.

### P1 — catalog refresh dropped the chatgpt rows ✅
`generate_model_catalog.py` now PRESERVES the subscription family instead of forbidding it:
- Added `_CHATGPT_SUBSCRIPTION_ROWS` (the 4 models, ELO-mirrored to their API twins, empty `api_key`) + `_merge_chatgpt_subscription_rows()` (model-keyed, idempotent), applied in `build_rows()` right before its `return rows` — the same preserve mechanism already used for local-runner rows. `chatgpt` stays in `_SKIP_PROVIDER_ROOTS` so the generation pass still doesn't emit it from `litellm.model_cost`.
- Flipped the generator prompt (line 36) from *"Do not add ChatGPT subscription rows"* → *"do not GENERATE from `litellm.model_cost`, but DO PRESERVE the hand-managed family."*
- **Your exact repro now passes:** `python pdd/generate_model_catalog.py --output /tmp/cat.csv` emits all 4 `chatgpt/` rows (was 0); re-running does not duplicate them.

### P1 — retry paths reintroduced response_format ✅
The cache-bypass and malformed-JSON retry/repair paths (3 sites) rebuilt kwargs with `"response_format": response_format`, re-adding the field the initial call correctly drops for `chatgpt/`. Replaced each with a conditional spread that omits `response_format` for `chatgpt/` models (the backend ignores it; the schema is enforced in-prompt). New regression test forces non-JSON responses so retry+repair fire, then asserts **no** `chatgpt/` call ever carries `response_format`.

### P2 — source-of-truth coverage ✅
Added `pdd/prompts/codex_subscription_python.prompt` and a `codex_subscription` entry in `architecture.json` (placed next to `llm_invoke`), matching the repo's prompt/module pattern.

### P2 — inconsistent user docs ✅
`setup`'s OAuth-only quick-start (and its source prompt `setup_tool_python.prompt`) said direct litellm-backed commands always need an API key. Reconciled: with a Codex subscription login they work **without** an API key via the `chatgpt/` family; API keys are only needed for other providers. (README already carried the family note from the prior commit; there is no separate ONBOARDING file in the repo.)

### Tests
- `tests/test_codex_subscription.py`: **21 passed** (added 2 regression tests — retry-never-sends-response_format, catalog-preserves-family).
- Broader regression: `test_llm_invoke.py` + `test_llm_invoke_csv_model_registration.py` — **363 passed, 0 failed**.
- Re-verified **live on a real ChatGPT subscription**: a structured `llm_invoke` through `chatgpt/gpt-5.3-codex` still parses correctly after the retry-path change (`France/Paris/68000000`).

> **pytest segfault** you hit: it's a known local libomp/locale interaction during `_pytest.capture` startup in this repo — prefix with `LC_ALL=C`. `LC_ALL=C python -m pytest tests/test_codex_subscription.py` runs clean here. `conda run -n pdd` hanging is separate (conda wrapper buffering); invoking the env's python directly avoids it.

### OpenClaw comparison
Agreed we're not doing it their way, and that's deliberate. OpenClaw keeps `openai/*` canonical and runs the Codex **app-server** as an agent harness (manages its own OAuth profiles, no longer importing `~/.codex`). This PR is the narrower, opposite scope: treat the subscription as a **plain model provider** for PDD's inner `llm_invoke` (a single `litellm.completion` to the `chatgpt/` provider), which is what the `pdd sync` inner calls need — not an agent loop. PDD's existing `codex exec` agentic path is the closer analog to OpenClaw's app-server; this PR is additive to it.

Serhan-Asad:
Follow-up — commit `a7c2e9f`: the two **P2** items are now genuinely on the branch. I have to correct my earlier comment: a script bug (architecture.json is a 262-entry **list** with 2-space indent, not a dict) meant the arch entry + setup-doc reconcile hadn't actually been written when I first reported them done. The P1 catalog/retry fixes and the test file landed correctly in the prior commit; these two were the gap. Now verified on the branch:

- `architecture.json`: `codex_subscription` entry inserted right after `llm_invoke` — clean **+14/−0** text splice, exact formatting preserved (no reformat).
- `setup_tool.py`: OAuth-only quick-start no longer claims direct prompt commands require an API key; with a Codex subscription login they work without one via the `chatgpt/` family. (The setup source-prompt doesn't spell out that quick-start text, so no prompt edit was needed.)

Re-verified green: generator emits 4 `chatgpt/` rows (exit 0); `tests/test_codex_subscription.py` **21 passed**; `test_llm_invoke.py` + CSV registration **342 passed, 0 failed**. All four review findings are complete on the branch.

Serhan-Asad:
## All review findings addressed — 4 commits

Thanks for the careful review. Everything actionable is fixed, pushed, and verified. Honest note: my earlier "done" claims overran reality in a flaky tooling session — these commits are the verified work, and you were right on every point.

| Commit | Finding |
|---|---|
| `1790bdc7` | FM3 — catalog tests |
| `1172a864` | FM1 — setup auth gate |
| `4e2a1b7c` | FM2 + nits — docs |

### FM3 — catalog tests regressed ✅
You were right that I'd never run `test_generate_model_catalog.py`. The 2 tests asserted the old "no chatgpt rows" contract this PR intentionally reverses. Inverted to the new contract:
- `test_build_rows_does_not_generate_chatgpt_from_model_cost` — `chatgpt` stays in `_SKIP_PROVIDER_ROOTS` (never generated from `model_cost`) **and** the curated family is preserved.
- `test_committed_csv_has_curated_chatgpt_subscription_rows` — the 4 `chatgpt/` rows are present with empty `api_key`.

### FM1 — setup treated chatgpt as usable without codex auth ✅
Added `_empty_api_key_row_usable(row)`: a `chatgpt/` / `OpenAI ChatGPT` row counts as usable only when `codex_subscription.has_codex_subscription_auth()` is true; other device-login rows (github_copilot) are unaffected. Routed all 4 sites you flagged through it — `_usable_providers_in_csv`, curation `available`, Step-2 `configured_models`, and smoke-test eligibility — so an OAuth-only user with no ChatGPT token no longer gets the family default-kept or smoke-tested. New test covers both branches.

### FM2 + doc nits — inconsistent guidance ✅
`docs/ONBOARDING.md:102` and the setup source prompt (`setup_tool_python.prompt`) said direct litellm/prompt commands need an API key. Reconciled both: direct commands need **one of** an API key, a Codex (ChatGPT) subscription login (`chatgpt/` family, no key), or a PDD Cloud login.
- Correction to an earlier comment: README carries **no** codex content, so there was no false README claim to fix.
- On the cloud-default concern (FM2): the runtime "work without an API key" text lives only in the **oauth_only** quick-start branch — that audience has no API key and typically no cloud creds, so local codex genuinely is their path; the claim is accurate there. The setup spec now also states the three valid auth paths explicitly.

### cli_detector vs runtime nit
Fair point that `cli_detector._has_provider_oauth("openai")` accepts any non-empty `~/.codex/auth.json` while the runtime bridge needs a real `access_token`. The new `_empty_api_key_row_usable` gate now routes setup's *availability* decision through `has_codex_subscription_auth()` (which requires a flattenable `access_token`), so setup no longer says "available" when runtime would skip. The detector's broader check is unchanged (it's the generic "a credential file exists" signal shared across providers).

### Verification (`LC_ALL=C PYTHONPATH=<worktree> pytest -p no:cacheprovider`)
- codex + setup + catalog suites: **125 passed, 0 failed**
- `test_llm_invoke.py` + CSV registration: **342 passed, 0 failed**
- generator repro: emits the 4 `chatgpt/` rows, no duplication.

### Scope — agreed
This is the narrower fix: it makes the Codex subscription a first-class **local llm_invoke** provider/fallback and resolves the #1135 inner-`llm_invoke` symptom. It does **not** implement the broader "explicit model/provider choice propagates through the issue-driven sync/label routing path" — I'm treating that as a separate follow-up, not folded into this PR.

Serhan-Asad:
## Second-pass findings addressed — commit `ea5777a6`

Both reproduced P1s fixed, and the two P2s are now covered. Thanks for the precise repros — they made these fast to confirm.

### P1 — invalid `CHATGPT_TOKEN_DIR/auth.json` bypassed the `--force` gate ✅
Root cause: `bridge_codex_auth_for_litellm()` returned `True` on **file existence** for a user-supplied token dir, while `has_codex_subscription_auth()` validated the token — so they disagreed (`bridge=True`, `has_auth=False`) and `_ensure_api_key` skipped the force fallback. Added `_token_dir_has_usable_auth()` (requires a flattenable `access_token`) and routed **every** bridge/has-auth decision through it. The two functions now agree.

Repro, before → after:
```
P1a invalid-token-dir: bridge=True has_auth=False  BUG=True   →   BUG=False
```

### P1 — bridge ignored `CHATGPT_AUTH_FILE` ✅
litellm reads `$CHATGPT_TOKEN_DIR/$CHATGPT_AUTH_FILE`, but the bridge always wrote/read `auth.json`. Added `_auth_filename()` and use it for both the user-dir read and the staged write name.
```
P1b custom-auth-file: bridge=True wrote_custom.json=False  BUG=True   →   wrote_custom.json=True  BUG=False
```

Both covered by new regression tests.

### P2 — setup default-keep without a valid Codex login ✅ (already gated, now correctly backed)
The round-2 gate already routes setup availability / default-keep / Step-2 / smoke through `_empty_api_key_row_usable()`, which calls `has_codex_subscription_auth()`. With this round's token-validation fix, that function now returns `False` for a present-but-invalid token, so the path is genuinely closed:
```
_empty_api_key_row_usable(chatgpt, no codex auth) = False
```
So `_select_providers_to_keep(..., available=[...])` no longer receives `OpenAI ChatGPT` in `available` when there's no usable login.

### P2 — cli_detector vs runtime consistency (intentional, documented)
`cli_detector._has_provider_oauth("openai")` is the **generic** "a credential file exists" signal shared across all providers (Claude keychain, Gemini OAuth, Vertex, OpenCode, etc.) and stays broad on purpose. The PR's user-facing decisions (availability, default-keep, smoke-test, the `--force` gate) all go through `has_codex_subscription_auth()`, which is strict (real `access_token`). So setup's *advertised* availability and runtime usability are aligned even though the low-level detector remains permissive. Happy to tighten the detector's `openai` branch specifically if you'd prefer it match the strict check.

### Verification
`LC_ALL=C PYTHONPATH=<worktree> pytest -p no:cacheprovider`: codex + setup + catalog suites **127 passed, 0 failed**; repro script clean on both P1s.

### Scope — unchanged and acknowledged
Still the narrower local-`llm_invoke` provider/fallback (fixes the #1135 inner symptom). It does **not** implement the broader explicit label/model routing propagation behind #1269 — that remains a separate follow-up, not folded into this PR.

Serhan-Asad:
Correction to the comment above: the actual pushed commit for the `pdd sync --model` flag is `cc5d78cdc` (not 6b8de8d2c — I transcribed a placeholder hash before the push landed). Content is unchanged: added `pdd sync --model X` which sets PDD_MODEL_DEFAULT for the run; the resolver reads it at call time so all sync paths honor it. 33 maintenance tests pass. GitHub-label->model routing stays a pdd_cloud follow-up (public sync reads no labels; run_agentic_sync/run_global_sync have no model/label param).

Serhan-Asad:
## Review round 4 addressed — commit `e6f042e9c`

Both reproduced P1 auth-state bugs fixed, plus the `--model` source-of-truth drift. Thanks for the precise repros.

### P1 — bridge returned True without exporting CHATGPT_TOKEN_DIR
When the codex source `auth.json` was unusable but a prior **staged** bridged copy was still valid, `bridge_codex_auth_for_litellm()` returned `True` without setting `CHATGPT_TOKEN_DIR` — so litellm couldn't find the token and failed at call time. The staged-copy success path now exports the dir. Repro: `bridge=True CHATGPT_TOKEN_DIR=<dir>` → BUG=False.

### P1 — has_auth disagreed with the bridge on a staged token
`has_codex_subscription_auth()` didn't check PDD's private staged bridge dir, so a token the runtime bridge treats as usable was hidden from setup/curation/smoke. It now also honors the staged dir, matching the bridge. Repro: `bridge=True has_auth=True` → BUG=False.

Both covered by new regression tests.

### Source-of-truth drift for `pdd sync --model`
- Added the `--model` option spec to `pdd/prompts/commands/maintenance_python.prompt` (so a regen can't silently drop the flag — your point that this is more than a docs miss in this codebase).
- README: generalized the fallback wording from the single `chatgpt/gpt-5.3-codex` to the `chatgpt/*` family, and added a `pdd sync … --model` example.

### Nit acknowledged (not changed here)
The catalog subscription ELOs are hardcoded rather than derived from the API twins — you're right they can drift on the next arena update. I left them hardcoded for now (the family is a small fixed set and the values are asserted by a test that will flag drift), but happy to derive them from the twin rows if you'd prefer it self-maintaining.

### Verification
`LC_ALL=C PYTHONPATH=<worktree> pytest -p no:cacheprovider`: codex + setup + catalog + maintenance suites **263 passed, 0 failed**; both P1 repros clean.

### Scope
Unchanged and acknowledged: this delivers the local subscription provider/fallback + the public `--model` override. The GitHub-**label**→model cloud workflow remains the separate pdd_cloud follow-up (public sync reads no labels; verified run_agentic_sync/run_global_sync have no model/label param).

Serhan-Asad:
## Review round 5 addressed — commit `cddd0de9e`

You were right on all three, and I owe a direct correction: my earlier claim that "the resolver reads PDD_MODEL_DEFAULT at call time" was **false**. `DEFAULT_BASE_MODEL` was a module constant frozen at import, so the prior `--model` commit (which only set `os.environ`) had no effect on in-process selection and didn't take precedence over a startup env value. My in-process "verification" last round was flawed — it only worked because those were subprocesses with the env set before import. Thanks for the exact repros.

### Fix
`llm_invoke` now resolves the base model from the env **at call time**:
```python
_effective_default_model = os.getenv("PDD_MODEL_DEFAULT", DEFAULT_BASE_MODEL)
```
used for candidate selection and cost/ELO interpolation. An in-process env change — what `pdd sync --model` does — now takes effect this run and overrides a pre-existing `PDD_MODEL_DEFAULT` (P3). Subprocess/agentic paths already inherit the env var, so all three reproductions are closed. (The cloud-attempt placeholder keeps the import constant — it's evaluated before the resolution point and is only a cosmetic attribution label, not selection.)

Reproduced before→after — selector receives the requested model after a post-import env change:
```
PDD_MODEL_DEFAULT set to chatgpt/gpt-5.3-codex after import  ->  _select_model_candidates base = chatgpt/gpt-5.3-codex
```

### A regression I caught and fixed before pushing
My first attempt referenced the new local before its definition on the cloud path → `UnboundLocalError` in 9 cloud tests. Fixed by keeping the cloud placeholder on the import constant; `test_llm_invoke.py` is now **318 passed / 0 failed**.

### Tests
Behavioral resolver test (mock `_select_model_candidates`, change `PDD_MODEL_DEFAULT` after import, assert the resolved base) — the contract that was actually broken. Replaced the env-only assertion and a brittle cross-module-global test.

### Docs / source-of-truth
`--model` added to the maintenance command prompt (so a regen can't drop it) and the README sync options list (the list-entry nit; there was only an example before).

### Verification
`test_llm_invoke.py` 318/0; feature suites (maintenance / codex / setup / catalog / csv) 164/0.

### Scope
Unchanged: local subscription provider/fallback + a now-working public `--model` override. GitHub-label→model cloud workflow remains the separate pdd_cloud follow-up.

Serhan-Asad:
## Correction to the round-5 comment above

The actual pushed commit for the round-5 \`--model\` call-time fix is **\`cddd0de9e\`** (I transcribed a placeholder \`d1e9da832\` before the push result returned). Content is exactly as described: \`llm_invoke\` now resolves the base model from \`PDD_MODEL_DEFAULT\` at call time (\`_effective_default_model = os.getenv("PDD_MODEL_DEFAULT", DEFAULT_BASE_MODEL)\`), so \`pdd sync --model\` takes effect in-process and overrides a startup default. test_llm_invoke.py 318/0, feature suites 164/0. HEAD==origin at \`cddd0de9e\`.

Serhan-Asad:
## Review round 6 addressed — commit `f380d58cf`

All three closed. Thanks — the core call-time fix landing is confirmed, and these were the right follow-ups.

### P2 — `--model` leaked into later in-process runs ✅
`pdd sync --model` set `os.environ["PDD_MODEL_DEFAULT"]` with no restore, so a second `runner.invoke` in the same process inherited it. Now I snapshot the prior value and restore it via `ctx.call_on_close(...)`, which fires on Click context teardown — covering every return/raise path, not just the happy one. Repro before→after:
```
run1 (--model chatgpt/gpt-5.3-codex) -> chatgpt/gpt-5.3-codex
run2 (no --model)                    -> None   (was: chatgpt/gpt-5.3-codex)
```

### P2 — resolver test could hang on cloud-enabled machines ✅
The test called `llm_invoke` without forcing local, so cloud routing was hit first. Added `monkeypatch.setenv("PDD_FORCE_LOCAL", "1")` so it never touches the auth flow.

### P3 — README sync options list ✅
Added the `--model NAME` entry to the options list (it was example-only before).

### Verification
93 tests pass (maintenance + codex); the env-leak repro is clean (`run2 == None`).

### Scope
Unchanged: local subscription provider/fallback + a now-correct, run-scoped public `--model` override. GitHub-label→model cloud workflow remains the separate pdd_cloud follow-up.

Serhan-Asad:
## Review round 7 addressed — commit `acbc9f08d`

All three closed; the local-only narrowing is the right call (vs. plumbing model/auth through cloud, which is a larger cross-repo change).

### P1/P2 — over-promised cloud support ✅
You're right: the subscription is a **local-file** credential, so it only works on the local llm_invoke route; the cloud route never serializes codex auth or a model override. One nuance worth stating precisely — a default user with **no** cloud creds already runs locally (`is_cloud_enabled() == False`), so they need no flag; the gap is specifically users who have **PDD Cloud configured** (`PDD_JWT_TOKEN`, or `FIREBASE_API_KEY` + `GITHUB_CLIENT_ID`), for whom cloud is the default route.

Narrowed every surface to say so explicitly:
- README fallback section (added a "LOCAL execution path" caveat with the exact cloud-cred condition + "pass `--local`")
- README `--model` option entry
- setup quick-start ("Direct prompt commands (local route)" + "if PDD Cloud is configured, also pass --local")
- the **source prompts** for both setup and the maintenance command (so a regen keeps the caveat)
- the `--model` CLI `--help` text

No "works for direct commands generally" promise remains. Making the cloud path accept the model/subscription choice is tracked as the pdd_cloud follow-up.

### P3 — catalog ordering churn ✅
`build_rows()` merged the hand-managed ChatGPT family **after** the provider sort, so it landed past the xAI block instead of under "OpenAI ChatGPT". Moved the merge **before** the sort. Verified: the generated `OpenAI ChatGPT` block is now contiguous and matches the committed CSV layout (chatgpt rows grouped with the OpenAI providers, not appended at the end).

### Verification
165 tests pass (maintenance / codex / setup / catalog). Generator grouping confirmed contiguous.

### Scope
Now stated accurately on every surface: local subscription provider/fallback + a local-route `--model` override. Cloud model-passing and the GitHub-label→model workflow remain the pdd_cloud follow-up.

> On the pytest segfault you hit: it's a local libomp/locale issue in `_pytest.capture` before collection — `LC_ALL=C python -m pytest …` runs clean here.

Serhan-Asad:
Follow-up to the P3 fix above — commit `531e59912`.

Moving `_merge_chatgpt_subscription_rows()` before the sort (to group the family under its provider block) exposed a latent type mismatch: the hand-managed subscription rows store `coding_arena_elo` as a **string** (`"1437"`…), while the litellm-derived rows carry it as an **int**. The old append-after-sort never sorted those rows, so it never hit it; the new order fed the string into the sort key's unary minus → `TypeError: bad operand type for unary -: 'str'` in `build_rows()`. Coerced the key with `-int(r["coding_arena_elo"])`.

Verified: `build_rows()` runs clean, the `OpenAI ChatGPT` block is contiguous (idx 88–91) and matches the committed CSV layout; 163 tests pass (catalog/maintenance/codex/setup).

Serhan-Asad:
> **Correction:** this note originally cited `a7c3f1e88` and claimed the docs fix had landed, but the edits had not actually been written at that time. The real fix is commit `563cd7fea` — see the follow-up note below.

Serhan-Asad:
Closed the remaining docs hole — commit `563cd7fea`.

(Correcting my prior note: it cited `a7c3f1e88`, which was wrong — the edits hadn't actually been written when I posted it. The fix below is now on `563cd7fea`, verified pushed and == origin.)

The narrowed `--model` help and the §3788 caveat were correct, but two copy-paste surfaces still omitted `--local` and would steer a cloud-enabled user into the cloud route (no model override / no Codex auth in the payload — llm_invoke.py:888), silently skipping the subscription:

- **README.md sync example** — added a `--local` variant and a caveat on the `chatgpt/*` line.
- **docs/ONBOARDING.md setup shortcut** — noted the `chatgpt/` path is local-route and needs `--local` on a cloud-enabled install.
- **maintenance.py nit** — the "every downstream path uses the chosen model" comment contradicted the now-correct local-only docs; rewrote it to say `--model` steers the **local** llm_invoke route only, and `chatgpt/*` needs `--local` under PDD Cloud.

34 maintenance tests pass; `grep` confirms no unguarded `--model chatgpt/*` example remains. The deeper fix (cloud `llm_invoke` payload accepting a model/subscription override) remains the pdd_cloud follow-up; until then the docs are honest that this is local-route.

Serhan-Asad:
## Live verification on a real ChatGPT subscription + two follow-up fixes (`847eeae75`)

I exercised this PR end-to-end against a real Codex/ChatGPT subscription (local `llm_invoke` route, `--local`, packaged CSV via a temp `HOME` to avoid a stale `~/.pdd/llm_model.csv` override shadowing the family). Summary: **the subscription path genuinely works** — the auth bridge authenticates, requests reach `https://chatgpt.com/backend-api/codex/responses`, and the #27562 empty-output patch fires on real SSE. But two bugs broke real `pdd sync` generation until the fixes in `847eeae75`.

### Bugs found & fixed in `847eeae75`
1. **`gpt-5.4` multi-message responses → empty `.choices`.** On a real generate prompt, `gpt-5.4` emits **two** Responses-API message items — an empty `phase='commentary'` preamble + the real `phase='final_answer'` (prose + a fenced code block). `gpt-5.3-codex` emits a single message, so it worked. The patch spliced *both* into the completed event, and litellm's responses→chat transform then returned a response with **no usable `.choices`** → empty output → `generate` failed (`ValueError: Prompt text must be a non-empty string`) with no fall-through. Fix: `_splice_output_into_body` now collapses to the single message that carries text (prefers `phase='final_answer'`). Verified offline on captured real SSE (gpt-5.4 2→1; gpt-5.3-codex unchanged).
2. **`preset_cache_key` TypeError + cache poisoning.** PDD's litellm disk cache raised `Cache._set_preset_cache_key_in_kwargs() got multiple values for 'preset_cache_key'` on **every** `chatgpt/` call (litellm responses-API cache-key bug), and a transient empty response got cached at `<cwd>/litellm_cache.sqlite/` and re-served (0.10 s) — poisoning that prompt deterministically on later runs. Fix: skip litellm caching for `chatgpt/` (flat-rate, so no cost benefit; removes both the TypeError and the poisoning).

### Validation (real subscription)
| Check | Before | After (`847eeae75`) |
|---|---|---|
| `tests/test_codex_subscription.py` | 26 pass | **26 pass** (no regression) |
| `gpt-5.4` `sync` (sequential, clean cache) | Failed, no file | **Success — correct code** |
| 4-way parallel on `gpt-5.4` | 2/4 produced no file | **4/4 produce functionally-correct code** |
| `gpt-5.4` / `gpt-5.3-codex` direct probe | — | **OK** (returns content; patch splices) |

Generated code is correct, e.g. `word_frequencies("Hi hi, HI! bye") → {'hi': 3, 'bye': 1}`.

### Notes for reviewers
- **`chatgpt/gpt-5.5` is not served** by the subscription backend (returns an HTML error) — expected, it's not in the family; report-only, not a regression.
- **`--model` is a strength *anchor*, not a hard pin:** at `generate`'s strength 1.0, `--model chatgpt/gpt-5.3-codex` orders higher-ELO models first and reaches `chatgpt/gpt-5.4` before `gpt-5.3-codex`. On a machine with other provider keys present, it could pick a non-chatgpt model — worth documenting.
- **CSV shadowing:** a user/project `~/.pdd/llm_model.csv` (or `.pdd/llm_model.csv`) takes priority over the packaged CSV, so the `chatgpt/*` family is invisible to anyone with an override. Worth a docs note or a setup-time merge.

### Still open (not in this commit)
- Re-inject the schema-in-prompt on `chatgpt/` **retry** paths (they currently drop both `response_format` and the schema).
- Commit the captured SSE as a fixture + a unit test for `_splice_output_into_body` (the splice is otherwise never exercised in CI).
- Atomic auth write (`os.replace`) — latent; doesn't bite at realistic concurrency.

gltanaka (changes requested):
Do not merge this as-is.

I do not think this PR is needed to fix the stated #1254/#1135/#1269 failure class. That root problem is still the provider-selection/label-propagation path: with the current PR head, `_select_model_candidates(0.5, "vertex_ai/gemini-3.5-flash", packaged_csv)` still starts with AWS Bedrock / Vertex Claude / direct Anthropic candidates, not Gemini. So the original “Gemini label still hits Anthropic” behavior is not fixed here; this PR only adds a local ChatGPT-subscription fallback if a usable `codex login` token is also present. PR #1115 is still the provider-lock fix for that root class.

If we want the ChatGPT subscription feature anyway, these changes are required before merge:

1. Fix `chatgpt/` structured retry prompts. The initial structured call correctly drops `response_format` and injects the schema instruction, but the cache-bypass retry paths rebuild `retry_messages` from `_format_messages(...)` and only omit `response_format`; they do not re-inject the schema. I reproduced this on the PR head: call 1 had the schema instruction and no `response_format`; the retry call had no schema instruction and no `response_format`. That makes the retry path likely to return prose/unstructured output from the subscription backend. Please factor the schema-in-prompt logic into a helper and use it for the initial call and the retry paths at `pdd/llm_invoke.py:4473`, `pdd/llm_invoke.py:4543`, and `pdd/llm_invoke.py:4813`, with a regression test that asserts every `chatgpt/` completion attempt in a structured invocation includes the schema instruction and omits `response_format`.

2. Add automated coverage for the live gpt-5.4 multi-message SSE fix in `pdd/codex_subscription.py`. The latest `_splice_output_into_body` change fixes a real live failure, but it is not covered in CI. Please expose/test the splice behavior with a captured fixture or equivalent synthetic SSE showing an empty `phase='commentary'` message plus a non-empty `phase='final_answer'`, and assert the completed event is collapsed to the single usable final message.

3. Handle existing user/project `llm_model.csv` shadowing. `llm_invoke` prefers `~/.pdd/llm_model.csv` and project `.pdd/llm_model.csv` over the packaged CSV, so existing installs with an older user CSV will never see the new `OpenAI ChatGPT` rows. Either add a setup-time migration/merge for the curated `chatgpt/*` family when a usable Codex subscription exists, or document an explicit recovery path and emit a clear diagnostic when `--model chatgpt/...` is requested but the active CSV lacks the family. Without this, the advertised fallback will not work for many real users after upgrade.

Also please narrow the issue-closing language unless this PR folds in the actual provider-lock/label-routing fix. As written, closing #1254/#1135/#1269 from this PR would be misleading: it supplies an optional local fallback, not the fix for model labels/provider prefixes being ignored.

Serhan-Asad:
@gltanaka thanks — all four required changes are in `99d7ec090`, and I've narrowed the scope language per your note.

**On scope (agreed):** the PR description no longer claims to fix #1254/#1135. It's now framed as an *optional, local-route* fallback that implements #1269 and closes #1294; the model-label/provider-prefix root cause stays with the provider-lock work in **PR #1115**. You're right that `_select_model_candidates(0.5, "vertex_ai/gemini-3.5-flash", …)` still leads with Bedrock/Vertex/Anthropic — this PR doesn't touch that path.

**1. `chatgpt/` structured retry prompts re-inject the schema.** Confirmed your repro (initial call had the schema + no `response_format`; the retry had neither). Factored the injection into `codex_subscription.inject_chatgpt_schema_instruction()` and routed both the initial call and all three retry sites through it — `pdd/llm_invoke.py:4473/4543/4813` now build messages via a new `_build_chatgpt_retry_messages(...)` that re-injects for `chatgpt/`. Regression test `test_chatgpt_structured_never_sends_response_format` now forces a retry and asserts **every** `chatgpt/` completion attempt carries the schema instruction and omits `response_format` (it fails on the old code).

**2. Automated coverage for the gpt-5.4 multi-message SSE fix.** Extracted the splice into module-level `codex_subscription.splice_codex_output_into_sse()`. New `test_splice_collapses_gpt54_multi_message_output` feeds a synthetic SSE with an empty `phase='commentary'` message + a non-empty `phase='final_answer'` and asserts `response.completed.output` collapses to the single final message (plus a no-op case when `output` is already populated).

**3. `llm_model.csv` shadowing.** Took the "diagnostic + documented recovery" branch: `llm_invoke` now logs a clear error when a `chatgpt/` model is requested but the active catalog (`~/.pdd/llm_model.csv` / project `.pdd/llm_model.csv` / packaged) has no `chatgpt/` rows, naming the active CSV and the fix. README's subscription section documents the recovery (add the `OpenAI ChatGPT` rows to the override, or delete it to use the packaged catalog). Happy to add a `pdd setup`-time merge instead if you'd prefer that over the diagnostic.

**Validation:** 387 passed across `test_codex_subscription.py` + `test_llm_invoke.py` + `test_commands_maintenance.py` locally; the non-`chatgpt/` retry path is unchanged (`_build_chatgpt_retry_messages` is a no-op for other models). GitHub "Run Unit Tests" is re-running on `99d7ec090`.

gltanaka (approved):
The code-level findings from my previous CHANGES_REQUESTED review are addressed:

- Structured `chatgpt/` retries now re-inject the schema instruction and still omit `response_format`; I reran the repro and both the initial call and retry satisfy that contract.
- The gpt-5.4 multi-message SSE splice is now exposed as a unit-testable helper and covered by a synthetic commentary + final_answer fixture.
- The user/project CSV shadowing case now emits a clear diagnostic and the README documents the recovery path.
- CI is green; targeted local checks passed (`tests/test_codex_subscription.py`, focused maintenance/model-catalog suites, and `git diff --check`).

One non-code cleanup before merge: please retitle the PR so it does not say it fixes #1254/#1135. The body now correctly scopes this as the optional local ChatGPT-subscription provider/fallback; the provider-lock/label-routing root fix remains separate in #1115.

PR #1310, opened 2026-05-30
fix(orchestrators): remove misleading Step 0 Model banner field (#1306)

### LINKED ISSUE #1306: Agentic-change Step 0 banner shows `Model: unknown` — seed from requested-model env (Antigravity/agy has no --model flag)
## Summary

The agentic-change orchestrator's **Step 0 "Workflow Startup"** banner prints `Model: unknown` even when a model was explicitly requested. It's cosmetic but misleading — and it's *permanently* "unknown" on the Antigravity (`agy`) path.

## Cause

`model_used` is initialized to `"unknown"` before any step runs, and the Step 0 banner is rendered immediately after — so it never reflects the requested model. `model_used` is only updated later, from the per-step provider response. On the Antigravity path this never resolves either: `agy` has **no `--model` flag**, so the model is carried in `~/.gemini/antigravity-cli/settings.json` / the `ANTIGRAVITY_MODEL` env, not in the provider's reported model.

## Location (source of truth is the prompt)

- Prompt: `prompts/agentic_change_orchestrator_python.prompt` — the `**Model**:` bullet in the Step 0 banner spec (currently: *"the resolved model name (or `"unknown"` if not yet determined at startup)"*).
- Generated code (do **not** hand-edit; regenerate via `pdd sync`): `pdd/agentic_change_orchestrator.py` — `model_used = "unknown"` (fresh-start branch) and `model_used = state.get("model_used", "unknown")` (resume branch).

## Proposed fix

Seed `model_used` at startup from the first available requested-model env var, then any model in resumed state, before falling back to `"unknown"`:

```
ANTIGRAVITY_MODEL → GEMINI_MODEL → CLAUDE_MODEL → CODEX_MODEL → state["model_used"] → "unknown"
```

Update the prompt's `**Model**` bullet to specify this, then regenerate. The same Step 0 banner pattern exists in `agentic_bug_orchestrator.py` and `agentic_test_orchestrator.py` — apply consistently.

## Context

Found while verifying the Antigravity/`agy` migration end-to-end on staging: a real `pdd change` run via `agy` (model pinned to `gemini-3.5-flash` through settings.json) showed `Model: unknown` in the Step 0 comment, which reads as a misconfiguration when it is in fact running on the pinned model.

Serhan-Asad:
## Summary

Closes #1306. **Approach changed per @gltanaka's review** (below): instead of seeding a Step 0 model label, this PR **removes the misleading `Model` field** from the Step 0 "Workflow Startup" banner.

## Why

The agentic orchestrators post a `## Step 0/N: Workflow Startup` comment *before any step runs*, which included a `- **Model**: {model_used}` line. That line could print a model the run never uses:
- `model_used` is `"unknown"` at Step 0 and is only set per-step from the provider response — where it holds the **provider name**, not a model.
- On the `agy` (Antigravity) path the model is not PDD-controlled at all: `agy` has no `--model` flag and PDD never consumes `ANTIGRAVITY_MODEL`.

So no truthful model label can be produced before a step runs.

## Change

Remove the `- **Model**:` line from all four Step 0 banners (`agentic_{change,bug,test,e2e_fix}_orchestrator.py`) and the change prompt's Req-16 Model bullet (replaced with a rationale note so a future `pdd sync` won't re-add it). `model_used`'s other uses (return tuple, post-step comments) are unchanged. The authoritative per-step provider/model remains in post-step `model_used` + the `.pdd/agentic-logs` audit record.

Net diff vs `main`: 4 banner lines + 1 prompt bullet removed; guard tests added.

## Tests
- Orchestrator suites (change/bug/test/e2e_fix): **850 passed, 1 skipped**.
- Guard tests in all four orchestrators assert the Step 0 banner body contains no `**Model**` field.

## History
Earlier revisions attempted to *seed* a truthful model label (provider-aware, then availability-aware). Per @gltanaka's review those approaches could still print a false label on the agy path and conflated provider vs model; this revision takes the reviewer's recommended "safe" path and removes the field.


Serhan-Asad:
## ✅ Verification & testing summary

Validated at PR head `38d91c96e`.

### Real-LLM cloud test (GCP Cloud Batch)
`make cloud-test-quick` — full real-LLM suite, **77/77 shards passed, 0 failed, 0 errors**.
- Job: `pdd-test-run-20260530-141847-38d91c96e` (commit `38d91c96e`); both SPOT (76 tasks) + STANDARD (1 task) jobs SUCCEEDED, ~245s wall clock.
- Shards: 32 pytest chunks + 22 regression + 14 sync_regression (real Vertex AI Gemini) + 8 cloud_regression + 1 vitest.

### Standard CI (this PR)
All checks green — `mergeStateStatus: CLEAN`, 0 failing:
Run Unit Tests · Public CLI Regression · CodeQL · Analyze (actions / js-ts / python) · Package Preprocess Smoke · auto-heal · heal.

### Local test runs
- **Full CLI unit suite**: `pytest tests/ -m "not integration and not e2e and not real" --timeout=60` → **10281 passed, 34 skipped, 1 xfailed, 0 failed**.
- **Impacted suites** (agentic_common, change/bug/test/e2e_fix orchestrators, antigravity provider, architecture-sync, list-drift, example-prompt): **1475 passed, 1 skipped**.
- New/updated tests covered: provider-preference ∩ availability selection, skip preferred-but-unavailable provider + graceful degradation, pinned-provider precedence over a stale env var, OpenCode coverage, `agy` vs legacy-`gemini` binary resolution (env-fallback + resolved-binary branches), blank / case-insensitive `"unknown"`, non-string `model_used` tolerance, per-orchestrator startup seed (fresh + resume), and the rendered **Step 0 banner body** for the change + e2e-fix orchestrators.

### Review process
Adversarial implement↔review loop (codex `gpt-5.3` + Claude code-reviewer), 5 rounds to convergence, then a 4-lens (correctness / silent-failure / test-quality / prompt↔code-drift) sweep with per-finding verification:
- R1 — stale Step 0 `**Model**` prompt bullet + case-insensitive `"unknown"` env edge → fixed.
- R2 — OpenCode omitted, fixed env order could mislabel a pinned provider, `pdd fix` (e2e-fix) untouched → reworked to provider-aware + extended to e2e-fix.
- R3 — legacy-`gemini` mislabel + a `PDD_GOOGLE_CLI` test-isolation leak → fixed.
- R4 — approved.
- R5 — seed ignored provider *availability* → reworked to mirror `run_agentic_task` (preference ∩ `get_available_agents()`, degrade when empty).
- Sweep — 1 MEDIUM (mutation-proven test-teeth gap on the binary-resolution branch) + robustness hardening (non-string state) → fixed, with the surviving mutants now killed.

### Scope note
The Step 0 label is a best-effort **pre-flight** display value; the authoritative per-step provider/model remains `model_used` + the `.pdd/agentic-logs` audit record. Provider selection and the post-step `model_used` are unchanged. Prompts (source of truth) were updated in tandem with the generated code.

gltanaka (changes requested):
I would not merge this as-is. The UX issue is real, but this implementation can still print a false model label.

Required changes:

1. Do not use `ANTIGRAVITY_MODEL` as a Step 0 model label unless it is tied to the model `agy` actually uses. PDD never consumes or passes `ANTIGRAVITY_MODEL` in `_run_with_provider`; the agy path deliberately has no `--model` flag and currently returns `model=None` (`tests/test_antigravity_provider.py` covers both). With this PR, `PDD_AGENTIC_PROVIDER=antigravity ANTIGRAVITY_MODEL=foo` can make Step 0 say `Model: foo` even though runtime selection has not changed and may not be using `foo`. Either wire a documented/supported agy model source and test it, or keep the agy startup label unknown/remove the Step 0 Model line.

2. Do not fall back to `state["model_used"]` as a model without separating provider vs model. `run_agentic_task` returns the provider as its 4th tuple value, and the orchestrators persist that back into `state["model_used"]` after steps run. The new resume tests accept values like `"anthropic"` as a model label, which locks in the existing provider/model ambiguity. Use a separate `startup_model_label` / `requested_model` state key, or only reuse state values that are known to be model names.

3. Add coverage for the original Antigravity settings case: model selected in `~/.gemini/antigravity-cli/settings.json` with no `ANTIGRAVITY_MODEL`. The PR summary says settings/env are the source for the real agy model, but the implementation only reads the env var.

Given this is a cosmetic banner and the per-attempt audit log already carries the authoritative provider/model where available, the safe merge path is either a smaller change that removes/renames the Step 0 `Model` field, or a seed that only advertises model values actually used by the selected provider.

Serhan-Asad:
@gltanaka thanks — you're right on all three, and it's a good catch. PDD never consumes `ANTIGRAVITY_MODEL` (verified: the only references were in this PR's own code; `agy` has no `--model` flag), `model_used` holds the **provider** name post-step, and a pre-flight label can't be made truthful. So I took your lead "safe" option and **removed the Step 0 `Model` field** entirely rather than trying to seed it.

Mapping to your three points:
1. **`ANTIGRAVITY_MODEL` false label** → gone. No model is advertised at Step 0, so `PDD_AGENTIC_PROVIDER=antigravity ANTIGRAVITY_MODEL=foo` can no longer print `Model: foo`.
2. **`state["model_used"]` provider-as-model fallback** → gone. The `seed_startup_model` helper and its state fallback are fully reverted; `model_used`'s only remaining uses (the return tuple and post-step comments) are unchanged from `main`.
3. **settings.json coverage** → moot. We no longer claim `ANTIGRAVITY_MODEL`/`settings.json` as a model source.

Net diff vs `main`: the `- **Model**:` line removed from all four Step 0 banners (change/bug/test/e2e_fix) + the change prompt's Req-16 Model bullet replaced with a rationale note so a future `pdd sync` won't re-add it. Added guard tests asserting the banner omits the Model field. The authoritative per-step provider/model still lives in post-step `model_used` + the `.pdd/agentic-logs` audit record, as you said.

Orchestrator suites: 850 passed, 1 skipped. CI re-running on `3b21e3b27`.

Serhan-Asad:
Rebased onto latest `main` (`895f013f8`, conflict-free — **no functional change** since my last reply; the only delta is removing the Step 0 `Model` field per your lead option). Re-requesting review @gltanaka 🙏

gltanaka (approved):
This update addresses the prior blockers. Removing the Step 0 `Model` field is the right scope for this issue: it avoids advertising an unverified pre-flight model, avoids the `model_used` provider/model ambiguity, and no longer depends on unsupported Antigravity model env/settings behavior.

I checked the reduced diff and ran the focused guard tests for the four affected orchestrators locally; they pass. CI is green as well. This makes sense to merge.

Issue #1324, opened 2026-06-01
CLI: mid-run issue comment steering (orchestrators + run_agentic_task)


DianaTao:
Parent epic: #1323

## Summary

Implement **in-process** steering for agentic workflows in `promptdriven/pdd`: when `pdd_cloud` delivers pending issue comments (env/API/state), orchestrators drain them at safe boundaries and inject them into the next `run_agentic_task` call. This sub-issue does **not** cover webhook/job cancellation (see `pdd_cloud` sub-task on #1323).

**Locked behavior (from epic):**
- Human comments → steer at boundaries (not cancel).
- `/stop` and label removal → cancel (**cloud**); CLI must not treat organic comments as abort.
- Clarification pause → human reply **auto-continues** without re-label; merge steers into clarification step / `issue_content`.

---

### 1. Shared steer helper (`agentic_common.py`)

- [ ] Add `drain_issue_steers(owner, repo, issue_number, state, *, cwd) -> list[SteerEntry]` (name TBD).
- [ ] Read pending steers from cloud handoff (e.g. `PDD_STEER_JSON` env and/or poll `gh api` comments since `last_steered_comment_id`).
- [ ] Dedupe by `comment_id`; update `last_steered_comment_id` (and optional `steer_generation`) in GitHub workflow state via existing `github_save_state` / `save_workflow_state` paths.
- [ ] Filter bot/state/progress comments using same markers as `_fetch_issue_comments_text` (`PDD-INCREMENTAL-STATUS`, `PDD_WORKFLOW_STATE`, progress comment patterns).

### 2. Steer points (orchestrators + sync)

Call `drain_issue_steers` at boundaries before the next agent step:

- [ ] `agentic_change_orchestrator.py`
- [ ] `agentic_bug_orchestrator.py`
- [ ] `agentic_test_orchestrator.py`
- [ ] `agentic_e2e_fix_orchestrator.py`
- [ ] `agentic_checkup_orchestrator.py`
- [ ] `agentic_architecture_orchestrator.py` (if issue-driven)
- [ ] `agentic_split_orchestrator.py` (if issue-driven)
- [ ] `sync_orchestration.py` — between sync operations (GitHub App already uses `--no-steer`; this is **comment** steer, not TUI `maybe_steer_operation`)

**v1:** between orchestrator steps + before each `run_agentic_task` inside a step loop. Mid-subprocess steer deferred.

### 3. Prompt injection (`run_agentic_task`)

[...]

PR #1328, opened 2026-06-01
feat(agentic): add mid-run steer drain and injection

DianaTao:
## Summary
- Add `SteerEntry` and `drain_issue_steers(...)` to collect mid-run user steering from `PDD_STEER_JSON` and/or GitHub issue comments via `gh api`.
- Extend `run_agentic_task(..., steers=...)` to inject a `## Steered user input (mid-run)` block, while preserving existing `PDD_USER_FEEDBACK` behavior.
- Add unit tests for injection + drain + idempotency.

## Test plan
- `pytest -vv tests/test_agentic_common.py -k \"pdd_user_feedback_injected_into_prompt or steer_injection_into_prompt or drain_issue_steers\"`

Refs: #1324

Made with [Cursor](https://cursor.com)


gltanaka (changes requested):
This foundation is needed for #1324/#1336 if we want to merge the steering work incrementally, but I would not merge #1328 as-is.

Required changes:

1. Fix the GitHub polling command when `last_steer_at`/`since` is present. `drain_issue_steers()` appends `-f since=...` without forcing GET. GitHub CLI switches requests with field parameters to POST unless `--method GET` is supplied, so this path POSTs to `repos/{owner}/{repo}/issues/{issue}/comments`, gets a 422 because no `body` was supplied, and then silently returns no steers. I reproduced this against #1324: the current command exits 1 with `POST /repos/promptdriven/pdd/issues/1324/comments`, while the same command with `--method GET` exits 0 and sends `GET ...?since=...`.

2. Add regression coverage for the `last_steer_at` branch. The current tests cover env steers and GitHub polling without `since`, but they do not assert that the `since` request remains a GET query request.

3. Define/guard the initial cursor semantics for GitHub polling. With an empty state, the helper treats every existing human issue comment as a mid-run steer. If orchestration calls this before the first step or on a state without a cursor, stale issue discussion can be injected as new steering. Please seed `last_steered_comment_id`/`last_steer_at` at workflow start, pass an explicit run-start baseline, or otherwise document and test why draining all historical comments is correct.

Focused tests and the full `tests/test_agentic_common.py` pass, but they miss the failing `since` path above.

gltanaka (approved):
The updated head addresses my prior blockers: the GitHub comments list command now forces `--method GET`, there is coverage for the `last_steer_at`/`since` path, and GitHub polling no longer drains historical comments before a cursor is seeded.

This makes sense to merge as the 54.1 foundation after the remaining GitHub unit-test job completes green. Follow-up PR #1336 should rebase on this and wire `seed_issue_steer_cursor()` at workflow start; #1328 alone only adds the shared API and prompt injection.

PR #1344, opened 2026-06-02
feat: Add capability contracts and lightweight side-effect policy checks

### LINKED ISSUE #828: Safety: Add capability contracts and lightweight side-effect policy checks
## Problem

Natural-language contracts should describe not only return behavior, but also what a module is allowed to touch. Generated code can otherwise add “helpful” side effects such as external calls, logging, environment reads, or unrelated state mutations.

The prompting guide now documents `<capabilities>`, but PDD does not yet parse or check those sections against generated code.

## Proposal

Support capability sections such as:

```xml
<capabilities>
- MAY read payment records.
- MAY write refund records.
- MAY call the payment provider refund endpoint.
- MAY write audit events.
- MUST NOT send emails.
- MUST NOT modify customer profile records.
- MUST NOT read unrelated environment variables.
- MUST NOT log provider secrets, bearer tokens, card PAN, or CVV.
</capabilities>
```

Add:

```bash
pdd policy check prompts/refund_payment_python.prompt src/refund_payment.py
pdd policy check --json prompts/refund_payment_python.prompt src/refund_payment.py
```

Initial Python checks should be conservative and explainable:

- Forbidden imports.
- Forbidden network libraries.
- Forbidden file writes.
- Forbidden environment reads.
- Logging of variables named `token`, `secret`, `password`, `pan`, `cvv`, etc.
- Calls to out-of-scope modules/APIs.

## Scope

This is not full static taint analysis. V1 should prefer lightweight, deterministic checks with clear false-positive behavior and explicit waivers.

## Codebase pointers

- Prompt tag parsing can share code with contract checking.
- Python AST parsing should be preferred over regex where practical.
- CLI registration: `pdd/commands/__init__.py`

## Acceptance criteria

- `pdd policy check` supports Python initially.
- Capability sections are parsed and exposed to JSON output.
- Checks can be included in evidence manifests.
- Future `pdd gate` can require policy checks for critical modules.
- False positives can be waived with explicit waiver entries once the waiver model exists.
- Legacy prompts without `<capabilities>` do not fail unless strict mode is requested.

## Verification

- Add fixture prompts/code for allowed and forbidden imports/calls/env reads/logging.
- Run `pdd policy check --json` and assert expected warnings/errors.
- Confirm AST parser handles normal Python syntax and fails gracefully on invalid files.

DianaTao:
## Summary

- Adds **capability side-effect policy** under **`pdd checkup policy check`** (not a top-level `pdd policy` command), with AST-based Python scanning against `<capabilities>` contracts (forbidden imports, network/shell/file/env side effects, sensitive logging).
- Parses capability sections via shared contract IR and exposes them in `--json` output for CI and evidence manifests.
- Integrates capability policy into checkup review-loop gates, drift (when configured), and evidence manifest flows.

Closes #828

## CLI placement (important)

Capability policy is intentionally nested under **checkup**, matching **`pdd checkup gate`** (evidence YAML policy) and other local diagnostics:

| Command | Purpose |
|---------|---------|
| `pdd checkup policy check TARGET …` | **Capability / side-effect** checks from `<capabilities>` in prompts |
| `pdd checkup gate …` | **Evidence manifest** policy (`.pdd/policy.yml`, validation fields) |

There is **no** top-level `pdd policy` command. Review-loop deterministic gates invoke:

`python -m pdd.cli checkup policy check <file> --prompt <prompt>`

See [docs/policy_check.md](docs/policy_check.md) for usage, exit codes, and waivers.

## Test plan

- [x] `pdd checkup --validate-arch-includes` (architecture / `<pdd-dependency>` alignment)
- [x] `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py`
- [x] `pytest tests/test_drift_main.py -k policy`
- [ ] `make regression-public` (maintainer CI)

## Notes

Implements [promptdriven/pdd#828](https://github.com/promptdriven/pdd/issues/828), developed from [DianaTao/pdd#61](https://github.com/DianaTao/pdd/issues/61) / [DianaTao/pdd#68](https://github.com/DianaTao/pdd/pull/68).


DianaTao:
### CLI placement update

Capability policy is exposed only as **`pdd checkup policy check`** (same pattern as **`pdd checkup gate`** for evidence YAML policy). There is **no** top-level `pdd policy` command.

- **Capability checks:** `pdd checkup policy check <target> [--prompt …] [--json] [--strict] [--evidence]`
- **Evidence gate policy:** `pdd checkup gate …` (unchanged; different subsystem)

PR review-loop gates call `checkup policy check` when a changed `.py` file has `<capabilities>` in its prompt. Docs: `docs/policy_check.md`.

DianaTao:
## Merge conflict resolution (main → change/issue-61)

**Conflicted files:** `pdd/drift_main.py` only

**Resolution:** Kept PR capability checks (`policy_check` + `<capabilities>`) and upstream evidence gate checks (`gate_main` + `.pdd/policy.yml`). Drift runs either or both when configured; fails closed when a required backend is missing.

**CLI:** Capability policy remains `pdd checkup policy check` (not top-level `pdd policy`).

**Tests:**
- `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py -q` → **20 passed**
- `pytest tests/test_drift_main.py -k policy -q` → **2 passed**, 15 deselected
- `pytest tests/test_checkup_gates.py::TestDiscoverGates::test_emits_policy_gate_when_capabilities_present -q` → **1 passed**

**Touch point:** `pdd checkup policy check tests/fixtures/policy_check/forbidden_network.py --prompt tests/fixtures/policy_check/forbidden_network_python.prompt` → exit **1**, `[network]` violation (`Unauthorized network library import: requests`).

DianaTao:
## Branch cleanup (#828 / PR #1344)

- Confirmed work on `change/issue-61` only; capability policy is **`pdd checkup policy check`** (no top-level `pdd policy`).
- Drift: capability checks use `<capabilities>` / `validation.policy`; evidence gate uses `.pdd/policy.yml` / gate manifest keys (both paths preserved after main merge).
- Docs/README aligned with actual CLI options.

**Tests:**
- `pdd checkup --validate-arch-includes`: exit 0 — "No architecture / <include> mismatches found."
- `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py -q`: **20 passed**
- `pytest tests/test_drift_main.py -k policy -q`: **2 passed**
- `make regression-public`: exit 0 — "[public-regression] Public-safe CLI regression tests completed successfully"

**Touch point:** `pdd checkup policy check tests/fixtures/policy_check/forbidden_network.py --prompt tests/fixtures/policy_check/forbidden_network_python.prompt` → exit 1, `[network]` on `requests` import (line 3).

gltanaka (changes requested):
Decision: this feature is needed for #828, but I do not think this PR should merge as-is. The current checker can report PASS for ordinary side effects that the issue explicitly wants blocked, so it would give CI/review-loop users a false sense of safety.

1. `pdd/policy_check.py:78-90` and `pdd/policy_check.py:201-209` treat any non-`MUST NOT` capability containing `write` as authorization for filesystem writes. That means the issue-style/domain wording `MAY write refund records` allows `open(..., "w")` with no finding. Repro on this branch:

```python
prompt = """<capabilities>
- MAY write refund records.
- MUST NOT send emails.
</capabilities>
"""
code = """def refund():
    open('/tmp/pdd_policy_probe', 'w').write('x')
"""
# run_policy_check(...).passed == True, issues == []
```

Please separate domain writes from filesystem writes. File writes should require explicit file/disk/filesystem/storage wording, or another unambiguous filesystem capability, and this needs a regression test using the issue example style.

2. `pdd/policy_check.py:181-203` does not actually catch common `pathlib` writes even though `docs/policy_check.md:43-48` claims `Path.write_text` is supported. `from pathlib import Path; Path('/tmp/x').write_text('x')` passes with no issue when no file capability is present. Please resolve `Path` aliases / constructor calls and cover `Path.write_text`, `Path.write_bytes`, `Path.open(..., 'w')`, and destructive path operations with tests.

3. `pdd/policy_check.py:211-239` misses common environment reads through imported or aliased names. `from os import environ; environ['SECRET_TOKEN']` passes with no issue when no env capability is present. Aliased `import os as o; o.environ[...]` has the same shape. Please use `imported_names` consistently for attribute and subscript access and add regression coverage for these forms.

4. Resolve the command acceptance mismatch with #828. The issue acceptance asks for `pdd policy check`, but `tests/commands/test_policy_cli.py:22-27` asserts that top-level command does not exist. If the nested `pdd checkup policy check` placement is the intended product decision, the issue acceptance should be updated or this PR should add a backward-compatible top-level alias before closing #828.

5. Clean up diff hygiene. `git diff --check origin/main...origin/pr-1344` fails at `pdd/evidence_manifest.py:495` and throughout the regenerated `project_dependencies.csv` because of trailing whitespace / CRLF churn.

Verification I ran:

- `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py tests/test_drift_main.py -k policy` -> 22 passed
- `pytest tests/test_checkup_gates.py::TestDiscoverGates::test_emits_policy_gate_when_capabilities_present` -> passed
- The three probes above for domain-write filesystem access, `Path.write_text`, and `from os import environ` all incorrectly returned `passed=True` with no issues

gltanaka (changes requested):
Thanks for the update. The previous blockers are mostly addressed: `git diff --check` is clean, top-level `pdd policy check` is restored, and the prior repros for domain-write filesystem access, direct `Path(...).write_text`, and `from os import environ` now fail closed.

I still do not think this should merge yet because there is a remaining false PASS in the same side-effect surface: alias-bound `os` calls are not resolved before shell/env/file checks. `visit_Call` records imports, but the shell/file/env branches compare `module_name` directly to `"os"`, so normal Python aliasing bypasses enforcement.

Repro on current head `e14244130`:

```python
# prompt
<capabilities>
- MAY read records.
</capabilities>
```

```python
import os as o

def f():
    o.system("echo hi")      # shell side effect, currently passes
    o.getenv("SECRET_TOKEN") # env read, currently passes
    o.remove("/tmp/x")       # file side effect, currently passes
```

Observed via `run_policy_check`: all three variants returned `passed=True` with `issues=[]`.

Please resolve the imported binding before the shell/file/env call checks, e.g. use `_resolve_bound_module(module_name)` for `ast.Attribute` call handling, and add regression tests for `import os as o; o.system(...)`, `o.getenv(...)`, and `o.remove(...)` or equivalent file operation.

Verification I ran on the updated PR:

- `git diff --check origin/main...origin/pr-1344` -> clean
- `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py tests/test_drift_main.py tests/test_checkup_gates.py` -> 196 passed
- `python -m pdd.cli checkup --validate-arch-includes` -> pass
- GitHub checks are passing, including unit tests and public CLI regression
- Manual alias probes above still incorrectly pass

DianaTao:
## Fix: alias-bound `os` calls bypassing policy checks

Addresses the remaining false-PASS blocker: `import os as o` calls like `o.system()`, `o.getenv()`, and `o.remove()` were not enforced because `visit_Call` compared the local alias name (`o`) to `"os"` instead of resolving through import bindings.

### Changes
- **`pdd/policy_check.py`**: Added `_call_module_root()` and use `_resolve_bound_module()` for shell, env, and file `visit_Call` branches so aliased `os` / `shutil` / `subprocess` imports are checked the same as direct imports.
- **`tests/test_policy_check.py`**: Added `test_import_os_as_alias_call_methods_blocked` covering the reviewer's exact repro shape (`MAY read records` prompt + `o.system` / `o.getenv` / `o.remove`).

### Human-verifiable touch point
| | |
|---|---|
| **Test** | `test_import_os_as_alias_call_methods_blocked` |
| **File** | `tests/test_policy_check.py` |
| **Command** | `pytest tests/test_policy_check.py::test_import_os_as_alias_call_methods_blocked -vv` |
| **Expected** | 1 passed; each of shell/env/file variants fails policy with the matching category |
| **Proves** | Aliased `import os as o` cannot bypass shell, env, or filesystem side-effect enforcement |

### Validation
- `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py tests/test_drift_main.py -k policy` → **31 passed**
- Manual probes on head: `o.system`, `o.getenv`, `o.remove` all **FAIL** with `[shell]`, `[env]`, `[file]` respectively

Commit: `7c70bb6ac`

gltanaka (approved):
The previous blockers are resolved on the latest head.

Verified locally:
- `git diff --check origin/main...origin/pr-1344` -> clean
- Manual probes for `import os as o; o.system(...)`, `o.getenv(...)`, and `o.remove(...)` now fail closed with shell/env/file findings
- Extra probes for `subprocess` and `shutil` aliases also fail closed
- `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py tests/test_drift_main.py tests/test_checkup_gates.py` -> 214 passed
- `python -m pdd.cli checkup --validate-arch-includes` -> pass

GitHub checks are also passing, including unit tests and public CLI regression. This now makes sense to merge.

gltanaka (changes requested):
I am reopening this as a requested change for product usability, not because the previous static-analysis bugs are still present.

The current implementation is deterministic, which is good for CI, but the authoring model is drifting toward a hidden mini-syntax. Users should not have to learn or memorize an exact keyword taxonomy for `<capabilities>` before the checker behaves predictably. In particular, `docs/policy_check.md` documents a v1 keyword vocabulary, while `pdd/policy_check.py` has hard-coded category keyword lists and special filesystem phrases. That makes prompt authoring feel brittle: a user can write a reasonable natural-language capability and get surprising allow/deny behavior because they did not use the checker's exact words.

1. Keep `<capabilities>` as structured natural language, not a new syntax users must learn. The docs should lead with a small set of normal-language examples and state that wording is interpreted conservatively, rather than presenting the keyword table as something users need to memorize.

2. Add user-facing diagnostics when a capability section is present but the checker cannot map it clearly to supported categories. For example, if a side effect is denied because no explicit filesystem capability was recognized, the finding should explain the missing capability in plain language and suggest an acceptable phrase such as `MAY write audit files to disk`. Similarly, ambiguous capability bullets should be surfaced as warnings in JSON/human output rather than silently depending on keyword matching.

3. Add tests for the authoring UX, not just enforcement. We should verify that unclear/unsupported capability wording produces useful guidance, so users can fix the prompt without reading the implementation keyword lists.

The goal is to preserve deterministic enforcement while avoiding a new prompt sublanguage. Users should be able to write obvious capabilities in plain English and, when the checker cannot understand them, get actionable guidance rather than surprising behavior.

DianaTao:
## Re-review requested

This updates `change/issue-61` after the earlier review feedback and authoring-UX work.

### gltanaka blockers (addressed)

| Item | Status |
|------|--------|
| Domain `MAY write refund records` must not allow filesystem `open(..., "w")` | Fixed — explicit disk/file wording required |
| Pathlib `write_text` / `write_bytes` / `Path(...).open('w')` + aliases | Fixed + regression tests |
| `from os import environ` / `import os as o` env & call paths | Fixed (`_call_module_root` / binding resolution) |
| Top-level `pdd policy check` (#828 acceptance) | Restored + CLI tests |
| Diff hygiene | Addressed in prior commits |

### Capability authoring UX (`76719d7f6`)

- Unclear bullets (e.g. `MAY persist audit records`) → `capability_parser` **warnings** with suggestions; they do **not** grant permissions
- Denied effects → plain-language `missing_capability` messages + example capability phrases
- Surfaced in human CLI output and JSON (`capability_warnings`, `findings`)

### New fix (`e83ebb386`)

- **MUST NOT deny-wins over MAY** per effect category (order-independent). Repro that previously false-PASSed (`MUST NOT call external APIs` then `MAY call the Stripe refund API` + `import requests`) now fails closed.
- Regression: `test_must_not_overrides_later_may_same_category`, `test_must_not_overrides_earlier_may_same_category`
- Removed unused legacy `_capability_allows` / `_capability_allows_filesystem` helpers

### Branch sync

- Merged latest `origin/main` into `change/issue-61` (merge commit on branch).

### Follow-up issues (out of scope for #1344 / #828)

This PR stays **deterministic Python v1** only. Planned next steps:

- **#1370** — target-neutral contract/effect IR, `PolicyBackend` interface, `--target`, normalized `{action, resource}` findings, non-Python stub backend
- **#1371** — optional bounded **agentic** semantic review (`--semantic-review agentic`, default off) after deterministic checks for wrapper/indirection gaps

#1344 lays the enforcement + authoring UX foundation; #1370/#1371 extend architecture without replacing deterministic CI behavior.

### Touch points

```bash
pytest tests/test_policy_check.py tests/commands/test_policy_cli.py -q
# 36 passed (incl. deny-wins + authoring UX)

pdd policy check tests/fixtures/policy_check/forbidden_network.py \
  --prompt tests/fixtures/policy_check/forbidden_network_python.prompt
# exit 1, network finding
```

Please re-review when convenient.

gltanaka (changes requested):
This is close, and the new authoring warning path addresses the core usability direction, but I still do not think it should merge yet.

Two required changes remain:

1. The prompt sources are stale relative to the implemented behavior. This repo uses prompts as module contracts, so future sync/regeneration should not be based on obsolete requirements.

Examples on current head `e83ebb386`:

- `pdd/prompts/commands/policy_python.prompt` still says the command is invoked only via `pdd checkup policy`, says not to register top-level `pdd policy`, and says exit code `1` means warnings. The implementation now registers top-level `pdd policy`, keeps `pdd checkup policy` as an alias, and authoring warnings do not change the exit code.
- `pdd/prompts/policy_check_python.prompt` still defines only the old `PolicyIssue` / `PolicyResult` shape and does not describe `PolicyWarning`, `capability_warnings`, `kind="missing_capability"`, suggestions, unified findings, authoring warnings, or deny-wins behavior.
- `architecture.json` for `policy_check_python.prompt` still describes only `PolicyIssue`, `PolicyResult`, and `PolicyVisitor`, and should be aligned if the architecture metadata is expected to reflect exported contracts.

Please update the relevant prompts/architecture metadata so the prompt contracts describe the current feature and do not regenerate the pre-UX behavior.

2. The new deny-wins behavior has a misleading user-facing diagnostic. With:

```xml
<capabilities>
- MUST NOT call external APIs.
- MAY call the Stripe refund API.
</capabilities>
```

and code:

```python
import requests
```

`pdd policy check` correctly fails, but says:

```text
Generated code makes an external network or API call, but no capability allowing external calls was recognized.
Add a capability such as: MAY call the Stripe refund API.
```

That suggested capability is already present; the actual problem is that a `MUST NOT` bullet blocks the category. Please distinguish "missing recognized capability" from "blocked by a MUST NOT capability" and make the suggested remediation actionable, e.g. narrow/remove the contradictory `MUST NOT`, make the allowed API exception explicit in a supported way, or waive with justification. Add a regression test for the human/JSON diagnostic, not just the failure result.

Verification I ran on the updated PR:

- `git diff --check origin/main...origin/pr-1344` -> clean
- `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py tests/test_drift_main.py tests/test_checkup_gates.py` -> 222 passed
- `python -m pdd.cli checkup --validate-arch-includes` -> pass
- GitHub checks are passing
- Manual CLI probe confirms ambiguous capability warnings are visible and exit 0
- Manual CLI probe confirms the deny-wins diagnostic currently suggests adding a capability that is already present

DianaTao:
## Review fixes (`9d989b62a` on `change/issue-61`)

Addresses the two remaining review items from the latest pass.

### 1. Prompt / architecture contract alignment

Updated module contracts so sync/regeneration reflects current behavior:

- `pdd/prompts/commands/policy_python.prompt` — top-level `pdd policy` + `pdd checkup policy` alias, correct exit codes (authoring warnings alone → exit 0), JSON `capability_warnings` / `findings`, remediation wording by `issue.kind`
- `pdd/prompts/policy_check_python.prompt` — `PolicyWarning`, `capability_warnings`, `missing_capability` vs `blocked_by_must_not`, deny-wins, conservative ambiguity mapping
- `architecture.json` — matching interfaces/descriptions for `policy_check_python.prompt` and `commands/policy_python.prompt`

### 2. Deny-wins diagnostics (not misleading “add capability”)

When `MUST NOT` blocks a category that also has a `MAY` bullet, enforcement now emits `kind: blocked_by_must_not` with a message citing both bullets and deny-wins semantics. Remediation suggests narrowing/removing the `MUST NOT`, splitting modules, or a documented waiver — **not** re-adding the existing `MAY call the Stripe refund API` bullet.

### Human-verifiable touch points

| Test | File | Command | Expected | Proves |
|------|------|---------|----------|--------|
| `test_deny_wins_diagnostic_blocked_by_must_not_not_missing_capability` | `tests/test_policy_check.py` | `pytest tests/test_policy_check.py::test_deny_wins_diagnostic_blocked_by_must_not_not_missing_capability -vv` | 1 passed | Deny-wins failure uses `blocked_by_must_not`, cites MUST NOT + existing MAY, no “add Stripe API” suggestion |
| `test_deny_wins_json_and_human_diagnostic` | `tests/test_policy_check.py` | `pytest tests/test_policy_check.py::test_deny_wins_json_and_human_diagnostic -vv` | 1 passed | JSON `kind` + human `Suggested remediation:` for blocked cases |
| Policy suite (reviewer set) | — | `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py tests/test_drift_main.py tests/test_checkup_gates.py -q` | **224 passed** | No regressions across policy/drift/checkup |
| Arch includes | — | `python -m pdd.cli checkup --validate-arch-includes` | exit 0 | Prompt/architecture metadata in sync |

Manual probe:

```bash
pdd policy check /path/to/network.py --prompt /path/to/prompt-with-must-not-and-may.prompt
```

Expected: exit 1, `[network]` message mentions `MUST NOT` + existing `MAY`, `Suggested remediation: Narrow or remove...` (not “Add a capability such as: MAY call the Stripe refund API”).

Follow-ups **#1370** / **#1371** remain out of scope for this PR (linked in prior comment).

DianaTao:
## Follow-up review fixes (`35b9b5ed7`)

Addresses the remaining polish items from the latest review pass.

### Branch scope
- Reset away from the fork-only merge commit (`9d989b62a`) that pulled unrelated agentic/modify deltas not in the PR diff vs `main`.
- Merged current `origin/main` for sync; **`git diff origin/main...HEAD` remains 41 policy-focused files** (no agentic-only PR noise).

### Changes
| Item | Fix |
|------|-----|
| Test docstring | `tests/test_policy_check.py` documents both `pdd policy check` and `pdd checkup policy check` |
| MUST NOT–only diagnostic | `test_must_not_only_blocked_by_must_not_diagnostic` |
| Evidence manifest | `--evidence` writes `context_snapshot.policy_check.targets[].capability_warnings` (`test_policy_check_evidence_includes_capability_warnings`) |
| Prompt waivers | `_prompt_waiver_matches` accepts category + kind/effect + legacy phrases (no full message match required); `test_prompt_waiver_matches_category_without_exact_message` |

### Human-verifiable touch points

```bash
pytest tests/test_policy_check.py::test_must_not_only_blocked_by_must_not_diagnostic -vv
pytest tests/test_policy_check.py::test_prompt_waiver_matches_category_without_exact_message -vv
pytest tests/commands/test_policy_cli.py::test_policy_check_evidence_includes_capability_warnings -vv
pytest tests/test_policy_check.py tests/commands/test_policy_cli.py tests/test_drift_main.py tests/test_checkup_gates.py -q
```

**Expected:** all pass (227 tests in the reviewer suite on this branch).

**Manual deny-wins:** `pdd policy check` with MUST NOT + MAY network bullets → `Suggested remediation:` (not “Add a capability such as: MAY call the Stripe refund API”).

**Manual evidence:** `pdd policy check … --evidence` with ambiguous capability → manifest `context_snapshot.policy_check` includes `unmapped_capability` warnings.

gltanaka (changes requested):
Decision: this feature is still needed for #828, but I do not think this PR should merge as-is. The implementation now covers the earlier blockers, but two ordinary side-effect forms still false-PASS, so CI/review-loop users would get a misleading safety signal.

1. `pdd/policy_check.py:58-69` only treats a small set of module roots as network libraries, so stdlib `urllib.request` bypasses the network policy. Repro on current head: prompt has `<capabilities>\n- MAY read records.\n</capabilities>`, code has `from urllib.request import urlopen; urlopen("https://example.com").read()`. `run_policy_check(...).passed == True` and `issues == []`. Please detect at least the common stdlib URL/network entry points (`urllib.request` / `urlopen`, and preferably the adjacent stdlib HTTP client forms) and add regression tests for `from urllib.request import urlopen` and `import urllib.request; urllib.request.urlopen(...)`.

2. `pdd/policy_check.py:713-736` catches direct `Path(...).write_text(...)`, but not the normal variable-bound form. Repro: prompt has only `MAY read records`, code has `from pathlib import Path; p = Path("/tmp/pdd_probe"); p.write_text("x")`. This also returns `passed=True` with no issues. Please track simple names bound to `Path(...)` / `pathlib.Path(...)` constructors and enforce `write_text`, `write_bytes`, `open(..., "w")`, and destructive path methods on those names, with regression tests.

3. Documentation still has a command contradiction: `docs/ci.md:15` says there is no top-level `pdd policy` command, but this PR intentionally registers and documents `pdd policy check`. Please update that CI doc so it only scopes the “no top-level” statement to snapshot policy, or otherwise aligns with the final CLI decision.

Verification I ran:

- `conda run -n pdd pytest tests/test_policy_check.py tests/commands/test_policy_cli.py` -> 41 passed
- `conda run -n pdd pytest tests/test_drift_main.py -k policy tests/test_checkup_gates.py` -> 3 passed
- `git diff --check origin/main...HEAD` -> clean
- Manual probes above for `urllib.request.urlopen` and variable-bound `Path.write_text` both incorrectly returned `passed=True`, `issues=[]`

Once those false PASSes and the doc mismatch are fixed, I think the PR remains directionally appropriate to merge: it addresses a real gap between documented `<capabilities>` contracts and deterministic enforcement.

DianaTao:
## Review fixes: stdlib network + bound Path (`ba4b55cf8`)

Addresses the latest false-PASS blockers on `change/issue-61`.

### Changes

1. **Stdlib network (`pdd/policy_check.py`)**
   - Shared `_is_network_module()` covers third-party roots plus `urllib` / `urllib.*` and `http.client`
   - Imports and calls (`urlopen`, `urlretrieve`, `HTTPConnection`, etc.) fail when network is not allowed
   - Fixes reviewer repros: `from urllib.request import urlopen` and `import urllib.request; urllib.request.urlopen(...)`

2. **Variable-bound Path writes**
   - `visit_Assign` tracks names bound to `Path(...)` / `pathlib.Path(...)` constructors
   - `p.write_text`, `p.write_bytes`, and `p.open(..., "w")` are enforced like direct `Path(...).write_text`
   - Fixes: `p = Path("/tmp/pdd_probe"); p.write_text("x")` with only `MAY read records`

3. **`docs/ci.md`**
   - Clarifies there is no top-level **`pdd snapshot`** command (snapshot gate uses `pdd checkup snapshot`)
   - Points capability enforcement to **`pdd policy check`** / **`pdd checkup policy check`** (no contradiction with this PR)

### Human-verifiable touch points

| Test | File | Command | Expected |
|------|------|---------|----------|
| `test_urllib_request_urlopen_import_from_blocked` | `tests/test_policy_check.py` | `pytest tests/test_policy_check.py::test_urllib_request_urlopen_import_from_blocked -vv` | 1 passed; `network` finding |
| `test_urllib_request_module_urlopen_blocked` | `tests/test_policy_check.py` | `pytest tests/test_policy_check.py::test_urllib_request_module_urlopen_blocked -vv` | 1 passed; `network` finding |
| `test_path_variable_bound_write_text_blocked` | `tests/test_policy_check.py` | `pytest tests/test_policy_check.py::test_path_variable_bound_write_text_blocked -vv` | 1 passed; `file` finding |

**Suite:** `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py` → **44 passed**

**Manual repro (now fails closed):** prompt with `MAY read records` only + `urlopen("https://example.com")` or `p.write_text("x")` → `passed=False`.

gltanaka (changes requested):
Thanks for the update. The previous requested items are mostly resolved: stdlib `urllib.request.urlopen` now fails closed, simple `p = Path(...); p.write_text(...)` now fails closed, and `docs/ci.md` no longer contradicts the top-level `pdd policy check` command.

I still do not think this should merge yet because the new bound-Path tracking misses the annotated assignment form of the same ordinary pattern. `visit_Assign` at `pdd/policy_check.py:690` records `p = Path(...)`, but there is no corresponding `visit_AnnAssign`, so an annotated local bypasses filesystem enforcement.

Repro on current head `ba4b55cf8`:

```python
# prompt
<capabilities>
- MAY read records.
</capabilities>
```

```python
from pathlib import Path

def f() -> None:
    p: Path = Path("/tmp/pdd_probe")
    p.write_text("x")  # currently passes with no file issue
```

Observed via `run_policy_check`: `passed=True`, `issues=[]`.

Please handle simple `AnnAssign` bindings where the value is a `Path(...)` / `pathlib.Path(...)` constructor and add a regression test. This is the same side-effect surface as the fixed `p = Path(...)` case, just with a type annotation, so I would keep the PR blocked until it fails closed.

Verification I ran on the updated PR:

- Previous `urllib.request.urlopen` repro -> now fails with network findings
- Previous `p = Path(...); p.write_text(...)` repro -> now fails with a file finding
- `import pathlib; p = pathlib.Path(...); p.write_bytes(...)` -> fails with a file finding
- `p: Path = Path(...); p.write_text(...)` -> still false-PASSes
- `conda run -n pdd pytest tests/test_policy_check.py tests/commands/test_policy_cli.py` -> 44 passed
- `conda run -n pdd pytest tests/test_drift_main.py -k policy tests/test_checkup_gates.py` -> 3 passed
- `git diff --check origin/main...HEAD` -> clean

DianaTao:
## Fix: annotated Path assignment (`445d2aeca`)

Addresses the remaining false-PASS on bound-Path tracking.

### Change
- `visit_AnnAssign` registers names bound via `p: Path = Path(...)` / `pathlib.Path(...)` (shared helper with `visit_Assign`).
- Regression: `test_path_annassign_bound_write_text_blocked`

### Touch point
```bash
pytest tests/test_policy_check.py::test_path_annassign_bound_write_text_blocked -vv
```
**Expected:** 1 passed; `p: Path = Path(...); p.write_text("x")` with `MAY read records` only → `passed=False`, `[file]` finding.

**Manual repro:** `passed=False`, file issue present.

**Suite:** `pytest tests/test_policy_check.py tests/commands/test_policy_cli.py -q` → 45 passed

gltanaka:
Stepping back from the individual static-analysis misses: I think we should redesign this user experience before merging.

The core idea is valuable, but from a user perspective the current shape is getting hard to reason about. Users should not have to learn which exact words map to filesystem/network/env/email categories, or discover through trial and error that one natural Python spelling is checked while another equivalent spelling is not. That turns `<capabilities>` into a hidden mini-language plus a growing list of implementation edge cases.

What I would like to see instead is a simpler product design:

- Users write normal capability statements in plain English.
- PDD gives clear authoring feedback when a statement is ambiguous or unenforceable.
- Findings explain three things directly: what the code did, what the prompt allowed/forbade, and the smallest prompt/code change that resolves it.
- The deterministic checker remains underneath, but the user-facing workflow should make limitations explicit instead of making users infer the checker vocabulary from failures.
- We should decide whether this is primarily a prompt authoring/linting workflow, a CI enforcement workflow, or both, and make the command/output design reflect that.

I do not think we should keep incrementally adding AST cases until the user model is clearer. The feature should be easy for prompt authors to adopt and trust; right now it still feels too much like users need to understand the implementation to use it correctly.

DianaTao:
Redesigned in https://github.com/promptdriven/pdd/issues/1378, https://github.com/promptdriven/pdd/issues/1379, https://github.com/promptdriven/pdd/issues/1380, https://github.com/promptdriven/pdd/issues/1381

Issue #826, opened 2026-05-05
Reproducibility: Snapshot expanded prompts and dynamic context for replay


gltanaka:
## Problem

The prompting guide warns that `<shell>`, `<web>`, and `<include ... query="...">` introduce nondeterminism because shell output varies, web content changes, and semantic extraction can vary. The guide recommends snapshotting dynamic output before relying on it for durable contract behavior.

PDD needs a first-class workflow for capturing expanded prompt context and replaying/auditing it. This should extend the existing context-map work rather than duplicate it.

## Proposal

Add one or more of these commands/options:

```bash
pdd preprocess prompts/foo_python.prompt --snapshot
pdd generate prompts/foo_python.prompt --snapshot-context
pdd sync foo --snapshot-context
pdd replay .pdd/evidence/runs/<run_id>.json
```

Snapshot should capture:

- Fully expanded prompt text.
- Include inputs and hashes.
- Web content snapshots.
- Shell output snapshots.
- Semantic query extraction outputs.
- Selected grounding examples.
- Prompt/code/test hashes.

Replay should reconstruct the same expanded prompt context. It does not need to guarantee bit-for-bit identical generated code because LLM execution may remain nondeterministic; the promise is identical prompt/context reconstruction.

## Relationship to existing work

- Extends #318 (`Integrate context map sampling into code generation paths`). #318 captures generation context metrics/provenance; this issue adds durable snapshots and replay semantics for nondeterministic context.
- Should integrate with future evidence manifests by recording whether dynamic tags were used and where snapshots live.

## Acceptance criteria

- Evidence or context-map output records whether dynamic tags were used.
- Dynamic tag outputs can be snapshotted to stable files.
- A replay command can reconstruct the exact expanded prompt context from a run artifact.
- CI/policy layers can reject unsnapshotted nondeterministic context for critical modules.
- Snapshot files are documented, deterministic, and safe to inspect.
- Sensitive data handling/redaction is specified for shell/web snapshots.

## Verification

- Add fixture prompts containing `<include>`, `<shell>`, `<web>`, and `query=` includes.
- Snapshot them and assert the snapshot manifest contains the expected hashes and captured outputs.
- Replay from the run artifact and assert the reconstructed expanded prompt hash matches.
- Confirm static-only prompts report no nondeterministic context.

DianaTao:
Diana-June 2

gltanaka:
[...]

Issue #318, opened 2026-01-16
Integrate context map sampling into code generation paths


jamesdlevine:
# Context Map Implementation Plan for PDD

## Overview

Add context map capture to `pdd generate`/`pdd sync` to log detailed metrics about each generation for analysis and visualization.

**Approach**: This plan follows Prompt-Driven Development principles. All changes are made by modifying prompts, then regenerating code. Prompts are the source of truth; generated code is ephemeral.

### 1. `context_map_models_python.prompt` (NEW) ***STATUS: created***

**Purpose**: Pydantic models matching the JSON schema. Single source of truth for data structures.

**Location**: `pdd/prompts/context_map_models_python.prompt`

**Key Requirements**:
- Define models strictly matching `context_map.schema.json`  ***STATUS: created - context/context_map.schema.json***
- Export `ContextMap`, `Provenance`, `Input`, `Output`, nested models, and enums
- Provide `ContextMap.generate_sample()` for testing
- Convenience methods: `to_json()`, `from_file()`, `save()`

**Dependencies**: `context_map.schema.json` (via backtick include)

**Prompt-to-Code Ratio**: ~20% (schema drives structure)

---

### 2. `context_sampler_python.prompt` (NEW) ***STATUS: created***

**Purpose**: Storage layer for context map data during generation.

**Location**: `pdd/prompts/context_sampler_python.prompt`

**Key Requirements**:
- Accept `ContextMap` instances and persist to JSON
- Store in `.pdd_context/` directory alongside output
- File naming: `<basename>.context.<N>.json` with monotonic N
- Retention: keep most recent N files (default 5)
- CLI `--sample` flag outputs example via `ContextMap.generate_sample()`

**Dependencies**: `context_map_models` (import models)

**Contracts**:
- Input: ContextMap instance
- Output: path to written file, or None on error
- Invariant: file count per devunit ≤ configured maximum

---

### 3. `context_viz_python.prompt` (NEW) ***STATUS: created***

**Purpose**: CLI visualization of context map JSON files.

**Location**: `pdd/prompts/context_viz_python.prompt`

**Key Requirements**:
- Read JSON from file argument or stdin
- Two modes: summary (10x10 grid) and detailed (bar charts)
- Summary grid symbols: ▣ PDD system, ◆ devunit, █ includes, ▓ web, ⌘ shell, • variables, ◇ few-shot, ▤ prepend/append
- Show provenance (prompt file, model, timestamp, duration)

[...]

Issue #33, opened 2025-09-06
pdd sync 0.52.0 fails


rudi-cilibrasi:
using distribution llm_models.csv

```
pdd --force sync hello |& tee o.txt
```

[o.txt](https://github.com/user-attachments/files/22184918/o.txt)

gltanaka:
Thank you. Can you please attach everything in the .pdd directory as well?

rudi-cilibrasi-5:
Here is $HOME/.pdd

[pdd-directory.tar.gz](https://github.com/user-attachments/files/22185011/pdd-directory.tar.gz)

PR #1345, opened 2026-06-02
feat: snapshot expanded prompts and replay nondeterministic context (#826)

DianaTao:
## Summary

Implements reproducibility snapshots and replay for nondeterministic prompt context ([promptdriven/pdd#826](https://github.com/promptdriven/pdd/issues/826)).

- **`pdd preprocess --snapshot`** — capture expanded prompt text and dynamic tag outputs (shell, web, semantic `query=` includes) under `.pdd/evidence/runs/<run_id>/`
- **`pdd generate` / `pdd sync --snapshot-context`** — persist replayable context during generation/sync
- **`pdd replay <run-artifact>`** — verify and reconstruct expanded prompt context from a snapshot or evidence manifest
- **`pdd checkup snapshot <prompt>`** — policy gate that fails when active nondeterministic tags lack a replayable snapshot (aligned with other `pdd checkup` subcommands)
- **Evidence manifests** — record `uses_nondeterministic_tags`, snapshot artifact hashes, and `context_snapshot` linkage when `--evidence` is used with snapshotting
- **Redaction** — shell/web/query artifacts are scrubbed before persistence (tokens, bearer headers, secret assignments)

## Relationship

- Extends context-map / evidence provenance work; does not duplicate #318 metrics-only capture
- Fork tracking: DianaTao/pdd#62 / #33; upstream: #826

## Test plan

- [x] `pytest -q tests/test_context_snapshot_replay.py tests/test_context_snapshot_policy.py tests/test_evidence_manifest.py`
- [ ] Manual: `pdd preprocess prompts/foo_python.prompt --snapshot`
- [ ] Manual: `pdd generate … --snapshot-context --evidence` then `pdd replay .pdd/evidence/runs/<run_id>.json`
- [ ] Manual: `pdd checkup snapshot prompts/with_shell_python.prompt` (expect fail without prior snapshot)

Made with [Cursor](https://cursor.com)


gltanaka (changes requested):
This feature is needed for #826, but I would not merge this branch yet. Required changes:

1. `pdd checkup snapshot` must reject stale snapshots. `find_latest_snapshot_manifest()`/`check_snapshot_policy()` currently accept any replayable snapshot for the same prompt path. They do not compare the current prompt content/hash against the snapshot manifest prompt hash, so a prompt can be changed after snapshotting and the policy still passes using the old artifact. That defeats the CI/policy gate in #826. Please validate the current prompt against the snapshot manifest, and add a regression test where a nondeterministic prompt is snapshotted, edited, then `check_snapshot_policy()` fails until a fresh snapshot is created.

2. `preprocess --snapshot --recursive` still has unsafe behavior. The guard in `preprocess_main.py` checks only `detect_dynamic_tags(prompt)` on the raw parent prompt. That over-rejects documentation-only tags in code fences, but more importantly it misses dynamic tags introduced by included files. Repro: parent prompt contains only `<include>prompts/child_python.prompt</include>`, child contains `<shell>printf nested</shell>`, then `preprocess --snapshot --recursive` writes an expanded snapshot containing the unresolved `<shell>` tag while the manifest reports `uses_nondeterministic_context: false` and `dynamic_tags: []`. Please either disallow `--snapshot --recursive` entirely or make the guard inspect the fully expanded active prompt context and fail if any shell/web/query tags remain or were deferred. Add coverage for nested dynamic includes and documentation-only fenced tags.

The PR maps to the issue and the added targeted tests pass locally, but these two cases mean the snapshot/policy path can claim reproducibility without covering the current nondeterministic prompt context.

gltanaka (approved):
Re-reviewed the updated head. The two previous blockers are addressed: stale prompt snapshots are rejected by comparing the current prompt hash, and snapshot+recursive now fails when expansion leaves active nondeterministic tags, including nested dynamic includes while allowing fenced documentation examples.\n\nLocal checks run:\n- `pytest -q tests/test_context_snapshot_replay.py tests/test_context_snapshot_policy.py tests/test_evidence_manifest.py tests/test_issue_826_snapshot_touchpoint.py tests/test_preprocess_main.py tests/test_sync_main.py` -> 147 passed\n- stale-snapshot repro -> policy now fails with content changed\n- nested dynamic include CLI repro -> usage error before finalized snapshot\n- frontend `npm test -- --run tests/command-builder.test.ts` -> pass\n- frontend `npm run build` -> pass\n\nI also started `make test`, but stopped it because this local target enabled `PDD_RUN_REAL_LLM_TESTS=1` and blocked in a live `pdd generate` subprocess. GitHub CI is green for the PR. Approved.

PR #1352, opened 2026-06-02
feat(release): auto-create release video and post it to Discord

gltanaka:
## Summary

Makes the release video happen **automatically on every release** and surfaces it in the Discord **Updates** post — no manual `make release-video` step.

Today the video is created only by a local `make release-video` step, while the Discord post is a separate CI step in `release.yml` that posts **release notes only** (no video link) and runs decoupled from the local video. This PR closes that gap by moving video creation into the tag-triggered workflow and feeding the YouTube link into the existing Discord post.

## What it does

In the `publish-pypi` job, after the GitHub Release is created and notes are rewritten, a new step:
1. Installs the PDS CLI (from `vars.PDS_CLI_PACKAGE`) and `claude-code` (if needed).
2. Runs `make release-video` → renders the video and publishes it **unlisted to YouTube**.
3. Extracts the YouTube URL from the PDS response and **prepends it to the release notes**, so:
   - the release page shows it, and
   - the existing **"Post release notes to Discord"** step (which posts the release body) includes it — prepended so it survives that step's 4000-char tail-truncation.

No change to the Discord step itself was needed.

## Safety — best-effort, fully gated, absent-safe

- Runs **only** when `secrets.PDS_TOKEN` is present (`HAS_PDS` job env, same pattern as `HAS_CLAUDE_OAUTH`).
- **Skips cleanly** if `vars.PDS_CLI_PACKAGE` is unset, or if the pds/claude install fails.
- `continue-on-error: true` **+** internal `exit 0` on every failure path: PyPI publish and the GitHub Release already happened earlier in the job, so a video failure **can never fail a release**.
- **Until you provision the secret/var below, this is a no-op** — existing releases are unaffected.

## Activation (provision when ready)

| Type | Name | Purpose |
|---|---|---|
| secret | `PDS_TOKEN` | Release-authorized PDS agent token (scopes: `project:create,project:read,project:write,pipeline:run,artifact:read,distribution:package,distribution:publish`). A **project-locked** token (e.g. an e2e profile token) will not work. |
| var | `PDS_CLI_PACKAGE` | Install spec for the private `@promptdriven/pds` CLI — an npm spec once published, or an https/tarball URL. |
| var | `PDS_API_URL` | Optional PDS backend override. |
| var | `RELEASE_VIDEO_PROJECT_ID` | Optional stable PDS project to reuse across releases (otherwise a per-release `PDD <tag> release` project is created). |

> Note: `@promptdriven/pds` is currently `private: true` and not on npm, so CI needs an install source (`PDS_CLI_PACKAGE`). Options: publish it, stage a tarball (e.g. GCS, like the auto-heal pattern), or point at a release artifact.

## Test plan

- [x] `release.yml` parses as valid YAML; step order verified (video step sits between notes-rewrite and the Discord post).
- [x] Gating verified: with no `PDS_TOKEN` secret, `HAS_PDS=false` → step does not run (no behavior change to current releases).
- [ ] End-to-end (after provisioning `PDS_TOKEN` + `PDS_CLI_PACKAGE`): tag a release, confirm the video publishes and the YouTube link appears in the GitHub release notes + Discord post.

## Out of scope

- Provisioning the `PDS_TOKEN` / `PDS_CLI_PACKAGE` (requires a release-authorized PDS credential + deciding the CLI install source).
- Whether every release (incl. patch bumps) should publish a video vs. only minor/major — currently every release; easy to gate later on `github.ref_name`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -89,8 +89,8 @@ jobs:
             gh release create "$tag" --generate-notes --verify-tag --target main
           fi
 
-      - name: Setup Node for Claude Code / PDS
-        if: env.HAS_CLAUDE_OAUTH == 'true' || env.HAS_PDS == 'true'
+      - name: Setup Node for Claude Code
+        if: env.HAS_CLAUDE_OAUTH == 'true'
         uses: actions/setup-node@v6
         with:
           node-version: "20"
@@ -226,6 +226,12 @@ jobs:
             echo "vars.PDS_CLI_PACKAGE not set; skipping release video."
             exit 0
           fi
+          # Uses the runner's preinstalled Node (ubuntu-latest ships Node 20+);
+          # if it is ever absent, skip rather than fail the release.
+          if ! command -v npm >/dev/null 2>&1; then
+            echo "::warning::node/npm not available on runner; skipping release video."
+            exit 0
+          fi
           if ! npm install -g "$PDS_CLI_PACKAGE"; then
             echo "::warning::pds CLI install from PDS_CLI_PACKAGE failed; skipping release video."
             exit 0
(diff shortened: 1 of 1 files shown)


gltanaka:
### Codex round 1: REQUEST_CHANGES → addressed in 3cfbd90e7

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | major | Expanding the non-best-effort `setup-node` to `HAS_PDS` could fail the release post-publish if setup-node failed | Reverted — `setup-node` is Claude-only again. The video step uses the runner's preinstalled Node (ubuntu-latest ships Node 20+) and skips with a warning if `npm` is absent. No non-best-effort step runs post-publish for the video. |
| 2 | minor | `gh release view` / `gh release edit` unguarded under `set -uo pipefail` (no `-e`) | Guarded both: on failure they warn and `exit 0` rather than proceeding with an empty body or a false success. |

YAML re-validated; step order unchanged (video step best-effort, `continue-on-error: true`, between note-rewrite and Discord).

Issue #1364, opened 2026-06-02
GitHub App pdd sync regressed when pdd_cloud default moved to vertex_ai/gemini-3.5-flash before pdd main support landed


Serhan-Asad:
## Problem

The GitHub App `pdd sync` path used to work, then regressed.

The regression happened after the GitHub App side changed the inner `llm_invoke` default model to `vertex_ai/gemini-3.5-flash`, while `pdd` main still did not ship that model in the bundled catalog and still did not have the safer provider-aware fallback behavior on `main`.

As a result, default GitHub App sync runs started asking `llm_invoke` to use a model that `pdd` main did not know how to resolve directly.

## What changed

- In `pdd_cloud`, commit `428593b19` made the executor/orchestrator always pass `PDD_MODEL_DEFAULT` into the inner `llm_invoke` path.
- In `pdd_cloud`, commit `6e08eb488` changed `LLM_INVOKE_DEFAULT_MODEL` from `vertex_ai/gemini-3-flash-preview` to `vertex_ai/gemini-3.5-flash`.
- In `pdd`, `main` still ships `vertex_ai/gemini-3-flash-preview` in the bundled `pdd/data/llm_model.csv`, but not `vertex_ai/gemini-3.5-flash`.
- The branch that adds `vertex_ai/gemini-3.5-flash` to the bundled catalog (`#1136` / commit `62ae7aa7f`) did not land on `main`.
- The branch that hardens fallback to stay provider-safe by default (`#1115`) also did not land on `main`.

## Why it used to work

It used to work because the GitHub App default was `vertex_ai/gemini-3-flash-preview`, and that exact model already existed in `pdd` main.

So the GitHub App and `pdd` were aligned before the default changed.

## Root cause

This is a cross-repo drift bug:

1. `pdd_cloud` advanced the GitHub App inner default model.
2. `pdd` main did not advance the bundled catalog in sync.
3. `pdd` main also lacked the fallback hardening needed when the configured default model is newer than the bundled catalog.

So the break was not just “Gemini 3.5 is bad” or “Vertex is flaky”. The break was that the app started sending a new default model id into an older `pdd` runtime contract.

## Proposed fix

Fix this as one tracked regression:

1. Add `vertex_ai/gemini-3.5-flash` to the bundled `pdd/data/llm_model.csv` and preserve it in catalog generation.
2. Restore the provider-safe fallback behavior so a `vertex_ai/...` default tries same-provider usable models first before widening.
[...]

Issue #1363, opened 2026-06-02
llm_invoke should resolve cloud-preserved PDD_LLM_INVOKE credential aliases


Serhan-Asad:
## Problem

Cloud `pdd sync` can preserve direct provider API keys for the inner `llm_invoke` path using dedicated env aliases such as `PDD_LLM_INVOKE_ANTHROPIC_API_KEY`, while the outer CLI environment continues to scrub the real provider env names such as `ANTHROPIC_API_KEY`.

Public PDD currently needs to understand those aliases during direct `llm_invoke` auth checks and LiteLLM invocation. Without that, pdd_cloud can preserve the key correctly but the installed PDD CLI still behaves as if the key is missing.

This issue is no longer about prefiltering or reranking model candidates by credential availability. The root cloud credential-preservation fix lives in pdd_cloud; this public PDD issue is the supporting alias-resolution plumbing.

## Desired behavior

For single-key providers, `llm_invoke` should:

- treat `PDD_LLM_INVOKE_<KEY>` as satisfying `<KEY>` only when `llm_invoke` explicitly asks for runtime aliases
- pass the resolved alias value into the LiteLLM call as `api_key=`
- keep generic provider setup/key enumeration unchanged, so runtime-only aliases do not leak into normal setup flows
- skip interactive prompting in force/cloud contexts instead of calling `input()`
- continue honoring existing Gemini/Google aliases and Vertex project/location aliases

## Non-goals

- Do not remove Anthropic or other providers from `llm_model.csv`; the CSV is a global model catalog.
- Do not prefilter or rerank `_select_model_candidates()` by current-process credentials.
- Do not treat Claude Code OAuth as satisfying direct Anthropic API-key auth for LiteLLM.

## Acceptance criteria

- `resolve_api_key_from_env("ANTHROPIC_API_KEY", include_llm_invoke_aliases=True)` can resolve `PDD_LLM_INVOKE_ANTHROPIC_API_KEY`.
- The same alias is ignored when `include_llm_invoke_aliases` is false.
- `_ensure_api_key()` accepts `PDD_LLM_INVOKE_ANTHROPIC_API_KEY` when the CSV row requires `ANTHROPIC_API_KEY`.
- The LiteLLM invocation path passes the resolved alias value as `api_key=` for single-key providers.
- Force/cloud mode does not attempt an interactive API-key prompt.
- Model candidate selection remains catalog/ranking based, with no credential prefiltering.

## Related

[...]

PR #1138, opened 2026-05-22
feat(catalog): add vertex_ai/gemini-3.5-flash to bundled model catalog

Serhan-Asad:
## Summary

- Adds `Google Vertex AI,vertex_ai/gemini-3.5-flash` to `pdd/data/llm_model.csv`. Pricing `$1.50/$9.00` matches Google's published Vertex rate and the pdd_cloud worker catalog (commit `6e08eb488b`). ELO 1442 matches pdd_cloud. Preview row (`vertex_ai/gemini-3-flash-preview`) is kept for backward compat.
- Adds `vertex_ai/gemini-3.5-flash` to `_MANDATORY_MODEL_ROWS` in `pdd/generate_model_catalog.py` so catalog regeneration preserves it even if litellm dedup removes it; adds `"gemini-3.5-flash": 1442` to `STATIC_ELO_FALLBACK` to pass the `ELO_CUTOFF` guard in the mandatory-row seeder.
- Four new tests (2 catalog, 2 llm_invoke) pin the fix; 344 existing tests all pass.

## Why

`pdd_cloud` runs with `LLM_INVOKE_DEFAULT_MODEL=vertex_ai/gemini-3.5-flash` (pdd_cloud `origin/main`). Without this row, a local `PDD_MODEL_DEFAULT=vertex_ai/gemini-3.5-flash` run falls through to the surrogate-base path. Under the provider lock from PR #1115, the surrogate becomes `vertex_ai/claude-opus-4-7` (first Google Vertex AI row) — wrong model, wrong pricing, wrong ELO for strength interpolation, wrong temperature-clamp behavior.

## Test plan

- [x] `test_committed_csv_includes_vertex_gemini_3_5_flash_ga_default` — CSV substring assertion
- [x] `test_build_rows_includes_vertex_gemini_3_5_flash_ga_default` — seeded via `_MANDATORY_MODEL_ROWS` with fake litellm that lacks the row
- [x] `TestSelectModelCandidates::test_issue_1136_vertex_gemini_3_5_flash_resolves_directly` — resolves to GA row, not `vertex_ai/claude-opus-4-7` surrogate trap
- [x] `TestSelectModelCandidates::test_issue_1136_gemini_3_5_flash_in_gemini_3_family_clamp` — temperature clamp fires for GA id
- [x] `pytest tests/test_generate_model_catalog.py tests/test_llm_invoke.py tests/test_llm_invoke_csv_model_registration.py tests/test_update_model_costs.py` — 344 passed

Closes #1136.

[...]

Issue #1136, opened 2026-05-22
Add Gemini 3.5 Flash to the bundled LLM model catalog


Serhan-Asad:
## Summary

Add `vertex_ai/gemini-3.5-flash` to the bundled PDD model catalog (`data/llm_model.csv`).

## Context

`pdd_cloud` now uses `LLM_INVOKE_DEFAULT_MODEL=vertex_ai/gemini-3.5-flash`, and its GitHub App worker CSV contains that row. The public/local `pdd` repo CSV still only has `vertex_ai/gemini-3-flash-preview` and related preview rows.

That means local/public PDD runs with `PDD_MODEL_DEFAULT=vertex_ai/gemini-3.5-flash` cannot find the exact base model in the bundled catalog, so they rely on surrogate fallback instead of starting from the intended model.

## Proposed work

- Add a `Google Vertex AI` row for `vertex_ai/gemini-3.5-flash` to `data/llm_model.csv`.
- Use pricing/ELO/location values consistent with the current `pdd_cloud` worker catalog unless there is a newer source of truth.
- Consider whether matching non-Vertex rows are also needed, for example `gemini/gemini-3.5-flash`, if supported by the current LiteLLM/Gemini API path.
- Update any generated catalog metadata or tests that assert bundled model rows.

## Relationship to #1115

This ticket fixes the stale catalog issue: the intended default model should exist in the bundled CSV.

PR #1115 is still separate and still needed: it fixes fallback behavior so a provider-prefixed default like `vertex_ai/...` stays inside Vertex/Google by default instead of accidentally selecting Anthropic/Fireworks candidates that may not have keys.

## Acceptance criteria

- `data/llm_model.csv` includes `vertex_ai/gemini-3.5-flash`.
- A local `PDD_MODEL_DEFAULT=vertex_ai/gemini-3.5-flash` lookup finds the exact model row.
- Existing `llm_invoke` model-selection tests pass.
- The change does not remove the existing `vertex_ai/gemini-3-flash-preview` row unless there is a deliberate compatibility decision to do so.

Serhan-Asad:
## Step 1/8: Project Discovery

### Tech Stack
- **Languages:** Python 3.12+, TypeScript/TSX, JavaScript, shell, YAML/CSV/Markdown; examples also include Java/Maven/Gradle projects.
[...]

PR #1367, opened 2026-06-03
fix(catalog): add vertex_ai/gemini-3.5-flash cloud default (#1364)

Serhan-Asad:
## Summary

Scopes this PR down to the minimum pdd-side fix for the cloud default drift: add the exact `vertex_ai/gemini-3.5-flash` row to the bundled model catalog and preserve it during catalog regeneration.

This intentionally does **not** change `_select_model_candidates`, surrogate-base behavior, provider ordering, or `--strength` semantics. With the exact row present, `PDD_MODEL_DEFAULT=vertex_ai/gemini-3.5-flash` resolves normally and the existing strength selector keeps working from the configured base model.

## What changed

- Added `Google Vertex AI,vertex_ai/gemini-3.5-flash` to `pdd/data/llm_model.csv`.
- Added the row to `_MANDATORY_MODEL_ROWS` and `STATIC_ELO_FALLBACK` so regeneration keeps it even when `litellm.model_cost` lacks the id.
- Updated the catalog generator prompt to document the mandatory-row mechanism.
- Added tests that the row is generated, committed, present for the current cloud default, resolves directly through the existing selector, and is included in the Gemini 3 family classifier.

## What was removed from the PR

The earlier provider-safe surrogate change was removed to keep this focused and avoid changing selector behavior. Missing-row fallback policy can be handled separately if still needed.

## Validation

```bash
conda run -n pdd python -m pytest -q \
  tests/test_generate_model_catalog.py::test_build_rows_includes_vertex_gemini_3_5_flash_ga_default \
  tests/test_generate_model_catalog.py::test_committed_csv_includes_vertex_gemini_3_5_flash_ga_default \
  tests/test_generate_model_catalog.py::test_prod_cloud_default_models_present_in_catalog \
  tests/test_llm_invoke.py::TestSelectModelCandidates::test_vertex_gemini_3_5_flash_resolves_directly_from_packaged_catalog \
  tests/test_llm_invoke.py::TestGemini3TemperatureClamp::test_is_gemini_3_model_classifier
```

Result: `5 passed, 1 warning`.

Part of #1364. Related: #1136 / #1138, #1363.


Serhan-Asad:
Thanks — careful review. Triaged all three; one was a real (pre-existing) gap and is now fixed, the other two are addressed by clarification.

**1. Generator prompt drift — FIXED (`89bc8e35d`).** Correct catch. `generate_model_catalog.py` is mapped to `generate_model_catalog_python.prompt` in `architecture.json`, yet the prompt documented only the `_CHATGPT_SUBSCRIPTION_ROWS` and local-runner survival mechanisms — `_MANDATORY_MODEL_ROWS` was absent entirely (so a regen would also have dropped the *existing* `vertex_ai/gemini-3-flash-preview` / Opus relay rows, not just mine — a pre-existing gap this row surfaced). The prompt now documents the `_MANDATORY_MODEL_ROWS` mechanism (mirroring the `_CHATGPT_SUBSCRIPTION_ROWS` bullet), naming the GA Vertex Gemini Flash default and the ELO/`STATIC_ELO_FALLBACK`/`ELO_CUTOFF` seeding path, so prompt and code agree.

**2. Static (not live) cross-repo guard — acknowledged; reworded the claim.** Agreed this doesn't fully satisfy the cross-repo acceptance criterion, and I'd already flagged it as a static pin. A *live* guard is fundamentally impossible from the public `pdd` side: `pdd` cannot read `pdd_cloud`'s runtime `LLM_INVOKE_DEFAULT_MODEL` constant — that check has to live in `pdd_cloud`'s CI. I've changed the close keyword from `Closes #1364` → **`Part of #1364`** so merging does **not** auto-close the umbrella while that pdd_cloud follow-up is outstanding. The pdd-side production regression is fully fixed.

**3. Stale user/project CSV shadowing — working as intended, and Part 2 already mitigates it.** If a user ships their own `~/.pdd/llm_model.csv` that omits the new row, that omission is the user's explicit override — PDD deliberately honors the user CSV over the bundled catalog, so forcing the bundled row would break that feature. This isn't specific to this model (any newly-bundled model is absent from a stale user CSV by design). Crucially, the failing production path (GitHub App / Cloud Run) uses the **bundled** catalog → Part 1 fixes it directly; and for the stale-user-CSV case, Part 2 is exactly the mitigation (same-provider Vertex surrogate, not Bedrock) — which you noted is "better than Bedrock drift." So no further code change; documented in the PR body.

**Other:** synced with `origin/main` (was 2 behind; clean merge). The local pytest segfault you hit is the known `LC_ALL=C.UTF-8` capture crash — prefixing with `LC_ALL=C` avoids it. CI here is green (Run Unit Tests + Public CLI Regression + CodeQL all pass).

PR #1402, opened 2026-06-03
fix(llm_invoke): resolve cloud credential aliases for inner sync

### LINKED ISSUE #1363: llm_invoke should resolve cloud-preserved PDD_LLM_INVOKE credential aliases

Serhan-Asad:
## Summary

Narrow this PR to the public PDD plumbing needed by the cloud root-cause fix: inner `llm_invoke` can resolve dedicated `PDD_LLM_INVOKE_*` credential aliases that pdd_cloud preserves for `pdd sync`.

This PR no longer prefilters or reranks the model candidate pool by current-process credentials. Model selection stays catalog/ranking based; auth checks happen when a candidate is attempted.

Closes #1363.

## What changed

- `provider_manager.resolve_api_key_from_env(..., include_llm_invoke_aliases=True)` can resolve aliases such as `PDD_LLM_INVOKE_ANTHROPIC_API_KEY`.
- The alias mechanism is generic, so cloud can use the same `PDD_LLM_INVOKE_<KEY>` shape for other single-key providers.
- `llm_invoke._ensure_api_key()` uses that alias-aware resolver for single-key providers.
- The actual LiteLLM call path passes the resolved alias value as `api_key=`, instead of rereading only the literal CSV env name.
- Force/cloud mode skips interactive key prompts instead of calling `input()`.
- Vertex project/location handling still honors `VERTEXAI_*`, legacy `VERTEX_*`, and `GOOGLE_CLOUD_PROJECT`.

## Why

The root cloud issue is not that Anthropic is absent from the catalog or that selection must remove it. The root issue is that pdd_cloud may scrub the real `ANTHROPIC_API_KEY` from the outer CLI env while preserving it for inner sync as `PDD_LLM_INVOKE_ANTHROPIC_API_KEY`. Public PDD must understand that alias for the cloud-side fix to work.

This pairs with promptdriven/pdd_cloud#1914 / promptdriven/pdd_cloud#1916.

## Non-goals

- Do not remove Anthropic or other providers from `llm_model.csv`; the CSV is a global model catalog.
- Do not prefilter or rerank `_select_model_candidates()` by current-process credentials.
- Do not treat Claude Code OAuth as satisfying direct Anthropic API-key auth for LiteLLM.

## Verification

- `conda run -n pdd python -m pytest -q tests/test_llm_invoke.py tests/test_provider_manager.py tests/test_codex_subscription.py` -> 434 passed before merge
- `git diff --check`
- Direct `llm_invoke` regression covers Anthropic cloud alias auth reaching LiteLLM as `api_key=`.


gltanaka (changes requested):
I think the public PDD alias plumbing is needed for the pdd_cloud fix: pdd_cloud#1916 preserves provider secrets as `PDD_LLM_INVOKE_*` while continuing to scrub the real provider env vars, so the installed PDD CLI must understand those aliases for inner `llm_invoke` to work.

Do not merge this revision yet. Required changes:

1. Make the Vertex tests hermetic with the newly supported legacy aliases. Locally, this command fails:

   `python -m pytest tests/test_provider_manager.py tests/test_llm_invoke.py::TestEnsureApiKey tests/test_codex_subscription.py -q`

   Failures:
   - `tests/test_llm_invoke.py::TestEnsureApiKey::test_vertex_pipe_delimited_no_creds_no_project`
   - `tests/test_llm_invoke.py::TestEnsureApiKey::test_vertex_pipe_delimited_no_csv_location_fails`

   The branch now honors `VERTEX_PROJECT` / `VERTEX_LOCATION` / `GOOGLE_CLOUD_PROJECT`, but those tests do not consistently clear the new legacy aliases before asserting “no project/no location”. Please either add these aliases to the autouse provider-env isolation or explicitly `delenv` them in the tests that depend on their absence.

2. Add a direct `llm_invoke()` call-path regression for the cloud alias itself. Current coverage proves `PDD_LLM_INVOKE_ANTHROPIC_API_KEY` satisfies `_ensure_api_key()`, and separately proves the existing Google/Gemini alias is passed to LiteLLM. It does not prove the cloud-specific alias value is passed as `api_key=` when `ANTHROPIC_API_KEY` is absent, which is the key behavior pdd_cloud depends on.

3. Minor cleanup: remove the unused `resolved_key_name` assignment in the LiteLLM kwargs path.

After those are fixed, the feature looks justified and scoped enough to merge.

Serhan-Asad:
Addressed Greg's review in 152056f6a:\n\n1. Made the Vertex absence tests explicitly clear VERTEX_PROJECT and VERTEX_LOCATION, so legacy aliases cannot leak into no-project/no-location assertions.\n2. Added a direct llm_invoke() regression proving PDD_LLM_INVOKE_ANTHROPIC_API_KEY is passed to LiteLLM as api_key when ANTHROPIC_API_KEY is absent.\n3. Removed the unused resolved_key_name assignment in the LiteLLM kwargs path.\n\nVerification:\n- conda run -n pdd python -m pytest -q tests/test_llm_invoke.py tests/test_provider_manager.py tests/test_codex_subscription.py -> 434 passed\n- GitHub checks are all passing.

gltanaka (approved):
The requested fixes are in place: the Vertex env-absence tests now clear the legacy aliases, the cloud-specific Anthropic alias is covered through the actual llm_invoke -> LiteLLM api_key path, and the unused resolved_key_name assignment is gone. I re-ran the prior failing focused suite and the new/nearby regressions locally; they pass. This now makes sense to merge.

PR #1407, opened 2026-06-04
Fix DeepSWE catalog fallback retention

### LINKED ISSUE #1405: Restore Arena fallback rows in DeepSWE-ranked model catalog
## Problem

`pdd-cli v0.0.262` dropped models that were present in the Arena-backed catalog but absent from the DeepSWE manifest, including `gemini-3.1-pro-preview`. Public auto-heal in `pdd_cloud` still configures `PDD_MODEL_DEFAULT=gemini-3.1-pro-preview`, so its safety gate now sees that base missing from `llm_model.csv`, falls back to the highest-ranked model (`anthropic.claude-opus-4-7`), and fails instead of running on Gemini.

## Expected

The model catalog should use DeepSWE as the primary ranking source where available, then keep Arena/static catalog rows as fallback for models without DeepSWE coverage. DeepSWE ranking should not delete supported models just because they lack a DeepSWE score.

## Evidence

Regression commit: `dedea3d92` (`fix: rank models by deepswe first`).

Removed rows include:

```text
gemini-3.1-pro-preview
gemini-3.1-pro-preview-customtools
gemini/gemini-3.1-pro-preview
gemini/gemini-3.1-pro-preview-customtools
openrouter/google/gemini-3.1-pro-preview
openrouter/google/gemini-3-flash-preview
azure/gpt-5
azure/o3
azure/o4-mini
gpt-5
o3
o4-mini
```

Cloud Build failure from PR #1360:

```text
pdd-cli 0.0.262 will not route auto-heal to a Gemini model: at strength=0.5 with base='gemini-3.1-pro-preview', the first candidate is 'anthropic.claude-opus-4-7'. The configured base may have been renamed/removed from llm_model.csv, causing fallback to the highest-ELO model.
```

## Fix direction

Update catalog generation so DeepSWE score/rank fields override ranking when present, but rows not covered by DeepSWE remain in `pdd/data/llm_model.csv` with Arena/static fallback ranking metadata. Add regression coverage that a known non-DeepSWE fallback model such as `gemini-3.1-pro-preview` survives regeneration.

Serhan-Asad:
## Summary
- keep DeepSWE-ranked rows as the primary reviewed catalog source without letting them delete supported fallback routes
- preserve Arena-reviewed rows and static fallback rows during provider-local Pareto pruning
- regenerate llm_model.csv so removed routable models like gemini-3.1-pro-preview are present again

Closes #1405.

## Why
pdd-cli v0.0.262 removed fallback rows when DeepSWE-ranked models dominated them by rank/cost. That made configured models such as gemini-3.1-pro-preview disappear from the bundled CSV, so model selection missed the configured Gemini base and fell through to a Claude candidate.

## Test plan
- uv run pytest tests/test_generate_model_catalog.py -q
- python -m py_compile pdd/generate_model_catalog.py tests/test_generate_model_catalog.py
- smoke checked _select_model_candidates(0.5, "gemini-3.1-pro-preview", bundled CSV) returns gemini-3.1-pro-preview


Serhan-Asad:
Addressed the review items in the latest push:

- Updated `pdd/prompts/generate_model_catalog_python.prompt` so future prompt-driven regeneration preserves DeepSWE/Arena reviewed rows and static fallback routes.
- Added README troubleshooting for stale `~/.pdd/llm_model.csv` / project `.pdd/llm_model.csv` overrides shadowing the packaged catalog after upgrade.
- Reran `uv run pytest tests/test_generate_model_catalog.py -q` and `python -m py_compile pdd/generate_model_catalog.py tests/test_generate_model_catalog.py`.

Serhan-Asad:
The remaining red checks (`heal` / custom `auto-heal`) are the expected bootstrap failure for this PR. The workflow installs the latest released PyPI package (`pdd-cli 0.0.262`) before this fix exists in a release, so the guard still reports `gemini-3.1-pro-preview` falling through to Claude. Unit tests, package smoke, CLI regression, and CodeQL are green.

Path forward: merge this with maintainer/admin bypass for the self-referential auto-heal check, publish a new `pdd-cli` release, then rerun auto-heal. After the release, the same guard should pass because the packaged CSV contains the restored Gemini rows.

PR #1401, opened 2026-06-03
docs(research): repo-bloat localization benchmark — pilot design (#1209)

sohni-tagirisa:
## Summary

Opens the research workspace for issue #1209 — *measure agentic localization degradation under repository bloat* — with the **pilot design document** as the first deliverable.

Research question: given the same task, model, verifier, and starting code, does an agentic code-patching workflow read more, search more, spend more tokens, edit the wrong files, or fail hidden tests as the surrounding repo grows with plausible-but-irrelevant files?

Closes part of #1209 (design deliverable; scenarios/harness/runs land in follow-ups).

## What's in this PR

- `research/repo-bloat-benchmark/design.md` — the design document, covering:
  1. **Benchmark architecture** — fixtures → distractor pool → manifest builder → variant builder → run harness → analysis, with determinism/isolation guarantees.
  2. **Scenario format** — `scenario.json` descriptor, task brief, visible-vs-hidden split, isolated hidden verifier.
  3. **Distractor generation strategy** — LOC-budget sizing, nearness tiers, deterministic rename/templating, plus a content-addressed manifest schema + lockfile.
  4. **Instrumentation plan** — first-edit boundary, three independent taps (FS reads, tool-call/token transcript, edit diff), run-record schema, failure taxonomy.
  5. **Reporting format** — per-size tables, slope fits, plots, methodology note, and a pre-registered supports/weakens/inconclusive verdict.
- `research/repo-bloat-benchmark/README.md` — branch orientation, planned layout, and the locked-decision table.

## Pilot scope

3 frozen bug-fix scenarios · deterministic distractor manifests at **1x/5x/20x/50x** (committed before runs) · `agentic_code_patch` arm only (`pdd_prompt_space` deferred) · primary outcomes = hidden-pass-rate + token usage, plus localization cost (files read / tool calls before first edit).

## Locked pre-registration decisions (design §10)

| Decision | Choice |
|----------|--------|
| Pilot agent + model | Codex CLI (GPT) |
| Base repo source | Hand-authored minimal repos |
| Filesystem tap | Linux container + OverlayFS (edits) + FUSE passthrough, byte-extent reads |
| Seeds per cell | 5 (→ 60 runs) |
| Per-run timeout | 30 min |

Two non-blocking Codex specifics (exact model id + reasoning effort; in-process vs shelled-out reads) are flagged for confirmation before runs.

## Notes for reviewers

- Docs-only change; no `pdd` package code is touched.
[...]

Issue #1209, opened 2026-05-26
Research: measure agentic localization degradation under repository bloat


gltanaka:
## Research question

Does irrelevant repository bloat increase localization cost and reduce hidden-test success for agentic code patching?

More concretely:

> Given the same target task, same model, same verifier, and same starting code, what happens to an agentic code-patching workflow as the surrounding repository grows with plausible but irrelevant files?

This issue is intentionally scoped around the agentic code-patching side first. A PDD prompt-space arm can be added as a comparison, but the minimum useful result is to measure whether dynamic agentic localization degrades as irrelevant repo size increases.

## Why this is needed

PDD's context-window and context-efficiency claims depend on a sharper empirical question than “are prompts shorter than code?” The important operational question is whether a workflow with curated source-of-truth context avoids the search and context dilution costs that agentic code patchers pay in large repositories.

Right now, related context work exists, but it does not answer this question directly:

- #789 proposes a context budget audit command. That helps measure hydrated prompt cost, but it does not measure how agentic code patching behaves as repo size grows.
- #797 proposes context-window scoring in the architecture graph. That makes context pressure visible, but it does not establish whether repo bloat actually harms localization or success.
- #888 shows large dev units can overwhelm `pdd fix`, but that is about large target files/dev units, not irrelevant surrounding repository size.
- #1114 shows large repos can make broad test execution time out, but that is a test-scope problem, not an agentic localization benchmark.

Without this experiment, we cannot distinguish two very different realities:

1. **PDD has a real effective-context advantage.** Curated prompt/test/spec context stays stable while agentic code patching reads more irrelevant files, burns more tokens, touches wrong files more often, or fails hidden tests as the repo grows.
2. **Modern agentic tools handle repo bloat well.** They find the right files with little extra cost even in a large repo, meaning the PDD context-window claim should be narrowed or reframed.

Both outcomes are useful. A positive result supports investment in PDD context manifests, context audits, prompt-owned module boundaries, and source-of-truth workflows. A negative result prevents overclaiming and tells us that repo bloat is not the strongest differentiator.

This is also a good research task for an intern because it is measurable, bounded, and does not require solving the whole PDD-vs-agentic-coding question. The work is mostly experimental infrastructure, instrumentation, and analysis.

## Proposed experiment

Use a small set of frozen, testable maintenance tasks. For each task, create repo-size variants by adding plausible but irrelevant distractor files around the unchanged target task.

Suggested sizes:

```text
[...]

PR #1419, opened 2026-06-04
docs(research): real-OSS subset-and-regrow + context-penetration metrics (#1209)

sohni-tagirisa:
## Summary

Builds on #1401 (which adds the repo-bloat localization pilot design). Revises the **pilot design before any model run** to (a) source repos from real open source rather than hand-authoring them, and (b) add the context-window-penetration and token-dose metrics. 

Research question is unchanged: does irrelevant repository bloat increase localization cost and reduce hidden-test success for agentic code patching?

## What changed (and why)

**1. Real-OSS "subset-and-regrow" replaces hand-authored repos (locked decision §10 #3).**
Hand-authoring believable repos *and* a 50x distractor pool is costly and carries an "honesty cost" — synthetic siblings an agent may pattern-match as filler. Instead:
- Start from a **real OSS snapshot**, pinned + vendored (license/provenance recorded).
- **Dependency-slice** it to a minimal runnable **core** and **seed one controlled bug** → this is the 1x base repo.
- The repo's **own files outside the core** are the distractor pool; bloat is produced by **regrowing** them to a **token budget** at 1x/5x/20x/50x.
- A file is a distractor **iff it is outside the core's dependency closure** — decided statically *before* runs, so it stays a controlled variable (not "whatever the agent skipped").

**Contamination guardrail:** the repo may be memorized, but the **seeded fix is novel**, so the fix can't be recalled — only located. Residual layout-familiarity is a recorded caveat.

**2. Context-window penetration metrics (new in §6.1/§6.2).**
Distinguishes three layers the prior design collapsed: on-disk *dose* ⊇ *visited* (FS tap) ⊇ *in-context* (transcript). Adds `distractor_tokens_in_context`, `distractor_context_share`, `distractor_pool_penetration`, and the **visited-but-not-contextualized** set, reconciled against provider `usage`.

**3. Token-dose vs. fixability (§5.1, §7.2, §7.5).**
Dose axis is now token-budgeted; adds the headline `hidden_pass_rate ≈ f(distractor_tokens_in_context)` regression and a pre-registered penetration threshold.

Propagated through `scenario.json` / manifest / run-record schemas, architecture, directory layout, methodology note, execution checklist, non-goals, and the README locked-decision table. All internal anchor links re-validated.

## Notes for reviewers

- Docs-only; no `pdd` package code touched.
- This **revises a previously locked pre-registration decision (§10 #3)** — legitimate because no runs have occurred; the change is recorded in a header revision note rather than silently edited.
- The rendered `design.html` / PDF are not regenerated here; they need re-rendering from the updated `design.md`.


gltanaka (changes requested):
I think this direction is needed: issue #1209 asks for realistic plausible distractors, and replacing hand-authored repos with real OSS subset-and-regrow addresses the synthetic-filler honesty problem before any runs. I would not merge this revision yet. The changes make the design more realistic, but they introduce validity gaps that need to be fixed before this becomes the locked preregistration.

Required changes:

1. Fix the contamination / novel-fix claim. `research/repo-bloat-benchmark/design.md:22`, `:187`, and `:698` say the seeded bug means the fix is novel / absent from training. For a public OSS snapshot, a seeded bug often means the oracle fix is "restore the original upstream code", which may be exactly the memorized public file. A novel hidden verifier does not prevent code/layout recall. Require a seed-novelty audit proving the oracle patch is not a byte-for-byte or semantic restoration of the upstream version, or downgrade the claim and treat upstream-code recall as a measured caveat / stratification criterion. The current "fix cannot be memorized" assertion is too strong.

2. Add variant / oracle equivalence validation. `design.md:188` derives the core from visible tests, while higher sizes regrow out-of-core files at original paths. Hidden-verifier deps, dynamic imports, package discovery, plugins/config, optional deps, or import shadowing can make behavior differ by size, violating `design.md:57`'s "only one thing varies" invariant. Before model runs, the harness should apply the oracle fix and run visible + hidden verification against every materialized `1x` / `5x` / `20x` / `50x` variant under one fixed dependency environment; baseline expected outcomes should also be recorded. Hidden-verifier support/runtime deps must be included in the invariant core or harness support, not in the size-varying pool.

3. Fix context-penetration accounting before making it a primary / token-dose result. `design.md:393-395` defines `distractor_tokens_in_context` as summed over the run, then divides by `distractor_tokens_on_disk` for `distractor_pool_penetration`. Repeatedly surfacing the same file can make this "fraction" exceed 1. Split total window occupancy (repeat-counted; denominator `input_tokens_per_run`) from unique pool coverage (deduped/capped per file; denominator on-disk distractor tokens), or redefine the metric. Also, `design.md:418` and `:694` allow token-level attribution to fall back to best-effort if Codex transcript content is unavailable, but `design.md:37`, `:564-572`, and `:610-611` make token-dose / penetration primary and threshold-bearing. Either make tool-result content capture a pre-run blocker for these metrics, or demote/remove the token-dose thresholds until the source is confirmed.

CI/content note: docs-only; `git diff --check origin/main...pr-1419` passes. This is not a package-code blocker. It is a preregistration validity blocker.

Decision: needed, but do not merge until the above are fixed.

sohni-tagirisa:
Addressed the three review blockers in `c9fc8830b`:

- Downgraded the contamination/novel-fix claim: public upstream recall is now an explicit caveat, and each scenario requires a `seed_novelty.md` audit that rejects byte-for-byte upstream restorations and classifies semantic restorations as upstream-recall caveats.
- Added a pre-run variant/oracle equivalence gate across `1x` / `5x` / `20x` / `50x`, requiring registered baseline outcomes plus oracle-fixed visible + hidden verification under one fixed dependency environment. Hidden-verifier support/runtime deps must live in invariant core/harness support, not the size-varying pool.
- Split context-window accounting into repeat-counted occupancy (`distractor_context_tokens_total`, `distractor_context_share`) and bounded unique pool coverage (`distractor_unique_tokens_entered_context`, `distractor_pool_coverage`). Token-level context penetration is now a pre-run blocker unless transcript/tool wrappers capture surfaced content; no best-effort token-dose thresholds remain.

Verification: `git diff --check` passed, and targeted scans for stale metric names / over-strong memorization claims passed. CI is queued on the pushed commit.

sohni-tagirisa:
Follow-up in `2b81e5125` after re-verifying the three blockers against the design end-to-end. All three hold:

- **Contamination/novel-fix (blocker 1):** claim is downgraded everywhere — no `cannot be memorized` / `absent from training` assertions remain (`design.md:22`, §10 consequence, §11), and §4.1.1 step 6 gates each seed on a `seed_novelty.md` audit that rejects byte-for-byte upstream restorations and classifies semantic restorations as an upstream-recall stratum.
- **Oracle equivalence (blocker 2):** §4.1.2 runs visible+hidden verification on the oracle-fixed and baseline state for every `1x/5x/20x/50x` variant under one fixed dep env, with an import/module-resolution diff vs `1x`, and keeps hidden-verifier support/runtime deps in invariant core/harness, never the size-varying pool.
- **Context-penetration accounting (blocker 3):** repeat-counted occupancy (`distractor_context_share` = total / `input_tokens_per_run`, reconciled so per-request attribution ≤ measured input ⇒ bounded ≤1) is now split from unique coverage (`distractor_pool_coverage` = unique / on-disk, bounded [0,1]); token-level penetration/dose is a pre-run blocker, no best-effort fallback remains.

This commit is verification + three small accuracy fixes only:
1. `Branch:` header referenced a nonexistent branch — corrected to `research/issue-1209-oss-subset-regrow`.
2. Manifest field `distractor_tokens` → `distractor_tokens_on_disk` so the provisioned on-disk dose has one canonical name (the `(distractor_tokens_on_disk, §5)` cross-ref in §6 and the run-record field now match the manifest).
3. Added `oracle_equivalence_passed` to the run-record schema so each run is auditably tied to a passed §4.1.2 gate, mirroring `calibration_passed`.

All 28 internal anchors re-validated; docs-only, no package code touched.

gltanaka (approved):
The latest revision addresses my previous preregistration blockers well enough to merge as the design/workspace update.

Verified fixes:

1. The contamination claim is now appropriately downgraded. The design no longer asserts that a seeded public-OSS fix is automatically absent from training; it requires `seed_novelty.md`, rejects byte-for-byte upstream restorations, and treats semantic restorations as an upstream-recall caveat/stratum.

2. The variant/oracle equivalence risk is now covered by a pre-run gate across `1x` / `5x` / `20x` / `50x`: registered baseline outcomes plus oracle-fixed visible + hidden verification under one fixed dependency environment, with verifier/runtime deps kept out of the size-varying pool. `oracle_equivalence_passed` is also recorded in the run schema.

3. The context-penetration metrics are split correctly: repeat-counted occupancy (`distractor_context_tokens_total`, `distractor_context_share`) is separate from bounded unique pool coverage (`distractor_unique_tokens_entered_context`, `distractor_pool_coverage`). Token-level attribution now requires transcript tool-result content or a harness wrapper; otherwise token-dose thresholds must be demoted/removed before runs, so the design no longer relies on best-effort attribution.

Validation: docs-only diff; `git diff --check origin/main...pr-1419` passes. Unit tests, public CLI regression, package smoke, CodeQL, and member-triggered auto-heal pass. The remaining `heal` failure appears to be the same non-content authorization-triggered job pattern seen on prior PRs.

PR #1440, opened 2026-06-05
fix(ci): scope-preserving PR auto-heal — never escalate to test_extend (#1403)

### LINKED ISSUE #1403: PR auto-heal should not escalate narrow fix PRs into unrelated test_extend coverage churn
## Problem

PR auto-heal is currently optimizing for "make the whole module synchronized" instead of "keep this PR narrowly scoped".

That is causing narrow fix PRs to get re-bloated by unrelated `test_extend` output.

Concrete example:
- PR: `promptdriven/pdd#1390`
- Core fix commit: `9fbf9e041` (`fix(agentic): fast-fail interactive Claude on subscription/usage caps`)
- Auto-heal head: `4ca11f357` (`chore: auto-heal prompt/example drift for agentic_common`)

At PR head, the actual bug fix is still fine, but auto-heal has appended a large unrelated test block to `tests/test_agentic_common.py` and changed `.pdd/meta/agentic_common_python.json` to `"command": "test_extend"`.

Greg's review on June 4, 2026 is correct: the PR is non-mergeable because auto-heal reintroduced unrelated generated tests.

## Why this happens

PR auto-heal runs the installed `pdd-cli` against changed modules:
- `scripts/ci/run_pdd_cli_auto_heal_pr.sh` calls `python -I -m pdd.ci_drift_heal --diff-base ... --modules ...`

Inside `pdd.ci_drift_heal`:
- `detect_drift()` calls `sync_determine_operation(..., target_coverage=90.0, log_mode=True)`
- `heal_module()` sends several operations through plain `pdd sync <module>`

Inside `sync_determine_operation()`:
- for Python modules, if tests pass but coverage is below target, the decision is intentionally `test_extend`

That behavior is valid for full module sync, but it is wrong for PR auto-heal. In PR mode, it turns a narrow production fix into an unrelated coverage-expansion PR.

## Expected behavior

When `ci_drift_heal` is running in PR auto-heal mode (`diff_base` set / changed-modules mode), it should preserve PR scope.

Allowed in PR auto-heal:
- narrow prompt/code/example/meta drift fixes
- `update`
- `example`
- possibly `auto-deps` when clearly required by the changed prompt/code pair

Disallowed in PR auto-heal:
- coverage-driven `test_extend`
- broad `pdd sync` fan-out that appends unrelated tests to satisfy module-level coverage targets
- any auto-generated test block unrelated to the PR's changed behavior

## Required fix

Add a PR-scope guard to the auto-heal path so `test_extend` is never auto-committed during PR heal runs.

Important: detection-only is not enough.
- Even if `detect_drift()` suppresses `test_extend`, `heal_module()` can still route other ops through `pdd sync`, and `sync` can escalate into `test_extend` internally.
- The guard needs to exist at execution time too, not just at initial drift classification.

Acceptable fixes include either:
1. A dedicated PR-heal mode in `ci_drift_heal` that only permits narrow operations and never calls module-wide `pdd sync` for review-scope heals, or
2. A scoped `pdd sync` mode / flag that forbids coverage-expanding `test_extend` and returns a no-op / manual-follow-up decision instead.

## Regression coverage wanted

Add tests proving that in PR auto-heal mode:
- a narrow PR does not get reclassified into coverage growth
- `test_extend` is not dispatched
- review-only / scope-preserving example drift still works
- the resulting commit does not append unrelated generated tests

## Temporary workaround

For affected PRs today:
- keep the PR in draft before force-pushing cleanup, so `synchronize` on a draft PR does not re-run auto-heal
- do not mark it ready for review again until this bug is fixed, because `ready_for_review` retriggers auto-heal
- if the change must land immediately, merge via a path that temporarily bypasses PR auto-heal rather than letting the bot re-bloat the branch

Serhan-Asad:
## Problem (#1403)

PR auto-heal was optimizing for "make the whole module synchronized" instead of "keep this PR narrowly scoped", re-bloating narrow fix PRs with unrelated `test_extend` output (e.g. #1390, flagged non-mergeable on 6/4).

For a Python module whose tests pass but coverage is below target, `sync_determine_operation` returns `test_extend`. In PR auto-heal, `heal_module` routes `verify/generate/test/crash` through `pdd --force --strength 0.5 sync <module>`, and that nested `pdd sync` re-derives the same coverage gap **internally** and appends generated tests (rewriting `.pdd/meta` command to `test_extend`). As the issue stresses, **detection-only suppression is not enough** — the guard must also exist at execution time.

## Fix — a two-layer guard from one default-off signal

`PDD_DISABLE_TEST_EXTEND`, set only in PR auto-heal mode (`is_pr_mode = not skip_ci`):

| Layer | Where | Behavior when flag set |
|-------|-------|------------------------|
| Detection | `sync_determine_operation` (`test_extend_disabled()`) | coverage-gap branch returns the existing `all_synced` no-op for **all** languages. Called by **both** the parent `detect_drift` and the nested `pdd sync`, so one branch covers detection **and** execution re-derivation. |
| Execution backstop | `sync_orchestration` `test_extend` branch | mirrors the existing non-Python skip: logs `test_extend_skipped`, accepts current state, writes no test file. |
| Signal source | `ci_drift_heal.main` | sets `os.environ` flag **only** around the in-process `detect_drift` (restored in `finally` — no leak) and passes it explicitly in the `pdd sync` subprocess `env`. |
| Clean no-op | `ci_drift_heal.detect_drift` | treats `all_synced` as "no drift" (with `nothing`/`synced`) so the guarded module is a clean skip, not an "unknown operation" heal failure. |

Push-to-main (`--skip-ci`) is **unaffected** — whole-module coverage growth still runs. The signal is default-off, so unset behavior is byte-for-byte unchanged. `update` / `example` / required `auto-deps` remain allowed in PR heal.

Prompts (`ci_drift_heal`, `sync_determine_operation`, `sync_orchestration`) updated to match (source of truth).

## Tests (genuine RED → GREEN; verified failing on unpatched source)

- `sync_determine_operation`: flag → `all_synced`; **unset still returns `test_extend`** (regression guard).
- `sync_orchestration`: suppressed `test_extend` appends no tests, logs `test_extend_skipped`, saves no `test_extend` fingerprint; **Python `test_extend` still runs when unset** (baseline).
- `ci_drift_heal`: PR mode sets the flag for in-process detection **and** the subprocess env, and restores `os.environ` (no leak); **push-to-main with `--diff-base` keeps it unset** (the discriminator is `skip_ci`, not `diff_base`).
- e2e (`test_ci_drift_heal_e2e.py`): real `python -m pdd.ci_drift_heal` parent + PATH-level fake `pdd` child proves the env contract crosses the parent→child boundary and no `test_extend` churn is committed.

Local: `test_ci_drift_heal` + `test_sync_determine_operation` + e2e = **315 passed**; `test_sync_orchestration` = **214 passed**. Zero new ruff findings on the changed files.

## Supersedes

- **#1416** (`pdd change`) — edited prompt prose only; code was never regenerated, so the guard was non-functional.
- **#1432** (`pdd fix`) — failing tests + partial code; missed the `all_synced` detect_drift filter (guarded module became an "unknown operation" failure) and left `os.environ` mutated.

This PR is the complete, tested version of both. Opened as **draft** per the issue's workaround so the (still-unpatched) auto-heal bot doesn't re-bloat this branch before merge.

Closes #1403.


Serhan-Asad:
### Review loop converged (claude implement ↔ codex review, 3 rounds)

- **Round 1** — codex found a real regression: `detect_drift` skipped `all_synced` *before* git reclassification, dropping code-only-change modules that should become `update`. **Fixed** (`e6b34498c`): the `all_synced` skip now runs *after* reclassification; added tests `test_all_synced_with_code_only_change_reclassified_as_update` + `test_all_synced_terminal_is_skipped_not_unknown_operation`. Also renamed `test_extend_disabled` → `is_test_extend_disabled` (so pytest doesn't collect the helper) and added falsey-parser + env-restore-on-exception tests.
- **Round 2/3** — codex flagged that the coverage branch returns before file-state analysis, so an example-only change coexisting with a low-coverage run report is masked. Confirmed **pre-existing** (identical on `main`, which returns `test_extend` → fails the heal) and independent of this guard; the shipped non-Python `all_synced` branch has the same property. There is **no scenario where `main` heals that example but this PR doesn't** — the divergence is strictly fail/re-bloat → clean skip, and `PDD_HEAL_SKIP_REVIEW_ONLY_EXAMPLE_DRIFT=1` already skips review-only example drift by design. Codex agreed it is **not a blocker** and merge-ready with a documented follow-up. Tracked in #1442.

Kept as **draft** intentionally (issue #1403's own workaround) so the still-unpatched auto-heal bot can't re-bloat this branch before merge.

PR #1521, opened 2026-06-09
Compare architecture conformance semantically

### LINKED ISSUE #1520: Make architecture conformance semantic instead of text-exact
## Problem

PDD architecture conformance catches real drift, which is valuable, but parts of it compare interface text too literally. This causes false failures during `pdd sync` when generated code is semantically correct but spelled differently.

Observed false-positive examples:
- `dict[str, Any]` vs `Dict[str, Any]`
- `list[str]` vs `List[str]`
- `set[str]` vs `Set[str]`
- `Optional[T]` vs `T | None` / `Union[T, None]`
- entrypoint/bootstrap files that correctly export no named symbols
- declared camelCase public API names, such as Firebase exports

Related:
- #1458 / #1509: entrypoint and semantic-conformance work in progress
- #1446 / #1448: declared camelCase exports
- #865 / #866: structured conformance repair diagnostics

## Desired Behavior

Conformance should compare public interface contracts semantically:
- Equivalent Python typing spellings should pass.
- Explicit `entrypoint` modules should skip named-export checks.
- Declared camelCase public exports should pass.
- Missing declared symbols, missing params, changed defaults, changed container base types, and undeclared public-surface regressions should still fail.

## Acceptance Criteria

- `Dict[str, Any]` and `dict[str, Any]` are compatible.
- `Optional[Set[str]]`, `Set[str] | None`, and `Union[Set[str], None]` are compatible.
- Bare container declarations remain compatible with typed versions of the same container, e.g. `set` vs `Set[str]`.
- Different container bases still fail, e.g. `List[str]` vs `Dict[str, str]`.
- Optionality drift still fails, e.g. `Optional[str]` vs `str`.
- `interface.type = "entrypoint"` allows files with no named exports.
- Declared camelCase public API exports continue to pass; undeclared camelCase still fails.
- Existing conformance diagnostics and repair directives remain structured/actionable.

Serhan-Asad:
## Summary
- Add a stdlib-only semantic interface helper for normalized type and callable compatibility decisions.
- Use semantic annotation checks for <pdd-interface> verification and semantic callable compatibility for public-surface regressions.
- Add first-class entrypoint interface support in validation, architecture templates, prompt docs, and frontend architecture types.

## Tests
- env -u CONDA_PREFIX -u CONDA_DEFAULT_ENV -u CONDA_EXE -u CONDA_PYTHON_EXE -u CONDA_SHLVL -u CONDA_PROMPT_MODIFIER PYTHONPATH=. uv run --project . python -m pytest tests/test_interface_semantics.py -q
- env -u CONDA_PREFIX -u CONDA_DEFAULT_ENV -u CONDA_EXE -u CONDA_PYTHON_EXE -u CONDA_SHLVL -u CONDA_PROMPT_MODIFIER PYTHONPATH=. uv run --project . python -m pytest tests/test_architecture_sync.py -q
- env -u CONDA_PREFIX -u CONDA_DEFAULT_ENV -u CONDA_EXE -u CONDA_PYTHON_EXE -u CONDA_SHLVL -u CONDA_PROMPT_MODIFIER PYTHONPATH=. uv run --project . python -m pytest tests/test_code_generator_main.py -q

Closes #1520


gltanaka (changes requested):
This change is needed in principle: #1520 describes real architecture-conformance false positives (typing aliases, Optional/Union spellings, and entrypoint files), and the PR covers the main advertised cases with passing tests. I would not merge it as-is, though. Required changes:

1. Fix the public-surface false negative for added defaulted positional parameters in front of an existing `*args`. `pdd/interface_semantics.py` currently allows any new non-required param at lines 269-275, so `def f(a, *args)` -> `def f(a, b=0, *args)` is treated as compatible. That changes existing calls like `f(1, 2)`: `2` used to land in `args`, but now binds to `b`. This should remain a signature regression unless explicitly allowed by a BREAKING-CHANGE directive. Please tighten `callable_contracts_compatible` and add tests for both `signature_entries_compatible("[function] (a, *args)", "[function] (a, b=0, *args)") is False` and `_verify_public_surface_regression` raising for that code change.

2. Clean the whitespace failures in the generated examples. `git diff --check origin/main...HEAD` currently fails on trailing whitespace in `context/architecture_sync_example.py` lines 15/17/20 and `context/code_generator_main_example.py` line 5.

Validation I ran locally on the PR branch: `tests/test_interface_semantics.py`, `tests/test_architecture_sync.py`, `tests/test_code_generator_main.py`, and the combined three-file pytest run all pass. `pylint` was not available in the `uv` environment. The blocker is the false negative above; the whitespace cleanup is also required before merge hygiene is acceptable.

Serhan-Asad:
Both required changes are in, plus the symmetric case closed in the same pass (commits 113d937, 2ff8863).

**1. Positional parameter before `*args` (the false negative).** `callable_contracts_compatible` now treats a new positional-capable parameter as a regression when the old signature has `*args`: `def f(a, *args)` → `def f(a, b=0, *args)` is rejected because `f(1, 2)` rebinds the `2` from `args` to `b`. Keyword-only additions after `*args`, and defaulted positionals with no `*args` to absorb the shift, stay compatible. Added the requested `signature_entries_compatible("[function] (a, *args)", "[function] (a, b=0, *args)") is False` case and a `_verify_public_surface_regression` integration test that raises for the equivalent code change (`changed_signatures == ["process"]`).

**2. Trailing whitespace.** Stripped from `context/architecture_sync_example.py` and `context/code_generator_main_example.py`; `git diff --check upstream/main...HEAD` is clean.

I also closed the symmetric `**kwargs` hole, since it has the same rebinding property: `def f(**kwargs)` → `def f(a=0, **kwargs)` rebinds `f(a=5)` from `kwargs['a']` to `a`, so a new keyword-capable parameter is now a regression when the old signature has `**kwargs`. A positional-only addition can't be passed by keyword, so it never steals from `**kwargs` and stays compatible. Both directions are covered by tests.

`test_interface_semantics.py`, `test_architecture_sync.py`, and `test_code_generator_main.py` all pass locally; regressions in either direction require an explicit `BREAKING-CHANGE: change signature` opt-out, consistent with the existing gate.

gltanaka (approved):
The requested blockers are fixed. The `*args` false negative now raises, the follow-up `**kwargs` capture case is covered too, `git diff --check` is clean, local targeted tests pass, and the rerun of `auto-heal` is green. This makes sense to merge from my review perspective. Note: GitHub currently marks the branch as behind `main`, so update it first if branch protection requires an up-to-date head.

PR #1546, opened 2026-06-10
fix(conformance): compare interface default values semantically (#1545)

### LINKED ISSUE #1545: Conformance should compare semantic interface contracts, not raw default strings
## Problem

Issue #1541 / PR #1542 exposed a deeper conformance design gap: PDD has started moving architecture/interface checks toward semantic comparison, but parts of the checker still compare raw source strings.

Current upstream already has `pdd/interface_semantics.py` for semantic annotation and public-surface signature compatibility, but `_verify_pdd_interface_signatures` in `pdd/code_generator_main.py` still compares default values as strings.

That causes false-positive conformance failures like:

```python
def _sanitize_comment_body(body: str, max_chars: int = 25000):
    ...
```

vs generated code:

```python
_COMMENT_MAX_CHARS = 25000

def _sanitize_comment_body(body: str, max_chars: int = _COMMENT_MAX_CHARS):
    ...
```

Those are equivalent for callers, but the checker reports drift because:

```text
"25000" != "_COMMENT_MAX_CHARS"
```

This makes conformance feel like it needs daily hard-coded exceptions. The fix should be a general semantic contract layer, not another special case.

## Evidence

On current `main` after #1521:

- `Dict` vs `Dict[str, Any]` is already semantically accepted by `annotations_compatible`.
- `25000` vs `_COMMENT_MAX_CHARS` still fails as default drift.
- Missing functions and public symbol removals are correctly blocked and should stay strict.

Related:

- #1541
- #1542
- #1521

## Desired Design

Centralize conformance into a shared semantic interface comparator.

Inputs should normalize into one internal model:

```text
prompt <pdd-interface> -> InterfaceContract
architecture.json      -> InterfaceContract
generated code AST     -> InterfaceContract
existing code AST      -> InterfaceContract
```

Then one comparator decides compatibility.

`architecture.json` should remain the stored declared contract/index. It should not store ad-hoc exception rules.

`pdd/interface_semantics.py` should become the shared place for interface compatibility, including default values, not only annotations/callable shapes.

## Default Semantics

Add a `DefaultContract` / default-normalization helper that returns a tri-state result:

```text
compatible
incompatible
unknown
```

Safe compatible examples:

```python
25000
25_000
_COMMENT_MAX_CHARS  # when module-level _COMMENT_MAX_CHARS = 25000
None
False
"literal"
("a", "b")
```

Real incompatible examples:

```python
25000
5000
```

Unknown examples should fail closed or produce a clear hard diagnostic:

```python
DEFAULT_LIMIT = get_limit()
DEFAULT_LIMIT = os.getenv("LIMIT")
DEFAULT_LIMIT = SOME_IMPORTED_VALUE
```

The checker must not execute code, import modules, or evaluate arbitrary expressions. It should only resolve safe AST literals and same-module constants assigned to safe literals/containers.

## Keep These Strict

Conformance should still hard-fail on real contract breaks:

- function/method declared in `<pdd-interface>` is missing
- declared parameter is missing
- generated code drops a declared default
- default value is provably different
- annotation is semantically incompatible
- public class/function/constant is removed without explicit `BREAKING-CHANGE`
- public callable signature changes incompatibly

## Implementation Notes

- Move default compatibility out of `code_generator_main.py` raw string comparison and into `interface_semantics.py`.
- Make `_verify_pdd_interface_signatures` use the shared default comparator.
- Make public-surface signature compatibility use the same default comparator where applicable.
- Consider registering `interface_semantics.py` as a first-class prompt/architecture module so this shared conformance engine is itself PDD-maintained rather than an untracked utility.

## Acceptance Criteria

- Add regression tests proving `25000`, `25_000`, and a safe module constant assigned to `25000` are treated as equivalent defaults.
- Add tests proving unsafe/dynamic defaults are not silently accepted.
- Add a fixture based on the #1541 / #1542 `agentic_common` failure.
- Keep tests proving missing `extract_step_report` still fails.
- Keep tests proving public API removals like `AgenticTaskResult.usage` still fail.
- No new per-symbol/per-module hard-coded allowlist for `_COMMENT_MAX_CHARS` or `agentic_common`.

Serhan-Asad:
Closes #1545

## Problem

Conformance compared parameter **default values as raw source strings**, so a
prompt `<pdd-interface>` declaring a literal default and generated code that
factored the same value into a module constant were reported as drift:

```text
declared: max_chars: int = 25000
actual:   max_chars: int = _COMMENT_MAX_CHARS   # _COMMENT_MAX_CHARS = 25000
=> "25000" != "_COMMENT_MAX_CHARS"  -> false-positive conformance failure
```

This made conformance feel like it needed hard-coded per-symbol exceptions
(issue #1545).

## Fix

A single **semantic default comparator** in `pdd/interface_semantics.py`
(the shared interface-compatibility module from #1521), used by both
conformance gates. It returns a tri-state verdict and **never executes code,
imports a module, or evaluates an arbitrary expression** — it only folds safe
AST literals, unary `+`/`-`/`~` on numbers, literal containers, and
same-module constants bound to those:

- `COMPATIBLE`   — both defaults resolve to the same safe value (`25000` vs
  `25_000`, or a module constant `_LIMIT = 25000` standing in for `25000`).
- `INCOMPATIBLE` — both resolve to safe literals that differ (`25000` vs `5000`).
- `UNKNOWN`      — at least one side is not statically resolvable (a call, an
  imported name, a dynamic expression).

**Regression-free by construction:** the comparator is consulted only inside
the pre-existing "default strings differ" branch of each gate, and a failure
is suppressed **only** on `COMPATIBLE`. `INCOMPATIBLE` and `UNKNOWN` keep
failing exactly as before, so the change can only remove false positives,
never add a false negative.

### Where it's wired
- `_verify_pdd_interface_signatures` (`code_generator_main.py`) — the prompt
  `<pdd-interface>` gate. It has the generated module AST, so it resolves
  same-module constants via `build_module_default_symbols`.
- `_existing_param_compatible` (`interface_semantics.py`) — the public-surface
  signature gate. It compares snapshot signature strings with no module
  context, so it is **literal-normalization only** (a literal→constant default
  refactor stays `UNKNOWN` there and is conservatively treated as a change).

### Module-constant safety
A name is trusted only if it is bound **exactly once anywhere in the module**
and that single binding is a top-level assignment to a safe literal. Any
reassignment, augmented/conditional assignment, tuple-unpack rebind, or a
`def`/`class`/`import`/`except as`/`match` capture that shadows the name drops
it to `UNKNOWN` (fail closed). Name-to-name aliases are not resolved
transitively.

## Kept strict
Missing declared functions/parameters, dropped defaults, provably-different
defaults, semantically-incompatible annotations, and public symbol removals
(e.g. `AgenticTaskResult.usage`) all still hard-fail. No per-symbol or
per-module allowlist was added.

## Tests
- `tests/test_interface_semantics.py`: comparator equivalence
  (`25000`/`25_000`, quote/whitespace, set/dict order, unary), provable
  difference (`1` vs `True`, `int` vs `float`, `tuple` vs `list`), fail-closed
  `UNKNOWN` (calls, imported names, f-strings, dynamic/shadowed constants),
  symbol-table safety (reassign / augmented / conditional / unpack /
  `def`/`class`/`import`/`except as` shadows / no transitive alias), and the
  documented public-surface scope.
- `tests/test_code_generator_main.py`: the agentic_common literal→module-const
  fixture passes; `25_000` reformat passes; a constant with a different value
  still raises; a dynamic default is not silently accepted; a missing declared
  function still fails; `AgenticTaskResult.usage` removal still fails; a public
  literal-default reformat is not a regression.

All four affected test files pass; `pdd/interface_semantics.py` and
`pdd/code_generator_main.py` lint clean (no new categories; score up).


gltanaka (changes requested):
I think the underlying fix is needed: #1545 describes a real false-positive in the current raw-string default comparison. I would not merge this implementation yet, though, because the new comparator can mark defaults as `COMPATIBLE` when Python runtime behavior is not compatible.

1. Make same-module constant resolution source-order aware and scoped to the default expression being checked. `build_module_default_symbols(generated_code)` currently builds a whole-module table, so this resolves as compatible:

```py
def f(x=_LIMIT):
    pass

_LIMIT = 25000
```

`compare_default_sources("25000", "_LIMIT", symbols)` returns `COMPATIBLE`, but executing the module raises `NameError` because parameter defaults are evaluated when the `def` statement executes. Only resolve constants that are already bound before the relevant function/method definition, and add a regression for assignment-after-def staying `UNKNOWN`/drift.

2. Do not use the generated module's symbol table to interpret the prompt-declared default. The prompt is the contract; generated code should not be able to define the meaning of a prompt-side bare name. Today this returns `COMPATIBLE`:

```py
DEFAULT_LIMIT = 5000

def f(x=5000):
    pass
```

with `compare_default_sources("DEFAULT_LIMIT", "5000", build_module_default_symbols(src))`. Use separate declared/actual symbol contexts, or no generated symbol table for the declared side, and add a regression for this case.

3. Reconsider dict normalization. The comparator treats `{'a': 1, 'b': 2}` and `{'b': 2, 'a': 1}` as compatible, but Python dict insertion order is observable (`list(default)` differs). Either preserve order in normalized dict defaults or classify dict defaults as `UNKNOWN` unless the implementation can prove behavior is the same.

The focused tests pass, but they do not cover these false-compatible cases. Please add coverage for the cases above before this is merged.

Serhan-Asad:
Closing this PR rather than merging it because it only fixes the narrow `<pdd-interface>` default-comparison path from #1545.

The same user-visible sync failure can still happen in the public-surface regression gate: existing code with `def f(x=25000)` regenerated as `_LIMIT = 25000; def f(x=_LIMIT)` can pass architecture conformance but still fail as `signature_changed: f`. That means the PR reduces one symptom but does not fully solve the bad `pdd sync` UX.

Tracking the broader fix in #1558: conformance should use one semantic callable-contract/default comparator across all gates, including public-surface checks, with deterministic retry behavior.

PR #1550, opened 2026-06-11
Emit structured PDD_PROVIDER_LIMIT reset marker for cloud scheduling (#1541)

### LINKED ISSUE #1541: Emit structured agentic credential-limit reset metadata for cloud executor scheduling
## Summary

PDD Cloud needs to queue PDD executor runs when the current agentic credential hits a rate / usage / credential limit, then resume automatically after the provider reset time.

Today cloud can detect some limit cases from strings like `api_error_status:429` or `credential-limit`, but it cannot reliably extract a reset time across providers. Add a stable, secret-safe output marker from the PDD CLI whenever an agentic provider/credential hits a real limit with a known or inferred reset time.

## Proposed marker

```text
PDD_PROVIDER_LIMIT provider=<anthropic|openai|google|antigravity> status=<429|credential_limit> reason=<rate_limit|usage_limit|session_limit|credential_limit> reset_at=<ISO-8601 UTC or empty> reset_source=<provider|parsed_text|estimated|none>
```

The exact field names can change if there is already a better local convention, but the output must be stable enough for PDD Cloud to parse without scraping raw provider stderr.

## Requirements

- Claude Code limit messages like `You've hit your limit · resets ...` produce a structured marker with `reset_at` when parseable.
- Session, usage, and weekly limit variants classify as credential/provider limits.
- Existing `credential-limit` behavior remains backward-compatible.
- The marker never includes raw provider stderr, tokens, API keys, user prompt content, or other untrusted text.
- Generic 429s without reset information still produce a stable limit marker with empty `reset_at` / `reset_source=none` or equivalent.

## Acceptance criteria

- Tests cover exact reset timestamp/date strings.
- Tests cover time-only reset strings.
- Tests cover unparseable reset strings.
- Tests cover generic 429 without reset.
- Tests cover false-positive prose containing “hit your limit” but not an actual provider limit.
- PDD Cloud can consume the marker to schedule a retry without depending on provider-specific raw output.

## PDD Command Hint

This is a CLI behavior change. Use `pdd-change` / `pdd-sync` if implementing with PDD.

Serhan-Asad:
## Summary

Closes #1541. Emits a stable, secret-safe `PDD_PROVIDER_LIMIT` marker whenever an agentic provider/credential hits a real rate/usage/session/credential limit, so PDD Cloud (tracking: promptdriven/pdd_cloud#2046) can schedule a retry after the provider reset window **without scraping raw provider stderr**.

```text
PDD_PROVIDER_LIMIT provider=<anthropic|openai|google|antigravity|opencode> status=<429|credential_limit> reason=<rate_limit|usage_limit|session_limit|credential_limit> reset_at=<UTC ISO-8601 or empty> reset_source=<provider|parsed_text|estimated|none>
```

## Root cause / motivation

Today cloud can detect *that* a limit happened (`api_error_status:429`, the stable `credential-limit` token) but cannot reliably extract a **reset time**, the `credential-limit` diagnostic is quiet-gated, and generic 429s emit nothing machine-readable. This PR closes that gap additively.

## What changed

- **`pdd/agentic_common.py`** — new private helpers next to the existing classifier: `_parse_reset_at` (stdlib-only reset-time parser → UTC `YYYY-MM-DDTHH:MM:SSZ`, injectable clock), `_classify_provider_limit`, `_provider_limit_marker`, `_credential_limit_reason`, `_marker_provider_label`, `_emit_provider_limit_marker`. A single emission call at the **post-retry-loop seam** in `run_agentic_task`.
- **`pdd/prompts/agentic_common_python.prompt`** — documents the marker contract in the Error Classification section (body only; `<pdd-interface>`/`<pdd-reason>` untouched).
- **`README.md`** — documents the marker under Agentic Fallback Mode + the `--fixer-fallback` bullet.
- **`tests/test_agentic_common.py`** — `TestProviderLimitMarker` (26 tests).

## Design notes

- **Piggybacks on the already-vetted classification** (`_classify_permanent_error` `credential-limit` + `_is_rate_limited`) so it inherits their false-positive guards and **never changes retry/backoff/fallback semantics** — the marker is purely additive.
- **Fires once per provider** (the seam is reached at most once per provider; a 429 that recovers on retry returns earlier and emits nothing).
- **Not quiet-gated** — printed on plain stdout so the cloud scheduler sees it even when `quiet=True` suppresses human diagnostics. Best-effort: marker errors are swallowed so they can never crash the run.
- **Secret-safe** — only fixed enum fields + the normalized timestamp are ever emitted; raw provider text is never echoed. An explicit-but-unrecognized timezone degrades to unparseable rather than a confidently-wrong UTC time.
- **Implemented directly, not via full `pdd sync` regen** — regenerating this ~4.5k-line module drifts its own `<pdd-interface>` conformance (which is what blocked the earlier auto-generated PR #1542). This PR supersedes #1542 with a complete, tested implementation.

## Validation

- `pytest tests/test_agentic_common.py` — 424 passed (incl. 26 new), CI marker filter, no regressions.
- Prompt/architecture consistency: `test_issue_902_prompt_sync.py`, `test_prompt_contract_validation.py`, `test_architecture_sync.py` — 174 passed.
- Other `run_agentic_task` stdout consumers (`test_update_main`, `test_one_session_sync`, `test_agentic_split`, …) — 357 passed (the extra marker line breaks nothing).
- Zero new ruff findings vs `origin/main` baseline.
- Reviewed adversarially with codex (gpt-5.5) over multiple rounds; all findings addressed.

Supersedes #1542 (auto-generated, prompt-only, code-gen failed conformance).


Serhan-Asad:
Thanks — reviewed all three. Pushed fixes for two; the third is a deliberate non-fix, reasoning below. Head `1ac3a6331`, full `tests/test_agentic_common.py` green (433 passed).

**1. Fractional ISO + offset — fixed.** `resets 2026-06-11T15:00:00.123-07:00` was partially matched: `_RESET_ISO_RE` had no fractional-seconds part, so the `.123` was unmatched and the trailing offset truncated, emitting `15:00Z` instead of `22:00Z` (7h early, in the trusted `parsed_text` tier). Added an optional `(?:\.\d+)?` before the zone; sub-seconds are dropped from the second-granularity `reset_at`, the offset is honored. Covered for `.123-07:00`, `.123456+05:30`, and fractional+`Z`.

**2. Bare IANA zone — fixed.** `resets 3:30pm America/Los_Angeles` (unparenthesized) defaulted to UTC → `15:30Z` instead of `22:30Z` — the same "an explicit zone must not silently become UTC" class already handled for abbreviations/offsets. `_TZ_TAIL` now also captures a bare `Area/Location` name, **anchored on the closed set of IANA area prefixes** (`Africa|America|…|US|Etc`) so the required area + `/` cannot match prose like `and/or`, and tried **before** the abbreviation branch so `US/Pacific` isn't truncated to the `US` abbrev. Resolution stays via `_resolve_reset_tz` (ZoneInfo authoritative; unknown → unparseable). Added a guard test that `resets 9pm and/or contact` is unaffected.

**3. Retry-After / exact metadata without `resets` — intentionally not fixed.** The trigger isn't a real format: HTTP `Retry-After` carries **seconds or an HTTP-date**, never a bare ISO timestamp (the real seconds form isn't parsed either, by the same design). The only way to honor exact metadata that isn't phrased as `resets …` is to scan arbitrary text for any ISO timestamp, which would manufacture false `reset_at` values from unrelated log timestamps embedded in provider output — a regression of the deliberately secret-safe, `resets`-anchored parser (it reads only structured groups, never surrounding free text). No provider in evidence emits this (the captured Claude error uses `resets`), so I've left it as a conscious design boundary rather than build a speculative parser.

Serhan-Asad:
## ✅ Independent full E2E verification + regression — at head `1ac3a6331`

Ran a fresh, isolated verification sweep of this PR in a dedicated worktree (no shared state). **Verdict: ship — producer is correct, zero regression.**

### Test & verification matrix

| Check | Result |
|---|---|
| GitHub **Run Unit Tests** on exact head `1ac3a6331` | ✅ `success` (authoritative full merged-tree suite) |
| Full `tests/test_agentic_common.py` | ✅ **433 passed / 0 failed**; `TestProviderLimitMarker` **35/35** |
| Merged-tree regression — every `run_agentic_task` stdout consumer + prompt/arch suites | ✅ **3353 passed / 0 failed** (`update_main`, `one_session_sync`, `agentic_split`, `fix`/`verify`/`update`/`crash`/`bug`/`checkup`/`change`/`architecture`/`test` orchestrators, `prompt_contract`, `architecture_sync`, …) |
| Ruff delta on `agentic_common.py` | ✅ 4 findings, all `git blame`-proven pre-existing — **0 new** attributable to the PR |
| Fuzz / no-leak invariant | ✅ **360,000 adversarial iterations** (6 seeds, fixed New-Year clock), **0 violations** — no secret needle (`sk-ant-`, `ghp_`, JWT, `AKIA`, PEM) ever leaks; never raises; `_emit_provider_limit_marker` swallows a forced internal error and prints no malformed line |
| Secret-safety (direct, poison body w/ keys + tokens + ANSI + prompt content) | ✅ leaks nothing — only fixed enum fields + normalized UTC timestamp; injection `evil\nINJECT` coerced to `provider=unknown` |

### End-to-end (real process boundary, not capsys)

A child-process harness drove `run_agentic_task` through the documented `_run_with_provider` seam and observed the marker on a **real stdout fd**:

- **Credential-limit under `quiet=True`** → marker emitted once, `success=False`. ✅
- **Success-path multi-provider** (the sharpest regression vector) → provider A credential-limited emits the marker, provider B **succeeds** → `success=True provider=openai` with `PDD_PROVIDER_LIMIT provider=anthropic …` still on the **exit-0** run's stdout, count = 1. ✅ Matches the documented *per-provider detection, not job-failure* contract.
- **429 that recovers on retry** → **no marker**, `success=True`. ✅

### CLI-boundary survival (does it reach the fd cloud actually scrapes?)

Confirmed the marker survives the outermost CLI process, not just `run_agentic_task`'s frame:
- No `redirect_stdout` anywhere in the agentic path.
- `pdd/core/cli.py:506` installs `OutputCapture` on `sys.stdout`, but it's a genuine **tee** — `write()` unconditionally writes to `original_stream` first, then mirrors to a swallow-safe buffer; `flush()` flushes the real stream first.
- Verified by code-read **and** empirically (installed `OutputCapture` as `sys.stdout`, printed a marker, observed it on the real subprocess fd).
- This is the **same tee channel** through which cloud already scrapes `api_error_status:429` / `credential-limit` today → producer chain proven end-to-end.

### Scope / honest caveat
This proves the **producer**. It does **not** prove the cloud round-trip: the consumer (`pdd_cloud#2049`) is gated behind `PDD_EXECUTOR_LIMIT_QUEUE_ENABLED=False`, is prompt-only, and still parses the older `PDD_PROVIDER_FAILURE` format (`PDD_PROVIDER_LIMIT` has 0 live hits in pdd_cloud). The benefit is latent until that PR un-gates and cloud bumps the pinned `PDD_SOURCE_SHA`. Minor note for the consumer team: a run where two providers are both limited emits one marker per provider (each with its own `reset_at`) — that's the documented "combine with exit status" contract, not a defect.

`MERGEABLE` / `CLEAN`. 🤖 Verification by Claude Code.

PR #1559, opened 2026-06-11
fix(conformance): compare semantic callable contracts across all sync gates (#1558)

### LINKED ISSUE #1558: Make PDD sync conformance compare semantic callable contracts across all gates
## Problem

Issue #1545 and PR #1546 exposed a deeper UX problem in `pdd sync`: conformance is enforced by multiple gates, but they do not share one semantic callable-contract model.

PR #1546 fixed the narrow `<pdd-interface>` case where a prompt declares a literal default and generated code factors the same value into a module constant:

```py
# prompt interface
max_chars: int = 25000

# generated code
_COMMENT_MAX_CHARS = 25000
def f(max_chars: int = _COMMENT_MAX_CHARS): ...
```

That is semantically equivalent for callers, but raw string comparison reports drift. However, the PR does not fully solve the sync user experience because the public-surface regression gate can still reject the same literal-to-constant refactor after architecture conformance accepts it.

From a user perspective, this still looks like `pdd sync` failing for a weird conformance reason.

## Required approach

Make conformance use one semantic callable-contract system across all gates instead of patching one checker.

1. Build safe module default-symbol tables for both old and generated code.
   - Resolve only safe, static constants.
   - Use generated-code symbols for prompt `<pdd-interface>` -> generated-code checks.
   - Use old-code and new-code symbols for public-surface existing-code -> generated-code checks.

2. Replace raw signature-string default comparison with parsed callable contracts.
   - Track parameter name, kind, annotation, default expression, return annotation, and binding kind.
   - Compare defaults semantically, resolving each side against its own module symbol table.
   - Keep exact or conservative behavior where the contract cannot be parsed safely.

3. Preserve strict behavior for real contract breaks.
   - Missing declared functions/parameters still fail.
   - Dropped defaults still fail.
   - Provably different defaults still fail.
   - Dynamic or unresolvable defaults should fail closed with a clear diagnostic.
   - Public symbol removals still fail unless explicitly allowed.

4. Improve retry behavior for deterministic conformance failures.
   - If a sync retry sees the same semantic conformance failure repeatedly, stop quickly and print a clear checker diagnostic.
   - Do not burn repeated generation attempts for a deterministic checker mismatch.

## Acceptance criteria

- Existing literal default -> generated same-module immutable constant default passes both `<pdd-interface>` conformance and public-surface regression.
- Prompt literal default -> generated same-module immutable constant default passes architecture conformance.
- Existing same-module immutable constant default -> generated equivalent literal default passes public-surface regression.
- Existing constant value and generated different constant value still fail.
- Dynamic defaults, imported values, calls, mutable module constants, and unresolvable expressions fail closed with clear messages.
- Missing functions like `extract_step_report` and removed public fields like `AgenticTaskResult.usage` still fail.
- Full `code_generator_main`/`pdd sync` regression tests cover the multi-gate path, not only helper functions.

## Related

- Replaces #1545 with a broader design scope.
- PR #1546 was closed because it fixed only the prompt-interface gate and left public-surface false positives possible.

Serhan-Asad:
## Problem

`pdd sync` enforces parameter-default conformance with **raw signature-string comparison** in two independent gates, so a backward-compatible refactor that replaces a literal default with an equivalent same-module constant is falsely reported as drift:

```py
# before / prompt              # after / generated
def f(max_chars=25000): ...    _LIMIT = 25000
                               def f(max_chars=_LIMIT): ...
```

This is identical for every caller, yet both gates rejected it:

- the `<pdd-interface>` / architecture-conformance gate (`_verify_pdd_interface_signatures`) compared `declared_default != actual_default`, and
- the public-surface regression gate (`_verify_public_surface_regression`, via `signature_entries_compatible` → `_existing_param_compatible`) compared `old.default != new.default`.

PR #1546 fixed only the first gate; the public-surface gate could still reject the same refactor *after* the interface gate accepted it. To the user this still looks like `pdd sync` failing for a weird conformance reason.

## Fix

Route **both** gates through one semantic callable-contract model in `pdd/interface_semantics.py`:

- **One comparator.** `compare_default_sources` folds each default's AST into a type-tagged, hashable value and never executes code, imports, or evaluates arbitrary expressions (only safe literals, unary `+`/`-`/`~` on numbers, literal containers, and same-module constants bound exactly once to an immutable literal). It returns a tri-state `DefaultCompatibility` (`COMPATIBLE` / `INCOMPATIBLE` / `UNKNOWN`).
- **Per-side symbol tables.** `compare_default_sources(left, right, symbols=None, *, left_symbols=None, right_symbols=None)` resolves each side against its own module namespace. The public-surface gate builds `build_module_default_symbols(existing_code)` and `build_module_default_symbols(generated_code)` and passes them per-side; the interface gate passes one generated-code table for both sides (prompt and code describe the same generated module).
- **Threaded keyword-only.** `old_symbols` / `new_symbols` flow through `signature_entries_compatible` → `callable_contracts_compatible` → `_existing_param_compatible` as keyword-only optional params, so existing callers are unaffected.
- **Strict where it matters.** `_existing_param_compatible` now uses the tri-state verdict instead of a raw `!=` short-circuit, so a same-named constant resolving to a *different value* across the two module versions (`_LIMIT = 25000` → `_LIMIT = 5000`, identical signature text) is caught as a provable break — a false negative the old string compare missed — while two identical unresolvable defaults are never falsely flagged. Dropped defaults, missing functions, removed public fields, and dynamic/imported/mutable-container defaults all still fail closed.

The architecture.json gate (symbol-existence + camelCase only) compares no defaults and needs no change. The retry loop already breaks on an identical failure signature; removing the false-positive failures at their source is what stops the wasted regeneration attempts the issue describes, so no retry machinery is added.

`code_generator_main_python.prompt` is updated so a future sync of the generator regenerates the semantic comparison rather than reverting to string equality.

## Tests

- `tests/test_issue_1558_semantic_contracts.py` — gate-level acceptance matrix driving both `_verify_public_surface_regression` and `_verify_pdd_interface_signatures`: literal↔constant passes (both gates, both directions), different-value fails, dynamic/mutable/dropped-default/missing-function/removed-field fail closed.
- `tests/test_interface_semantics.py` — comparator + per-side unit tests (digit separators, signed zero, mutable containers, star-import poisoning, single-binding trust, alias non-resolution, same-name-different-value).

Validation (run with `PYTHONPATH=.` since `pdd` is an installed namespace package):

```
python -m pytest tests/test_issue_1558_semantic_contracts.py tests/test_interface_semantics.py \
  tests/test_code_generator_main.py tests/test_agentic_sync_runner.py \
  tests/test_one_session_sync.py tests/test_sync_orchestration.py -q
```

All green; no regressions across the conformance/sync blast radius.

Closes #1558.


Serhan-Asad:
Follow-up hardening (two commits on top of the original change), each red-first with gate-level + unit coverage; full conformance/sync blast-radius green (945 passed).

**1. Fail closed when a default resolves on only one side.** The public-surface gate's `UNKNOWN` branch passed a default whenever the two source strings matched textually. That let an asymmetric resolution slip through: the *same* `_LIMIT` text where the existing module leaves it opaque (`_LIMIT = load_limit()`, or `from cfg import _LIMIT`) but the generated module pins it to a literal (`_LIMIT = 5000`). Exactly one side being a known literal while the other is unresolvable is positive evidence the effective default may have changed, so matching text no longer waves it through — it now matches the change's own stated contract that dynamic/imported defaults fail closed. Only an `UNKNOWN` where *both* sides are opaque with identical text (an unchanged `x=helper()`) stays compatible.

Tradeoff, disclosed: this also flags the benign case where a default's binding *form* flips (inline literal ↔ imported name) across a sync while the value is unchanged — the value can't be proven equal once one side is opaque. It only fires when the binding form actually changes, and the conservative read is the right bias for a regression gate.

**2. Count only module-scope bindings when trusting a constant.** `build_module_default_symbols` disqualified a constant if its name was bound *anywhere* in the module, so an unrelated function-local `def g(): _LIMIT = 1` defeated the exact literal→constant refactor this gate is meant to accept. Counting is now module-scope-aware (recurses through module-level control flow, not into def/class/lambda/comprehension bodies). The two static ways a nested scope can still rebind a module global — `global X` + assign, and a leaking walrus `X := ...` — are caught by a conservative whole-tree scan, and the walk recurses by default on any unfamiliar node, so anything it can't prove local still fails closed.

Serhan-Asad:
Tightened the module-scope binding analysis so the "fail closed" binder no longer over-fires (commit `ac13825c4`). Red-first, full blast-radius green (950 passed).

**Fixed — both were provably-wrong over-broad heuristics, not safety relaxations:**

1. **Function-local walrus.** A walrus `X := ...` now disqualifies the constant only when evaluated at *module* scope. Comprehension scopes are transparent to walrus (a walrus in a module-level comprehension still disqualifies, and is tested), but one inside a `def`/`lambda` body binds a local and no longer poisons the module constant.

2. **Read-only `global`.** `global X` now disqualifies only when the *same function also assigns* X. A function that declares `global X` but only reads it rebinds nothing and is ignored. (Over-attributing a nested function's store only over-evicts toward UNKNOWN — safe, never a false match.)

Genuine module rebinds — `global`+assign, module-level-comprehension walrus, def-default walrus — still fail closed; all retained in the disqualify matrix.

**Declined, with reasoning — `_LIMIT = 25000; def f(x=_LIMIT); _LIMIT = 5000`:** you're right that defaults are definition-time, so `f`'s default is determinately `25000`. But the single per-module name→value table cannot carry a *position-dependent* value (the same `_LIMIT` would have to resolve to `25000` at `f`'s def site and `5000` after it). Capturing that needs position/flow-aware resolution — a disproportionate redesign for a shape a generator essentially never emits (and reassigning a constant around the def that defaults to it is a code smell arguably worth flagging). It fails closed, never a false match, and is now covered by an explicit `test_multiply_bound_constant_stays_conservative` documenting the limitation. The "reassign *before* the def" variant is the same class and stays disqualified too.

Serhan-Asad:
Self-review caught one more, now fixed (commit `a87ccf51b`): scoping the `global` rebind check to functions had narrowed it too far — `global X` is also valid in a **class body**, where an assignment to the global-declared name writes the module global. That let `class C: global X; X = 1` leave `X` trusted (a false match, since module `X` is rebound). The scan now includes class bodies, so it fails closed again; a plain class attribute with no `global` (`class C: _LIMIT = 1`) still does not disqualify the module constant. Added to the still-disqualify matrix; full blast-radius green (951 passed).

Serhan-Asad:
**Finding 2 — fixed** (commit `48b6027a8`). A nested function's local store was being attributed to an outer scope's `global` declaration (`ast.walk(scope)` descended into nested functions). Now each scope's `global` declarations are matched only against stores in its **own body**, stopping at nested def/async-def/class/lambda boundaries; a nested scope needs its own `global` and is matched when the walk reaches it. Comprehensions are still descended into, since a walrus there binds the containing scope honoring its `global` (PEP 572) — so `global X` + a comprehension walrus stays disqualified. Genuine rebinds (own-scope / class-body / nested-own `global`+assign, module-level and global-honored walrus) remain closed; added to the disqualify matrix. Full blast-radius green (955).

**Findings 1 and 3 — held, by design.** Neither is a scope bug; they are the two decidability boundaries of *static* default resolution:

- **1, `_LIMIT=25000; def f(x=_LIMIT); _LIMIT=5000`** — correct that Python captures `25000` at definition time. But a single per-module name→value table structurally cannot carry a *position-dependent* value (the same `_LIMIT` would need to resolve to `25000` at `f`'s def and `5000` after it). Resolving it requires flow/position-aware analysis — disproportionate for a shape a generator essentially never emits — and it fails **closed** (UNKNOWN), never a false match. Documented in `test_multiply_bound_constant_stays_conservative`.

- **3, `globals()["_LIMIT"]=5000`** — this is the comparator's foundational contract: it **never executes code**. Dynamic namespace mutation (`globals()[…]=`, `setattr`, `exec`) is resolvable only by running the module; it is already disclaimed in `build_module_default_symbols`'s docstring and has been uncovered since the original change (not a regression). A `globals()["literal"]=` matcher would miss `g=globals(); g[k]=`, `setattr`, and `exec` — implying a completeness it cannot have — so I'm keeping it an honest documented boundary rather than a leaky heuristic that invites a false sense of soundness.

Precise invariant the implementation now holds: **no false match among statically-resolvable bindings; dynamic namespace mutation is explicitly outside the static contract.**

Issue #624, opened 2026-03-02
PDD applies auth decorators at wrong granularity


Serhan-Asad:

Serhan-Asad:
3/2/2026

gltanaka:
Resolved in v0.0.162 (commit 6b359d78f). Added TypeScript/Python import validation functions in sync_orchestration.py and architecture generation blocking on validation failures. Covered by multiple test files.

PR #1575, opened 2026-06-13
fix(checkup): honor out-of-scope/non-blocking flags in Step 7 targeted-PR gate (#1574)

### LINKED ISSUE #1574: Checkup Step 7 gate blocks targeted PRs on out-of-scope/non-blocking criticals when no free-text reason is present
## Summary
In **targeted PR mode**, `pdd checkup`'s Step 7 gate (`_step7_passed` → `_step7_unfixed_critical_blocks_targeted_pr` in `pdd/agentic_checkup_orchestrator.py`) fails a PR on an unfixed `critical` finding that the verifier itself marked **out-of-scope / non-blocking**, when the finding does not also carry a free-text `*_reason` field. The exclusion is coupled to `_step7_nonblocking_reason(issue)` being non-empty, but the Step 7 verify prompt emits `blocking`/`scope` flags **without** a separate `non_blocking_reason` / `out_of_scope_reason` / `scope_reason` / `reason` field — so the `and reason` guard defeats the exclusion.

## Impact
A PR that only changes `README.md` is failed by a **pre-existing, repo-wide** TypeScript build break (`TS18003: frontend/src/ missing`) that the PR neither introduced nor can fix. Step 7's own verdict says `success: true`, `issue_aligned: true`, "PR-scoped verification: all clear", and tags the finding `scope: out-of-scope`, `blocking: false` — yet the gate returns `Step 7 reported unfixed critical issues`, skips the push, and the Final Gate reports `layer1_status: failed`. Observed on a real staging checkup run (promptdriven/test_repo_2 #510).

## Reproduction (against current `main`)
```python
from pdd.agentic_checkup_orchestrator import _step7_passed
issue = {"module":"frontend","severity":"critical","fixed":False,
         "scope":"out-of-scope","blocking":False,
         "description":"TS18003: frontend/src/ missing (pre-existing)."}
payload = {"success":True,"issue_aligned":True,"issues":[issue],"changed_files":["README.md"]}
print(_step7_passed("report\n"+__import__("json").dumps(payload),
                    pr_mode=True, has_issue=True, pr_test_scope="targeted"))
# -> (False, 'Step 7 reported unfixed critical issues: frontend: ...')   # BUG: should pass
```
`_step7_nonblocking_reason(issue)` returns `''`, so `_step7_unfixed_critical_blocks_targeted_pr` returns `True` despite `blocking is False` and `scope == "out-of-scope"`.

## Root cause
`_step7_unfixed_critical_blocks_targeted_pr` requires `... and reason` on each exclusion branch:
```python
if blocking is False and reason:        return False
if in_scope is False and reason:        return False
if scope in explicit_nonblocking_scopes and reason:  return False
```
The explicit `blocking: false` / `scope: out-of-scope` flags are themselves the authoritative non-blocking signal; requiring an additional free-text reason is too strict and mismatches the prompt's output contract.

## Proposed fix
Treat an explicit non-blocking flag/scope as sufficient on its own (keep the structured reason as an *additional* signal, not a precondition). In targeted PR mode the full-suite truth comes from the GitHub-checks gate, so out-of-scope criticals are informational by design.

## Acceptance
- The reproduction above returns `(True, "")`.
- A genuinely in-scope/blocking unfixed critical still fails the gate.
- Regression test added.

DianaTao:
Fixes #1574.

## Problem
In PR mode the Step 7 gate (`_step7_passed` → `_step7_unfixed_critical_blocks_targeted_pr`) failed a PR on an unfixed `critical` finding the verifier itself tagged **out-of-scope / non-blocking**, whenever the finding lacked a free-text `*_reason` field. Each exclusion branch was gated on `... and reason` (`_step7_nonblocking_reason(issue)` non-empty), but the Step 7 verify prompt emits `blocking: false` / `scope: out-of-scope` **without** a separate reason — so the guard never fired.

Real-world impact: a README-only PR was failed by a **pre-existing, repo-wide** `TS18003: frontend/src/ missing` TypeScript build break it neither introduced nor can fix. Step 7's own verdict said `success: true`, `issue_aligned: true`, *"PR-scoped verification: all clear"*, with the finding tagged `scope: out-of-scope`, `blocking: false` — yet the gate returned `Step 7 reported unfixed critical issues` and the Final Gate reported `layer1_status: failed`.

## Fix
Treat an **explicit verifier signal as authoritative on its own** — `blocking is False`, `in_scope is False`, or a non-blocking `scope` value each independently mark an unfixed critical as non-blocking. The free-text reason is corroboration, never a precondition.

The gate is also **broadened from targeted-only to all PR modes**: `_step7_unfixed_critical_blocks_targeted_pr` is renamed to `_step7_unfixed_critical_blocks_pr` and `_step7_passed` now applies it whenever `pr_mode` is set (no longer just `pr_test_scope == "targeted"`). The final Step 7 verdict owns the PR-introduced-vs-baseline distinction even when a full suite ran locally, so an explicitly non-blocking critical is informational in both modes; everything else still blocks.

## Reviewer follow-ups addressed
- **[P2] Generic `reason` no longer bypasses in-scope criticals.** `_step7_nonblocking_reason` previously matched a generic `reason` key, so an in-scope PR critical carrying ordinary explanatory text (e.g. `scope: "pr-diff"` + `reason: "introduced token leak"`) could slip through. The helper now honors only explicitly non-blocking reason fields (`non_blocking_reason` / `out_of_scope_reason` / `scope_reason`); a generic `reason` cannot bypass the gate. Covered by a regression test verified to fail pre-fix.
- **[P3] Restored the runnable orchestrator example.** `context/agentic_checkup_orchestrator_example.py` had been truncated mid-`main()` and no longer invoked the orchestrator. It now patches the per-step LLM runner, calls `run_agentic_checkup_orchestrator(...)` offline (`use_github_state=False`, `no_fix=True`) against the mock repo, and restores the `if __name__ == "__main__": main()` guard. Verified to run end-to-end (exit 0).
- **[Security] No clear-text logging in the example.** CodeQL flagged the example for printing the orchestrator's returned `message` / `cost` / `model` as clear text (the return tuple is taint-tracked; `message` can echo repo output). Those values are no longer echoed — the example prints only a static pass/fail summary derived from the boolean result, with a comment pointing to routing real values through a logging/redaction layer.

## Behavior preserved
- A genuinely **in-scope / blocking** unfixed critical still fails the gate (targeted **and** full scope).
- A critical with **no** non-blocking signal (no flag, scope, or explicit non-blocking reason) still blocks.

## Tests
Gate regression coverage in `tests/test_agentic_checkup_orchestrator.py` and `tests/test_checkup_pr_mode.py`:
- out-of-scope critical with flags only (no `*_reason`) → PR passes (targeted and full scope)
- `blocking: false` flag alone → PR passes
- in-scope `blocking: true` critical → still fails
- in-scope critical with only a **generic** `reason` → still fails (P2 regression)
- changed-file critical → still fails

The gate and `test_checkup_pr_mode.py` suites pass, and the exact staging reproduction from #1574 now returns `(True, "")`.


DianaTao:
Jun 18

gltanaka:
Required changes before merge.

The #1574 fix is needed, but this PR is currently broader than the issue requires and creates a fail-open path in the Step 7 gate.

1. Keep the non-blocking critical carveout scoped to the contract that justifies it, or add stricter corroboration.

The issue is specifically targeted PR mode: a README-only targeted run was blocked by a baseline critical that Step 7 marked `scope: "out-of-scope"` and `blocking: false`. This PR removes the `pr_test_scope` distinction and makes `blocking is False` alone authoritative for every PR mode.

That means this now passes even when the critical is in the PR diff and there is no out-of-scope signal:

```python
issue = {
    "module": "auth",
    "file": "auth.py",
    "severity": "critical",
    "fixed": False,
    "blocking": False,
    "description": "PR-introduced token leak",
}
payload = {"success": True, "issue_aligned": True, "issues": [issue], "changed_files": ["auth.py"]}
_step7_passed("report\n" + json.dumps(payload), pr_mode=True, has_issue=True, pr_test_scope="targeted")
# current PR: (True, "")
```

The Step 7 verify prompt still says full PR mode is the comprehensive gate, and only targeted PR scope gets the explicit unrelated baseline/project-wide carveout. Please either:

- keep the #1574 behavior targeted-only, or
- require the bypass to be corroborated by an out-of-scope signal (`scope` / `verification_scope` / `pr_scope`, `in_scope: false`, or an explicit non-blocking reason) and fail closed when the finding's `file`/`module` touches `changed_files` or the scope is PR/in-scope.

Please add a regression for a changed-file critical with `blocking: false` but no out-of-scope signal; it should still block.

2. The restored example is not actually offline.

`python context/agentic_checkup_orchestrator_example.py` exits 0, but it still attempts GitHub issue comment calls despite `use_github_state=False`:

```text
Warning: Failed to post step comment ... Could not resolve to a Repository with the name 'example-owner/example-repo'
```

If the example is meant to be offline, it should patch/suppress the comment-posting helpers as well (`post_step_comment_once`, `post_step_comment`, and any PR comment helper reachable in the selected mode), or otherwise avoid fake GitHub issue metadata that triggers remote calls.

CI and the focused tests are green, but these two points should be fixed before this merges.

gltanaka (changes requested):
Requesting one more change before merge: please add true end-to-end verification evidence for the updated gate.

The unit/regression tests cover the helper behavior well, and the mocked offline example now runs cleanly, but this PR changes the top-level `pdd checkup --pr` pass/fail gate. The important risk is in the full runtime path: PR metadata/worktree setup, Step 7 output generation/parsing, `_step7_passed`, final gate reporting, and push/no-push behavior.

Please run and document a real E2E/staging check using the actual CLI against a PR-shaped repo, not only a mocked orchestrator example. Minimum useful evidence:

- command used, with secrets redacted
- PR/test scope used (`--test-scope targeted` for the #1574 case)
- Step 7 JSON showing an unfixed critical marked out-of-scope/non-blocking without a `*_reason`
- final gate/result showing the targeted #1574 case passes
- a negative E2E or CLI-level check showing a PR-scoped/changed-file critical still blocks, or clear evidence that the new regression test covers this exact final gate path

The original failure was observed in staging, so a rerun of that reproduction or an equivalent staging PR would be the best evidence.

DianaTao:
## ✅ Staging2 verification PASSED (definitive)

Re-ran the github-checks gate E2E with the **complete** fix deployed (after the first run revealed the label-checkup path was uncovered).

**Deployed under test (staging2):**
- pdd_cloud worker image `17c4f358c` — both fixes: `checkup_verifier` (#ac82822fd) **and** `checkup_failure_classifier` non-blocking downgrade (this PR's follow-up)
- pdd CLI pinned to `2cc9b70` (promptdriven/pdd#1575, issue-1574 Step-7 gate)
- `PDD_CHECKUP_GITHUB_CHECKS_BLOCKING=false`, command built with `--full-suite-source github-checks`

**Run:** promptdriven/test_repo_2#623 / PR #624 (docs-only, `full-suite` check **red**), executor job `G0AXJXB5DQHFIy1OPFN8`.

**Result (vs the earlier ❌ run on the same PR):**
- Final comment: `### ℹ️ PDD Checkup — GitHub checks failed — not attributable to this PR (non-blocking)`
- Job outcome: `### PDD Execution Successful` (not hard-blocked)
- Executor log: `[#2237] github-checks non-blocking ... treating checkup as passed`

So a Layer-1-passing PR with an unrelated/pre-existing red GitHub check is no longer hard-blocked. Default behavior stays blocking (flag-gated); prod is unchanged.

DianaTao:
## ✅ Cloud checkup validated against this CLI build (2cc9b70)

Cross-reference from the PDD Cloud side. The staging2 GitHub App worker was pinned to this PR's branch head (`2cc9b7016d26`, issue-1574 Step-7 gate) and run end-to-end together with the cloud-side github-checks non-blocking change (promptdriven/pdd_cloud#2240).

**Run:** promptdriven/test_repo_2#623 / PR #624 — docs-only PR, `full-suite` GitHub check **red** — checkup executed with `--final-gate --full-suite-source github-checks`.

**Result:** the final gate produced the github-checks failure report (`failure_category=github_checks_failed`, Layer 1 passed, Layer 2 skipped) as expected, and the cloud then treated it as **non-blocking**:
- `### ℹ️ PDD Checkup — GitHub checks failed — not attributable to this PR (non-blocking)`
- `### PDD Execution Successful` (job not hard-blocked)

So with this CLI build + the cloud flag, a Layer-1-passing docs PR is no longer blocked by an unrelated red GitHub check. Noting here because this branch is the pinned CLI under test for that validation.

gltanaka (approved):
The latest updates address my prior blocking concerns.

The Step 7 carveout is now scoped to targeted PR mode, requires a positive out-of-scope signal, and fails closed for PR-scoped / changed-file criticals. Full PR mode blocks unfixed criticals again. The resume cache now records `pr_test_scope`, which is the right follow-up since Step 5/Step 7 semantics differ by scope.

I also reviewed the staging2 evidence: this PR's CLI build was pinned in a real cloud checkup run against a docs-only PR with an unrelated red full-suite check, and the final outcome was non-blocking / execution successful. Combined with the focused regression coverage and green CI, this is sufficient to merge.

PR #1604, opened 2026-06-15
fix(clean-restart): fall back to a fresh branch when the issue branch is locked by another worktree

### LINKED ISSUE #1596: Clean restart should fall back when issue branch is locked by another worktree
## Problem

Clean restart can fail before useful work starts when the standard issue branch is checked out in another local worktree.

Observed from `promptdriven/pdd_cloud#2101` while retrying `promptdriven/pdd#1458`:

```text
Cannot perform clean restart: local branch 'change/issue-1458' could not be deleted (it is likely checked out in another worktree). Remove that worktree first, then retry.
```

The attempted runner-side fix in `promptdriven/pdd_cloud#2104` was reverted because it cannot solve the root problem: the PDD CLI clean-restart path computes `{command}/issue-{N}`, creates/deletes its own worktree/branch inside `_setup_worktree` in the agentic orchestrators, and ignores whatever fallback branch the GitHub App executor has checked out.

## Expected behavior

When `--clean-restart` is used and the standard branch (`{command}/issue-{N}`) is locked by another worktree, the CLI should not fail solely because of that lock. It should create a fresh unique fallback branch/worktree from the requested base branch and continue.

Example fallback branch:

```text
change/issue-1458-job-zMxRhCSl
```

## Proposed approach

Implement the fallback where the branch deletion/reuse actually happens: the PDD CLI worktree setup path, likely `_setup_worktree` in the agentic change/bug/test orchestrators.

Suggested behavior:

1. Keep the normal branch path first: `{command}/issue-{N}`.
2. During clean restart, if deleting or reusing that branch fails because Git reports it is checked out/used by another worktree, detect that specific condition.
3. Generate a unique fallback branch name, preferably `{command}/issue-{N}-job-{job_id[:8]}` when a job id is available, or a timestamp/unique suffix otherwise.
4. Create the fallback from the requested base branch, usually `origin/main`.
5. Continue the run using the fallback branch/worktree.
6. Log/comment clearly that a fallback branch was used and that the stale worktree was not modified.

Do not delete user-owned worktrees or worktrees with uncommitted changes.

## Acceptance criteria

- `pdd change`, `pdd bug`, `pdd fix`, and `pdd test` clean-restart runs do not fail solely because the default issue branch is checked out elsewhere.
- A locked default branch causes a unique fallback branch/worktree to be used.
- The fallback branch is based on the requested base branch.
- The stale locked worktree is not modified or removed.
- Job output/comment explains that a fallback clean-restart branch was used.
- Existing non-clean-restart behavior is unchanged.
- Tests cover the locked-worktree branch condition and confirm the fallback branch is selected.

## References

- Original app-runner issue: https://github.com/promptdriven/pdd_cloud/issues/2101
- No-op/superseded app-runner PR: https://github.com/promptdriven/pdd_cloud/pull/2104

Serhan-Asad:
## Problem

Fixes #1596.

Under `pdd change/bug/test --clean-restart`, `_setup_worktree` computes the canonical issue branch (`change/issue-{N}`, `fix/issue-{N}`, or `test/issue-{N}`) and tries to `git branch -D` it so the worktree can be rebased on `origin/main` (the Req-15 fresh-base guarantee). When that branch is checked out in **another live worktree** — e.g. the GitHub App executor's own checkout — the delete fails and the whole run aborts before any useful work:

```
Cannot perform clean restart: local branch 'change/issue-1458' could not be deleted
(it is likely checked out in another worktree). Remove that worktree first, then retry.
```

Reusing the locked branch is not an option under clean restart (it would drag the previous run's commits back in), which is why the code aborted. The runner-side attempt (pdd_cloud#2104) was reverted because the root cause lives in the CLI worktree-setup path.

## Fix

When a clean restart finds the canonical branch locked by another worktree, create a **fresh, uniquely-named fallback branch from the requested base** and continue, instead of aborting:

1. **Prune + retry first.** On `git branch -D` failure under `clean_restart`, run `git worktree prune` and retry the delete — the common cause is a *stale* worktree registration, which this resolves cleanly (no fallback needed, canonical branch reused as before).
2. **Detect a genuine lock.** Only if the delete still fails *and* `git worktree list --porcelain` shows the branch held by a **different, live** worktree (path exists, not `prunable`, not our own target path) do we fall back. Any other delete failure keeps the original hard-fail error.
3. **Fallback branch.** Create `{cmd}/issue-{N}-job-{slug}` from `origin/main` (`slug` = first 8 chars of `JOB_ID`/`PDD_JOB_ID` so the name — and therefore the PR — is stable across retries of the same job; otherwise a random token). The locked worktree is never touched. We also fetch the fallback's own remote-tracking ref so Step 13's `git push --force-with-lease` works on a same-job retry where the remote fallback branch already exists.
4. **Thread the real head branch downstream.** The PR-creation steps previously hardcoded the canonical branch in `git push` / `gh pr list/create --head`. They now use a `{head_branch}` template variable derived from the worktree's actual HEAD (defaulting to the canonical name → **byte-identical rendering when no fallback occurs**). The change orchestrator's early existing-PR guard also matches `…-job-*` fallback heads so a later normal run won't open a duplicate PR.

A `[bold cyan]` notice is logged naming the locked worktree and the fallback branch.

### Scope
`pdd change`, `pdd bug`, `pdd test` (the three orchestrators with the `_setup_worktree` branch-delete path). `pdd fix` (agentic_e2e_fix) is **structurally exempt** — under clean restart it leaves stale worktrees alone and works from the repo root, never deleting an issue branch. `pdd checkup` already has its own deletion-failure recovery and is out of scope.

## Tests

Real-git regression tests (an actual second live worktree holds the branch — not mocked) for all three orchestrators:
- locked canonical branch → fallback worktree on `{cmd}/issue-{N}-job-*`, based on `origin/main`, **no commit leak** from the locked branch, locked worktree untouched, `JOB_ID`-derived slug;
- same-job retry where the remote fallback already exists but no local tracking ref → setup fetches the fallback tracking ref (so `--force-with-lease` succeeds);
- non-clean-restart delete-failure still hard-fails (behavior unchanged);
- prompt-render tests: identical output when `head_branch == canonical`; fallback head pushed/PR'd with no bare canonical branch token;
- existing-PR guard matches a `…-job-*` fallback head and rejects unrelated prefix heads.

The 3 orchestrator suites + `test_agentic_common.py` pass (one pre-existing, unrelated `PDD_PATH`-env failure in `TestIssue1080MonorepoCwd` is present on clean `main` too).

## Residual limitation (follow-up, not a regression)

`pdd sync`'s `_augment_architecture_from_pr_branch` reads `origin/change/issue-{N}` to pull new-module architecture entries from an open change PR. A fallback-branch change PR (`…-job-*`) won't be discovered there, so brand-new modules created on a fallback branch could be missed by a pre-merge global sync. This is **strictly better than today** (the locked path currently aborts → no PR and no sync at all) and only affects the narrow `change` + new-module + pre-merge-sync combination; it's tracked as a follow-up in #1609 rather than expanding this PR's scope.


Serhan-Asad:
## Verification summary

Verified end-to-end in an isolated worktree at this PR's head (already rebased on `main`, so this is the merged state). Your local checkouts were untouched.

- **Affected suites** (change/bug/test orchestrators + `agentic_common`): all green — **981 passed, 0 real failures**. The lone `TestIssue1080MonorepoCwd` failure is a local `PDD_PATH`-env artifact: it fails identically on clean `main` and passes once `PDD_PATH` is set.
- **Independent real-git reproduction** (written separately from this PR's own fixtures), **36/36** across all three orchestrators:
  - genuine lock → fresh fallback branch off `origin/main`, no commit leak from the locked branch, locked worktree left untouched, `JOB_ID`-derived slug;
  - stale/prunable registration → `git worktree prune` + retry → **canonical branch reused, no needless fallback**;
  - the worktree's actual HEAD is threaded as `head_branch` into the PR step, so the fallback branch is what gets pushed / PR'd.
- **Byte-identical** push/PR rendering when no fallback occurs (the common path) — test-backed.
- All CI checks green.

## On the fallback-PR behavior (by design)

When the canonical branch is locked under clean restart, this opens a PR on the fresh `…-job-*` fallback head rather than updating the prior canonical PR. **This is intended:** clean restart's contract is to discard the prior attempt and skip the existing-PR guard, so a fresh branch + fresh PR is the correct outcome; the superseded canonical PR is left for maintainer cleanup. Calling it out explicitly so it isn't mistaken for a duplicate-PR bug during review.

PR #1676, opened 2026-06-19
fix(agentic): resolve child sync --context per module cwd

### LINKED ISSUE #1675: Carry a resolved sync unit through agentic sync
## Problem

Agentic sync currently resolves some per-module information, but it does not carry one complete project/context identity through the whole sync pipeline.

Today the runner has per-module cwd data:

```text
module_cwds[module]
module_targets[module]
```

but child command construction still forwards one global context:

```text
sync_options["context"]
```

That lets PDD spawn a child command where `cwd` and `--context` come from different `.pddrc` files.

Concrete current failure seen in `promptdriven/pdd_cloud`:

```text
cwd = extensions/github_pdd_app
--context extensions-github_pdd_app
```

`extensions-github_pdd_app` exists only in the root `.pddrc`; it does not exist in `extensions/github_pdd_app/.pddrc`, so child sync fails with:

```text
Unknown context 'extensions-github_pdd_app'
```

This still appears on recent hosted sync runs, including `promptdriven/pdd_cloud#2093` on June 19, 2026, after a PDD-created PR. The same class is also described in `promptdriven/pdd_cloud#2250`.

## Why a narrow context patch is not enough

Adding an alias context to a nested `.pddrc`, or always forcing sync to run from repo root, would only hide this specific symptom.

It can break the opposite case:

```text
cwd = repo root
--context pdd_executor_pkg
```

because `pdd_executor_pkg` exists in `extensions/github_pdd_app/.pddrc`, not in the root `.pddrc`.

The real invariant should be:

```text
same cwd
same .pddrc
same context
same resolved paths
```

## Proposed fix

Introduce a resolved sync unit and pass it through the full agentic/global/durable sync flow.

Suggested shape:

```python
@dataclass(frozen=True)
class ResolvedSyncUnit:
    key: str                 # scheduler/display key
    target_basename: str      # actual basename passed to pdd sync
    cwd: Path                 # project root that owns the unit
    pddrc_path: Path | None
    context: str | None       # valid only for pddrc_path
    prompts_dir: Path
    architecture_path: Path | None
    generate_output_path: Path | None
    test_output_path: Path | None
```

The exact fields can change, but the important point is that all sync stages use the same resolved unit instead of recomputing pieces independently.

Update these flows to consume the resolved unit:

- dry-run validation
- already-synced fingerprint filtering
- dependency/global sync analysis
- `AsyncSyncRunner`
- `DurableSyncRunner`
- child `pdd sync` command building
- progress/reproduce-local diagnostics
- metadata/fingerprint path handling where applicable

Child command construction should become per-unit:

```text
cwd = unit.cwd
--context = unit.context, only if unit.context exists in unit.pddrc_path
sync target = unit.target_basename
```

Before spawning a child process, validate:

```text
if unit.context and unit.context not in contexts from unit.cwd/.pddrc:
    fail before running child pdd sync, with an actionable diagnostic
```

Do not silently pass a parent context into a nested cwd.

## Current code smell

In `agentic_sync_runner.py`, `_build_command()` uses the global context:

```python
context_override = self.sync_options.get("context")
if context_override:
    cmd.extend(["--context", str(context_override)])
```

but `_run_attempt()` uses per-module cwd:

```python
cwd_value = self.module_cwds.get(basename, str(self.project_root))
```

That split is the root cause.

## Acceptance criteria

- Agentic sync resolves each target into one `ResolvedSyncUnit` before execution.
- Child `pdd sync` commands use `unit.cwd`, `unit.context`, and `unit.target_basename` together.
- Parent/root contexts are never passed into nested project cwd unless that context exists in the nested `.pddrc`.
- Nested contexts are never forced back to repo root unless that context exists in the root `.pddrc`.
- If no valid context exists, PDD omits `--context` or fails early with a clear resolver error, not a later child-process `Unknown context` error.
- Durable sync preserves the same unit identity when mapping module cwd into worktrees.
- Reproduce-local output includes both cwd and command, for example:

```bash
cd extensions/github_pdd_app
pdd --context pdd_executor_pkg sync pdd_codex
```

## Regression tests

Please add tests covering these cases:

1. Root `.pddrc` has `extensions-github_pdd_app`; nested `extensions/github_pdd_app/.pddrc` does not. Agentic sync for nested modules must not pass root context into nested cwd.
2. Nested context such as `pdd_executor_pkg` works from `extensions/github_pdd_app` and is not forced to repo root.
3. Root context such as `extensions-github_pdd_app` still works from repo root and is not forced into nested cwd.
4. `backend/functions` case where context names may match in root and nested `.pddrc`, but output/prompts/meta paths differ. Assert generated files and `.pdd/meta` land in the intended project paths.
5. Duplicate basename case, for example two projects with `worker_app` or `config`. Assert sync picks the one identified by path/project root, not just basename.
6. Global/no-argument sync uses the owning `.pddrc` and owning `architecture.json` for nested modules.
7. Durable sync carries the same unit into module worktrees.
8. A new PDD-created nested project with its own `.pddrc` works without adding special aliases to the parent or child config.

## Non-goals

- Do not fix this by adding `extensions-github_pdd_app` to the nested `.pddrc` only.
- Do not fix this by always running from repo root.
- Do not fix this by always running from the nested directory.

Those approaches hide one direction and break the other.

Serhan-Asad:
## Summary

**Part of #1675 (hotfix only — do NOT close #1675 on merge).** This stops the
immediate context leak. The full `ResolvedSyncUnit` work — one resolved identity
(cwd / .pddrc path / context / target basename / prompts_dir / architecture.json
/ output+meta paths) carried through the whole pipeline, deeper nested `.pddrc`
discovery, path/architecture-based duplicate-basename ownership,
default-only-nested safe-omit, and a unit shared by global + durable sync — is
tracked as a follow-up. #1675 stays open until that lands.

Agentic sync runs each child `pdd sync` in a **per-module cwd** (`module_cwds`)
but `AsyncSyncRunner._build_command()` forwarded a **single global** `--context`
from `sync_options["context"]`. When a module's cwd is a nested project whose
`.pddrc` does not define that context, the child dies with
`Unknown context '<ctx>'` (e.g. `cwd = extensions/github_pdd_app` +
`--context extensions-github_pdd_app`).

## Root cause

The split between `_build_command()` (global context) and `_run_attempt()`
(per-module cwd) in `pdd/agentic_sync_runner.py`.

## Fix (hotfix scope)

`AsyncSyncRunner._compute_module_context(basename)` resolves `--context` per
module against that module's own cwd `.pddrc`: forward when valid, substitute the
cwd's own context when not (preferring a threaded `module_contexts` map), else
fail early with an actionable resolver error. Gated on `basename in module_cwds`
so the emitted command is byte-identical for every currently-valid case.
Threaded through global + targeted + durable runners; reproduce-local now shows
`cd <cwd> && pdd --context <ctx> sync <target>`; `_resolve_module_sync_context`
passes the found `.pddrc` path; prompt specs updated.

> Note: the follow-up changes the fail-early behavior for default-only nested
> projects to a safe `--context` omit (per maintainer guidance on #1675).

## Tests / validation

19 new tests (root/nested forwarding, fail-early, legacy paths, duplicate
basename, same-named-context, threaded-map precedence, durable remap, helper,
pddrc_path). `400 passed` across the 3 sync test files; `124 passed`
construct_paths. Codex adversarial review: 4 rounds → MERGE-READY.

## PDD source-of-truth / draft note

`.prompt` specs updated so regeneration reproduces the fix. `.pdd/meta`
`prompt_hash` is a hash of the *preprocessed* prompt (not hand-reproducible), so
it needs a `pdd sync` of the 3 modules to consummate. **Draft** so `auto-heal`
(non-draft only) does not LLM-regenerate the ~2800-line file against stale meta;
this also defers the `unit-tests` job (local run is the proof). Mark ready after
a `pdd sync` meta refresh.


Serhan-Asad:
Folded into the combined mega PR #1683 — this per-module sync-context hotfix is included there (the branch carries the full ResolvedSyncUnit on top of it). Closing in favor of #1683.

Issue #1663, opened 2026-06-19
pdd sync: 'Public surface regression' hard-fails multi-module sync for removed imports/typing helpers (mirrors pdd_cloud#2193)


DianaTao:
## Summary
`pdd sync` aborts modules with **"Public surface regression"** when only *imported / typing helper* names change, not the real public API. Same root cause as the #2190 mirror; recorded separately because it reproduced on a second, independent set of modules (the checkup service split).

Mirror of cloud issue **promptdriven/pdd_cloud#2193** (checkup redesign: split `checkup_verifier` into services). Fix belongs in the `pdd` CLI sync surface check; tracked in both repos so we can fix on both ends.

## Where it happened
Cloud `pdd-issue` run on promptdriven/pdd_cloud#2193. Multiple modules paused: `checkup_command_builder`, `checkup_execution_service`, `checkup_verdict_service` failed.

## Error (verbatim)
```
│ Total time: 9.02s | Total cost: $0.2523 | Overall status: Failed │
Configuration error: Unknown context 'extensions-github_pdd_app'. Available
Configuration error: Unknown context 'extensions-github_pdd_app'. Available
Configuration error: Unknown context 'extensions-github_pdd_app'. Available
Configuration error: Unknown context 'extensions-github_pdd_app'. Available
An unexpected error occurred during sync for 'python': Public surface regression
│ python   │ Failed │    $0.2523 │ Public surface regression for               │
```
Resulting in:
```
### ❌ PDD Execution Failed
Execution failed: PDD command failed with exit code 1.
```

## Why this is a bug
Same as #2190 mirror: the surface-regression check treats removed imports / typing-only helpers as breaking public-API changes and **hard-fails the sync**, demanding a `BREAKING-CHANGE:` directive. Here it blocked **three** generated service modules at once, so the whole service-split sync could not land and the work was reimplemented manually (cloud PR #2219).

## Expected
- Exclude imported names / typing-only helpers from the public-surface comparison, or downgrade import-only deltas from a hard failure to a warning.
- A multi-module sync should not be fully blocked by import-only surface deltas.

## Related
- Same signature as promptdriven/pdd_cloud#2190 mirror (single root cause — likely one fix covers both).
- Different signature from promptdriven/pdd_cloud#2196 (agentic-sync input-too-large).

gltanaka:
Resolved by PR #1678 (9a3b2b9d), which excludes implementation imports from the public-surface gate and adds regression coverage. Closing.

PR #1678, opened 2026-06-19
fix(sync): exclude implementation imports from the public-surface gate

### LINKED ISSUE #1662: pdd sync: false-positive 'Public surface regression' blocks sync on import-only changes (mirrors pdd_cloud#2190)
## Summary
`pdd sync` aborts a module with **"Public surface regression"** when the only thing that changed is the set of *imported / typing helper* names — not the module's real public API. This blocked the cloud sync for the checkup verdict-engine work.

Mirror of cloud issue **promptdriven/pdd_cloud#2190** (checkup redesign). Filed here because the fix belongs in the `pdd` CLI sync surface check; tracked in both repos so we can fix on both ends.

## Where it happened
Cloud `pdd-issue` run on promptdriven/pdd_cloud#2190 (module `checkup_verdict_engine`). End of the PR/issue comment thread shows the sync failing.

## Error (verbatim)
```
│ Total time: 7.85s | Total cost: $0.2646 | Overall status: Failed │
Configuration error: Unknown context 'extensions-github_pdd_app'. Available
Configuration error: Unknown context 'extensions-github_pdd_app'. Available
Configuration error: Unknown context 'extensions-github_pdd_app'. Available
Configuration error: Unknown context 'extensions-github_pdd_app'. Available
An unexpected error occurred during sync for 'python': Public surface regression
│ python   │ Failed │    $0.2646 │ Public surface regression for               │

=== public surface regression ===
output: /tmp/pdd_job_.../checkup_verdict_engine.py
removed: Any, removed:Literal, removed:dataclass, removed:field, removed:json
signature_changed: <none>
pre surface size: 10
post surface size: 4

To allow this surface change, add a `BREAKING-CHANGE:` directive to the prompt bo[dy]
```
Resulting in:
```
### ❌ PDD Execution Failed
Execution failed: PDD command failed with exit code 1.
```

## Why this is a bug
- The "removed" symbols are `Any`, `Literal`, `dataclass`, `field`, `json` — these are **imports / typing helpers**, not genuine public API of the module. `signature_changed: <none>`.
- The surface-regression check counts these toward the public surface (pre 10 → post 4) and **fails the sync**, demanding a `BREAKING-CHANGE:` directive for a non-breaking change.
- This forces contributors to either annotate `BREAKING-CHANGE:` for cosmetic import changes or abandon the automated sync (which is what happened — the module was implemented manually instead).

## Expected
- Imported names and typing-only helpers should be **excluded from the public-surface comparison**, OR the regression should be a warning rather than a hard, sync-aborting failure for import-only deltas.

## Secondary observation
The repeated `Configuration error: Unknown context 'extensions-github_pdd_app'. Available …` lines suggest a context-resolution misconfiguration in the same run; may be unrelated but worth confirming it isn't masking/short-circuiting the surface computation.

## Related
- Same signature as the `pdd sync` failure on promptdriven/pdd_cloud#2193 (mirrored separately).
- Different signature from promptdriven/pdd_cloud#2196 (agentic-sync input-too-large).

Serhan-Asad:
## Problem

`pdd sync` hard-fails with `Public surface regression` when a regenerated Python module drops
import-only / typing-helper names such as `Any`, `Literal`, `dataclass`, `field`, `json`, `Dict`,
`GoogleAuthError`, `List`, or `Optional`. These are implementation details, not the module's public
API, yet the public-surface gate counted **every** non-underscore top-level import as public surface
(when the module had no `__all__`) and demanded a `BREAKING-CHANGE:` directive to remove them.

This blocked real cloud syncs and forced whole modules to be re-implemented by hand:

- promptdriven/pdd_cloud#2190 — `checkup_verdict_engine`: `removed: Any, Literal, dataclass, field, json`
- promptdriven/pdd_cloud#2193 — `checkup_command_builder`, `checkup_execution_service`, `checkup_verdict_service`
- promptdriven/pdd_cloud#2093 — `removed: Dict, GoogleAuthError, List, Optional`

Fixes #1662, #1663. Root-cause tracker: promptdriven/pdd_cloud#2256 / #2250 (category 1).

## Root cause

`_snapshot_public_surface` in `pdd/code_generator_main.py` recorded every non-underscore
`ast.Import` / `ast.ImportFrom` binding as public surface in the no-`__all__` fallback. So removing
an unused `from typing import Any` looked identical to removing a real exported symbol — the gate was
*more aggressive than Python's own re-export rules*.

## Fix

Treat an import as public surface **only when it is a deliberate re-export**, following Python's own
convention (PEP 484; the rule mypy enforces under `implicit_reexport = False`):

- a **redundant alias** — `import x as x` / `from m import y as y` (bound name == imported name), or
- a name listed in a readable **`__all__`** (already authoritative).

Plain imports (`import git`, `from typing import Any`) and *renaming* aliases (`from m import y as z`)
are implementation details. A new `_reexport_binding(alias)` helper encapsulates the redundant-alias
decision and is applied in **both** `_snapshot_public_surface` and `_snapshot_public_signatures`, so
the removed-symbol diff and the signature-drift diff agree on which imports are public (an asymmetry
here is the exact class of bug the #1012/#1015 work fixed). The `imported_names` collection used for
dataclass-base resolution is untouched.

### `__all__` interaction

A **clean readable `__all__`** is still authoritative (pre-existing behavior): a plain-import binding
listed in it is protected, and impl imports not listed are not surface. `__all__` is resolved in
**source order** (last write wins): a clean literal sets it, a later clean literal restores it, and any
unreadable write — computed `= sorted(...)`, augmented `+= [...]`, an in-place mutation
(`__all__.append(...)`, `__all__[i] = ...`, `del`), a conditional/looped rebind
(`if platform: __all__.append(...)`), or an import that binds `__all__` (`from exports import __all__`)
— makes it unresolvable, so the gate falls back to the heuristic.

In that fallback the import rule stays **lenient** (re-export-only): it deliberately does **not** switch
to a "treat every import as public" conservative mode, which would reintroduce the exact #2256 false
positive for the conditional-export pattern (an unrelated `from typing import Any` flagged on removal).
But the heuristic **does protect DEFINED symbols** added via the mutation — a function/class appended to
`__all__` is a non-underscore module-level name, so removing it is still blocked. (Protecting a
plain-*import* re-export listed *only* in an unresolvable `__all__` remains a documented limitation;
aliasing `x = __all__; x.append(...)` and `global __all__` inside a called function are also accepted
limitations — declare such a re-export in a clean `__all__` literal or use a redundant-alias import.)

### Other small correctness fixes in the touched paths

- Redundant aliases canonicalize to the plain signature marker, so normalizing
  `from pathlib import Path as Path` → `from pathlib import Path` under a stable `__all__` is not a
  phantom signature change.
- Relative-import re-exports record their level, so `from . import Foo as Foo` →
  `from .. import Foo as Foo` (same name, different package) is a detectable flip.

### What still fails the gate (protection preserved)

- Removing a real exported function / class / module constant.
- Removing an import declared in `__all__`, or a redundant-alias re-export (`import git as git`).
- Flipping a declared re-export to a same-named `def` / assignment
  (`from pathlib import Path as Path` → `def Path()`).

### Deliberate behavior change

A module that re-exports via a *plain* import (`import git`) is no longer auto-protected — it must opt
in with `__all__` or the redundant-alias form, which is also what static type checkers require.

## Prompt

This is a prompt-driven module, so `pdd/prompts/code_generator_main_python.prompt` is updated in
lockstep with the code (the `_snapshot_public_surface` / `_snapshot_public_signatures` /
`_reexport_binding` rules and the source-order `__all__` resolution) so a future regeneration
reproduces the new behavior instead of reverting it.

## Tests

New / updated in `tests/test_code_generator_main.py`: the verbatim #2256 import-only-removal cases;
explicit-re-export and clean-`__all__`-listed protection; redundant-alias signature canonicalization;
unresolvable-`__all__` import-leniency AND defined-symbol protection (a function appended via
`__all__.append`/`.extend`/`+=`/subscript/conditional is protected); source-order restore;
`from x import __all__` rebind; value-less annotation no-op; bare-annotation and relative-import edge
cases; and the updated import→def/assignment flip tests. Full `tests/test_code_generator_main.py` (361)
green; related public-surface suites (`test_issue_67*`, `test_issue_1558_semantic_contracts`,
`test_one_session_sync`, `test_interface_semantics`, `test_sync_orchestration`,
`test_agentic_sync_runner`) green.

Reviewed adversarially across multiple rounds with an independent reviewer.


Serhan-Asad:
## Verification summary

This change has been validated locally, in CI, by adversarial review, and on real staging cloud infrastructure (pinned to this PR's commit `d2b3f28`, **no merge required** — the cloud worker installs `git+…@<sha>`).

### Local + CI
- `pytest tests/test_code_generator_main.py` → **352 passed**; related public-surface suites (`test_issue_67*`, `test_issue_1558_semantic_contracts`, `test_one_session_sync`, `test_interface_semantics`, `test_sync_orchestration`, `test_agentic_sync_runner`) green.
- GitHub CI: **9/9 checks pass** (incl. `Run Unit Tests`, `Public CLI Regression`, `Package Preprocess Smoke`, CodeQL).
- Multi-round adversarial review converged with **no open findings**.

### Independent regression matrix (fresh `pip install` from `d2b3f28`)
Run as raw gate before/after pairs through `_verify_public_surface_regression`:

| Case | Expected | Result |
| --- | --- | --- |
| `2190_clean_import_only` | pass | ✅ pass |
| `2093_worker_imports_only_shape` | pass | ✅ pass |
| `mixed_import_plus_real_removal` | block | ✅ `error: real_api` |
| `dunder_all_reexport_removed` | block | ✅ `error` |
| `redundant_alias_reexport_removed` | block | ✅ `error` |
| `logger_assignment_still_blocked` | block | ✅ `error: logger` |
| `required_signature_change_still_blocked` | block | ✅ `changed: f` |

Also passed on **staging-1** with this commit: real `pdd sync` smoke (Cloud Build `c901827e-b48c-49a4-adf7-605b1548735e`) and the exact regression matrix (Cloud Build `1a05f3b6-2ca3-4722-8232-4ac466efe7d0`).

### Staging-2 deployable-artifact verification (Cloud Build `0f2e8499-9997-41c6-a2b5-644352c96d36`)
Built the **actual staging-2 worker image** from `Dockerfile.worker` with `PDD_SOURCE_SHA=d2b3f28` (the same image the executor runs `pdd` from), then ran the matrix *inside that image*:

- Installed `pdd` source commit: `d2b3f28f928a280fcc4babdb83240481167aa5cf` ✅
- Fix symbol `_reexport_binding` present in the installed package ✅
- **Public-surface matrix: 7/7** — both import-only cases **pass**; all five controls (real-function removal, `__all__` re-export, redundant-alias re-export, `logger` assignment, required-signature change) **still raise**.

### Conclusion
The import/typing-helper false-positive (`removed: Any, Literal, dataclass, field, json`, …) is fixed, and the gate still blocks real public-API regressions, declared re-export removals, and signature changes — verified deterministically on the deployable staging-2 artifact. Ready to merge.

Known limitation (pre-existing, unchanged here, out of scope): a re-export reachable only through a mutated/conditional `__all__` (e.g. `if cond: __all__.append('Foo')`) is not statically resolved by `_extract_dunder_all`; protect such a re-export with a clean `__all__` literal or a redundant alias.

gltanaka (changes requested):

The core import-only false-positive fix looks needed, but this PR currently documents and tests behavior for mutated/conditional `__all__` that the implementation does not provide.

`pdd/prompts/code_generator_main_python.prompt` now says unresolvable/mutated `__all__` patterns such as `if platform: __all__.append(...)` fall back to the heuristic, and `tests/test_code_generator_main.py` parametrizes `__all__.append('Extra')` / conditional append as "unresolvable". But `_extract_dunder_all()` only resets on assignments and `AugAssign`; it ignores `__all__.append(...)` calls entirely.

I verified this on the PR head:

```python
from pdd.code_generator_main import _extract_dunder_all, _snapshot_public_surface, _verify_public_surface_regression
import ast

src = "__all__ = ['keep']\n__all__.append('Extra')\ndef keep(): return 1\ndef Extra(): return 2\n"
print(_extract_dunder_all(ast.parse(src)))      # {'keep'}
print(_snapshot_public_surface(src, 'python')) # {'keep'}

# Removing Extra is not blocked, even though runtime __all__ exported it.
_verify_public_surface_regression(src, "__all__ = ['keep']\n__all__.append('Extra')\ndef keep(): return 1\n", 'module_Python.prompt', 'pdd/module.py', 'python', '')
```

Please either:

1. implement top-level runtime-mutation detection for `__all__` (at least `__all__.append/extend/...` and conditional containers) so those cases really return `None` and add a test that a non-import symbol added via mutation is protected by fallback, or
2. remove the mutated/conditional `__all__` claims from the prompt/test names/comments and keep the documented limitation aligned with the actual parser.

The current added test only proves that removing `Any` passes; it would pass whether append was treated as unresolvable or silently ignored, so it does not guard the behavior it describes.

Serhan-Asad:
Good catch — you're right on both counts: the prompt/tests described mutated/conditional `__all__` as falling back to the heuristic, but `_extract_dunder_all` ignored `.append(...)`, so a runtime-exported symbol could be silently dropped and the test only proved the import half.

Went with **option 1 (implement it)**. Pushed `35262b127`.

**What changed**

`_extract_dunder_all` now resolves `__all__` in **source order** (last write wins, matching runtime):
- a clean top-level `__all__ = <string-literal list/tuple>` (or bound `__all__: T = [...]`) sets the resolved set;
- a value-less annotation (`__all__: T`, `__all__[0]: T`) is a no-op (the `Store` ctx is an AST artifact, not a write);
- **any other write resets to `None`** (→ heuristic): a mutator-method call (`append`/`extend`/`insert`/`remove`/`pop`/`clear`/`sort`/`reverse`/dunder mutators — not read-only `.copy()`/`.index()`), subscript/slice store/del, augassign, `del`, a conditional/looped rebind, tuple/walrus/`for`/`with`/`except`/`case` targets, **and an import that binds `__all__`** (`from exports import __all__`);
- a **later clean literal restores** resolvability, so `__all__ = ['A']; __all__.append('B'); __all__ = ['A']` correctly resolves to `{'A'}` again.

The scan (`_subtree_mutates_dunder_all`) descends into `if`/`for`/`while`/`with`/`try`/`match` bodies but **not** into def/class/lambda scopes or comprehension loop targets, so a function-local `__all__.append(...)` or a comprehension var doesn't false-trip it.

**Your exact repro now:**
```
_extract_dunder_all(...)       -> None
_snapshot_public_surface(...)  -> {'Extra', 'keep'}
_verify_public_surface_regression(..., remove Extra)  -> PublicSurfaceRegressionError, removed=['Extra']
```

**Key constraint kept:** this does NOT capture imports conservatively — the surface/signature import branches stay re-export-only whenever `__all__` is unresolvable, so a plain `from typing import Any` in a `if platform: __all__.append(...)` module is still removable (no #2256 reintroduction). The fallback protects DEFINED symbols (the appended function) while imports stay lenient.

**Tests** now guard the behavior, not just the import half:
- `test_..._mutated_dunder_all_protects_appended_symbol` (parametrized over append/extend/`+=`/subscript/conditional) — appended function removal raises;
- `test_..._mutation_overridden_by_later_literal` — source-order restore;
- `test_..._import_rebind_falls_back` — `from x import __all__`;
- `test_..._value_less_dunder_all_annotation_is_no_op`;
- the existing lenient test now explicitly only asserts the import half.

Prompt updated to describe the source-order resolution + the import-rebind reset. Full `tests/test_code_generator_main.py` (361) green.

gltanaka (approved):
Approved. I rechecked the updated head and the prior blocker is fixed: mutated/conditional __all__ now falls back as documented, defined symbols added through that path are protected, and import-only cleanup remains lenient. Targeted local tests and the PR checks are green.

Issue #1664, opened 2026-06-19
pdd agentic sync: module-identification prompt exceeds provider input limit, then misreports 'no modules to sync' (mirrors pdd_cloud#2196)


DianaTao:
## Summary
Agentic `pdd sync` fails to start because the module-identification prompt is sent to the model **over the provider input limit** (1,096,015 > 1,048,576 chars), and the fallback path then reports **"LLM identified no modules to sync"** and exits 1 — so no sync work is attempted at all.

Mirror of cloud issue **promptdriven/pdd_cloud#2196** (checkup redesign: observability + structured failure handling). Fix belongs in the `pdd` CLI agentic-sync module-identification step; tracked in both repos so we can fix on both ends.

## Where it happened
Cloud `sync` job on promptdriven/pdd_cloud#2196, base branch `epic/issue-2189-redesign-checkup-architecture`.

## Error (verbatim)
```
## PDD Agentic Sync - Error

LLM failed to identify modules: All agent providers failed: openai: Exit code 1:
Error: turn/start: turn/start failed: Input exceeds the maximum length of
1048576 characters. (code -32602), data: {"input_error_code":"input_too_large",
"max_chars":1048576,"actual_chars":1096015}
```
then:
```
## PDD Agentic Sync - Error
LLM identified no modules to sync
```
then:
```
### ❌ PDD Execution Failed
Execution failed: PDD command failed with exit code 1.
Output: Fetching issue #2196 from promptdriven/pdd_cloud...
Successfully loaded prompt: agentic_sync_identify_modules_LLM
Identifying modules to sync via LLM...
Status: Failed
Message: LLM identified no modules to sync
Cost: $0.7204
Model: anthropic
```

## Why this is a bug
1. **Unbounded context** — the `agentic_sync_identify_modules_LLM` step packs the whole repo/diff context into a single prompt with no chunking or size guard, so on a large branch it exceeds the 1,048,576-char provider limit and every provider fails.
2. **Misleading fallback** — after the input-too-large failure, the step reports `LLM identified no modules to sync` (which sounds like "nothing to do") instead of surfacing the real cause (input too large). A real, oversized sync silently looks like a no-op and exits 1.

## Expected
- Bound / chunk / summarize the module-identification context so it stays under the provider input limit (and/or pick a provider by context window).
- On `input_too_large`, fail with the **actual** reason and a retryable/terminal classification — do **not** collapse it into "no modules to sync".

## Related
[...]

PR #1683, opened 2026-06-19
fix(sync): canonical module resolution (#1677) + carry a ResolvedSyncUnit through agentic sync (#1675)

### LINKED ISSUE #1675: Carry a resolved sync unit through agentic sync

### LINKED ISSUE #1677: pdd sync should resolve modules by canonical path, not ambiguous short basenames
## Problem

`pdd sync` can still resolve a module too loosely when multiple modules share the same short basename, especially in Next.js projects where many files are named `page.tsx`, `layout.tsx`, or `config.ts`.

When the CLI receives or derives a short module name such as `page`, it can generate artifacts for the wrong path instead of using the exact module path already available in `architecture.json`.

This is a CLI/root-cause issue, not just an individual repo cleanup issue. A downstream repo can patch the generated file, but the same mistake can happen again in a future `pdd sync` run unless module identity is path-aware inside PDD.

## Recent evidence

This happened recently in `pdd_cloud`, not only in old issue history.

In `promptdriven/pdd_cloud` PR #1828 for issue #1821:

- `pdd change` created PDD bot commit `82f2723b` on 2026-05-29.
- That commit added the intended root architecture entry:
  - prompt: `frontend/app/login/page_TypeScriptReact.prompt`
  - filepath: `frontend/src/app/login/page.tsx`
- The frontend nested architecture entry used a short filename:
  - filename: `page_TypeScriptReact.prompt`
  - filepath: `src/app/login/page.tsx`
- `pdd sync` then created PDD bot commit `7782f10c` on 2026-05-29.
- That sync generated the wrong path:
  - `frontend/src/page.tsx`
  - `.pdd/meta/page_typescriptreact.json`
  - generic `examples/page_example.tsx`
  - generic `tests/test_page.tsx`

The correct generated file should have been:

- `frontend/src/app/login/page.tsx`

The PR was later manually/auto-heal corrected before merge, so the bad file did not remain on `main`, but the CLI behavior occurred inside a real May PR.

Related older context:

- `promptdriven/pdd_cloud` issue #646 was an older architecture/sync issue.
- `gltanaka/pdd` issue #860 was closed because tracking moved, not because every path-resolution class was necessarily fixed.
- `promptdriven/pdd` issue #1201/PR #1243 likely fixed part of the inconsistent output-path handling, but the May 29 `pdd_cloud` evidence shows why we still need a clear resolver contract and regression coverage for ambiguous basenames.

## Root cause hypothesis

PDD currently allows multiple forms of module identity:

- short basename, for example `page`
- prompt-derived path, for example `app/login/page`
- architecture filename, for example `frontend/app/login/page_TypeScriptReact.prompt`
- architecture filepath, for example `frontend/src/app/login/page.tsx`
- metadata key, for example `.pdd/meta/page_typescriptreact.json`

The unsafe part is that short basenames are sometimes treated as enough identity even when they are ambiguous.

For projects like Next.js, `page` is not a stable module id. It is only the leaf filename. The stable identity is the path-qualified module, and for existing modules the strongest source of truth should be `architecture.json.filepath`.

## Expected behavior

PDD should use one central resolver for module identity.

For existing modules:

1. Prefer exact `architecture.json.filepath` when deciding where code should be written.
2. Preserve and use path-qualified prompt identity where available.
3. Treat short basenames as aliases only when they resolve to exactly one module.
4. If a short alias matches multiple modules, fail before generating files and list the valid choices.
5. In GitHub/App/agentic flows, canonicalize from branch context where possible instead of asking the user to rerun manually.

Example:

- `app/login/page` should resolve to the login page module.
- `frontend/app/login/page` should resolve to the login page module.
- `page` should fail if multiple `page` modules exist.

## GitHub App behavior

For automated cloud runs, the failure should happen early, before child sync/generation writes files.

If the LLM returns `page` but the PR changed `prompts/frontend/app/login/page_TypeScriptReact.prompt`, PDD should canonicalize that to `frontend/app/login/page` automatically.

If there is truly no disambiguating context, PDD should stop with a clear message like:

```text
Ambiguous module alias 'page'. Use one of:
- frontend/app/login/page
- frontend/app/settings/security/page
- frontend/app/settings/github/page
...
```

That is better than generating `frontend/src/page.tsx` and requiring manual repair later.

## Proposed implementation

Add or formalize a central module resolver, used by all sync-related code paths.

The resolver should build a module table from architecture data with:

- canonical id: path-aware module id, preferably derived from architecture filepath or full prompt path
- prompt path aliases
- filepath aliases
- legacy basename aliases only when unique
- metadata keys, with compatibility fallback

Then wire the resolver into:

- `pdd sync` module selection
- agentic sync module/child job selection
- `sync_determine_operation.get_pdd_file_paths`
- metadata read/write naming
- any sync/update/change path that currently strips down to a basename

Likely areas to inspect:

- `pdd/agentic_sync.py`
  - `_architecture_module_basenames`
  - `_architecture_sync_modules`
  - `_filter_invalid_basenames`
- `pdd/agentic_sync_runner.py`
  - architecture alias/dependency graph helpers
- `pdd/sync_determine_operation.py`
  - `get_pdd_file_paths`
- operation metadata handling under `.pdd/meta`

## Backward compatibility

Do not hard-break existing short names that are unambiguous.

Compatibility rules:

- Unique short aliases should continue to work.
- Ambiguous short aliases should fail early with choices.
- Existing old metadata files such as `.pdd/meta/page_typescriptreact.json` should still be readable.
- New metadata should be written under canonical path-aware keys, for example `frontend_app_login_page_typescriptreact.json`.
- Avoid mass-renaming/churning existing metadata unless needed.

This should avoid breaking existing projects while preventing wrong-path generation.

## Acceptance criteria

Add regression tests covering:

- two architecture entries with the same leaf basename, for example two `page.tsx` modules
- `pdd sync page` fails as ambiguous and lists choices
- `pdd sync app/login/page` resolves correctly
- architecture `filepath` wins over `.pddrc` pattern/default output rules for existing modules
- old metadata key is still read
- new metadata key is path-aware
- agentic sync does not pass ambiguous short basenames into child sync jobs
- a fixture similar to `pdd_cloud` PR #1828 does not generate `frontend/src/page.tsx`

## Why this matters

Downstream projects should be able to trust `pdd sync` to write to the exact module path already recorded in architecture data.

Without this resolver contract, every downstream repo has to add local guards or manually clean up bad generated files, which fixes the symptom but leaves the CLI capable of repeating the same mistake.

Serhan-Asad:
> Resolves two related agentic-sync issues in one change set. Also incorporates the `#1664` provider-input-limit fix already on `main`.

Closes #1675
Closes #1677

Flow in `run_agentic_sync`: detect modules → **normalize + bound the identify prompt under the provider input limit (#1664)** → **drop ambiguous bare names / resolve by canonical path (#1677)** → **resolve each module to a `ResolvedSyncUnit` and thread it through dry-run / dep-graph / runner / durable (#1675)**.

---

## Part 1 — Canonical module resolution + fail-fast on ambiguity (#1677)

`pdd sync` could silently resolve an ambiguous **bare** basename (e.g. the Next.js leaf `page`) to one module, or mis-place a path-qualified module. Now:
- **`AmbiguousModuleError`** — `get_pdd_file_paths` fails fast (listing the conflicting targets) when a bare basename maps to ≥2 architecture outputs, before any generation; re-raised through the sync paths.
- **Canonical path resolution** — path-qualified modules resolve under their real directory; an explicit `/`-terminated output dir is used directly (no double-pathing).
- **Agentic** `_drop_ambiguous_basenames` drops bare ambiguous names (keeping path-qualified + unique ones).

## Part 2 — Carry a ResolvedSyncUnit through agentic sync (#1675)

Agentic sync re-derived a module's identity (cwd, `.pddrc`, context, target, prompts_dir, architecture, output/meta) inconsistently across stages; the visible symptom was a nested project dying on `Unknown context` when the global `--context` was forwarded into a per-module cwd.
- **`pdd/resolved_sync_unit.py`** — a frozen `ResolvedSyncUnit` with `resolve_sync_unit(...)` and `relocate(old_root, new_root)`; one canonical identity threaded through global, targeted, durable, and runner stages.
- **Per-module context** (never fails): forward `--context` only when valid for the module's own `.pddrc`; else substitute that cwd's specific context; else omit.
- **Consistent downstream** — `_run_dry_run_validation` returns per-module cwd + target; `_build_targeted_dep_graph` groups by owning project and remaps edges to full keys; durable staging keys `.pdd/meta` by target and `relocate()`s the unit into each worktree.

Both halves are reconciled in `agentic_sync.py` (the only file the two issues both touch): #1664's normalization, #1677's ambiguity guard, and #1675's ResolvedSyncUnit coexist in one `run_agentic_sync` flow.

## Validation
- **817 tests pass** across the combined suite: `test_agentic_sync`, `test_resolved_sync_unit`, `test_agentic_sync_runner`, `test_durable_sync_runner`, `test_issue_1677_canonical_module_resolution`, `test_explicit_output_paths`, `test_sync_determine_operation`, `test_construct_paths`, `test_agentic_sync_fallback_realgit`, `test_e2e_issue_557_codex_ndjson`, `test_fix_error_loop_sync_e2e` (`-p no:randomly -m "not real"`).
- Includes #1664's prompt-size/normalization suites and #1675's ResolvedSyncUnit suites, with dry-run mocks updated to the 5-tuple + `module_targets` API.


Serhan-Asad:
## ✅ Verification & testing — both halves validated

### #1675 (ResolvedSyncUnit) — unit + CLI + live staging cloud E2E
- **593 tests pass** across the combined suite (`test_resolved_sync_unit`, `test_agentic_sync`, `test_agentic_sync_runner`, `test_durable_sync_runner`, `test_issue_1677_canonical_module_resolution`, `test_construct_paths`), `-p no:randomly -m "not real"`.
- **Deterministic CLI replay** of the original nested-context bug on the real `.pddrc` layout (root context whose `prompts_dir` points into a nested project): fails with `Unknown context` before, resolves the correct nested context after. Covers every key shape — full repo-root, nested-relative, bare, root-layout, root+nested collision, root-context-points-into-nested.
- **Staging cloud E2E (deployed worker):** built the staging `pdd-executor-job` image from the current staging worker source + this branch's pdd (only the pdd version changed), deployed it, and triggered a **real agentic sync** on a nested-path module (`commands/contracts`). The deployed build resolved **identically to the production baseline**:
  - `Modules to sync: ['commands/contracts']` (nested path preserved)
  - `Dependency graph: {'commands/contracts': []}`
  - dry-run **passed** — `✓ commands/contracts → cwd: .`
  - child sync invoked with the resolved unit; reached preprocessing + codegen
  - **no** `Unknown context`, **no** `AmbiguousModuleOwnerError`, **no** resolver traceback

  The run's only failure was a downstream public-surface regression in the regenerated code (model codegen quality) — **identical to the baseline run on the same module**, unrelated to this change. The staging executor was reverted to its baseline image afterward.

### #1677 (canonical module resolution) — validated end-to-end
- Ambiguous bare basename now **exits non-zero** with the conflicting-target list instead of silently generating to the wrong file; path-qualified names (`app/login/page`) and unique short names are unaffected. Verified end-to-end.

### Merge integrity
The two features touch only `pdd/agentic_sync.py` and there the additions are complementary (`_drop_ambiguous_basenames` runs before `_resolve_module_cwd_and_target`); everything else is disjoint. All 593 tests pass on the merged tree, confirming both features work together.

**Net: this is 100% done and ready for review.**

Serhan-Asad:
Re-integrated #1675's full ResolvedSyncUnit onto the clean #1677 line (it had been dropped during an earlier merge). The two features are reconciled in `agentic_sync.py` alongside #1664's input-limit fix, and the combined suite is green — **817 passed** (incl. #1664's prompt-size/normalize suites and #1675's ResolvedSyncUnit suites, dry-run mocks on the 5-tuple + `module_targets` API). The PR now resolves both #1675 and #1677.

gltanaka (changes requested):
I think the underlying fixes are needed, but I would not merge this PR yet.

#1675 and #1677 are both real issues, and the focused resolver/unit tests pass locally. I found two merge blockers that should be fixed first:

1. `DurableSyncRunner` rejects correctly anchored metadata when the governing `.pddrc` is an ancestor of the module cwd.

`pdd/durable_sync_runner.py:470` allowlists only root `.pdd/meta` plus `<module_cwd>/.pdd/meta`. But `operation_log._resolve_meta_dir()` anchors metadata at the nearest `.pddrc` parent. For a module running from `backend/functions` with `backend/.pddrc`, the child writes metadata under `backend/.pdd/meta`, which durable mode currently treats as unsafe.

Minimal method-level reproduction on this branch:

```text
allowed: ['.pdd/meta', 'backend/functions/.pdd/meta']
unsafe: ['backend/.pdd/meta/report_python.json']
```

That means durable sync can either fail to force-add ignored metadata or reject a correctly staged metadata file for the exact ancestor-`.pddrc` case #1675 is trying to support. Please include the resolved unit's governing project root, e.g. `unit.pddrc_path.parent`, in `_allowed_metadata_prefixes()` / `_force_add_module_metadata()` and add a regression test where `cwd=backend/functions` and `.pddrc`/metadata live at `backend/`.

2. Prompt specs are stale for changed prompt-backed modules.

This PR changes `pdd/agentic_sync.py`, `pdd/sync_determine_operation.py`, `pdd/sync_main.py`, `pdd/sync_orchestration.py`, `pdd/commands/maintenance.py`, and `pdd/core/errors.py`, but only updates the runner prompts. For example, `pdd/prompts/agentic_sync_python.prompt:91` still says `_run_dry_run_validation()` returns `(all_valid, module_to_cwd_map, errors, total_llm_cost)`, and `pdd/prompts/agentic_sync_python.prompt:95` still documents `_filter_already_synced(modules, module_cwds, quiet)`. The code now threads `module_targets`, builds `ResolvedSyncUnit`s, and changes those contracts. `pdd/prompts/sync_determine_operation_python.prompt` also does not describe `AmbiguousModuleError` or the new ambiguous-bare-basename contract.

Since prompts are the source/spec for future PDD syncs, please update the corresponding prompts/architecture metadata for the changed prompt-backed modules so a later sync does not regenerate away these fixes.

Serhan-Asad:
Thanks @gltanaka — both blockers + the CodeQL note are addressed in `5c75a61`:

**1. DurableSyncRunner ancestor-`.pddrc` metadata.** `_allowed_metadata_prefixes()` now also allows the governing project root's `.pdd/meta`, derived from the resolved unit's `unit.pddrc_path.parent` (exactly as you suggested) — in addition to root `.pdd/meta` and `<module_cwd>/.pdd/meta`. So a module run from `backend/functions` governed by `backend/.pddrc` can force-add / validate `backend/.pdd/meta` instead of having it rejected as unsafe. `_force_add_module_metadata` and `_unsafe_staged_paths` both consume that list, so both paths are fixed. Added a regression test (`test_durable_allows_ancestor_pddrc_metadata`) with exactly your repro: `cwd=backend/functions`, `.pddrc`/metadata at `backend/` — it asserts `backend/.pdd/meta` is allowed, the correctly-anchored file is **not** flagged, and a wrong-name file there still is.

**2. Stale prompt specs.** Updated the specs for the changed prompt-backed modules so a future sync reproduces these fixes:
- `agentic_sync_python.prompt`: `_run_dry_run_validation` now documented as the 5-tuple (`…, module_to_target_map, …`), `_filter_already_synced(…, module_targets=None)`, `AsyncSyncRunner` dispatch with `module_targets`/`module_units`, `_build_targeted_dep_graph`, and a `resolve_sync_unit`/`ResolvedSyncUnit` contract entry.
- `sync_determine_operation_python.prompt`: documents `AmbiguousModuleError` (the `ValueError` subclass, fail-fast before resolution) and the canonical / no-double-path reanchor contract.
- `sync_main_python.prompt` + `sync_orchestration_python.prompt`: document the `AmbiguousModuleError` fail-fast / propagation.
- `durable_sync_runner_python.prompt`: documents the ancestor-`.pddrc` metadata allowlist from (1).

**CodeQL (path expression at `sync_determine_operation.py`).** `_reanchor_under_basename_subdir` now drops `.`/`..` segments from a basename's directory, so a basename can't introduce a path-traversal component into the re-anchored output path.

Full combined suite is green — **818 passed** (incl. the new durable test). CI re-running on the new head.

gltanaka (approved):
Re-reviewed the updated head after the requested changes. The durable ancestor-.pddrc metadata allowlist is fixed, the prompt specs/architecture metadata were updated, focused local verification passed, and full CI is green. Approved for merge.

PR #1705, opened 2026-06-22
docs(onboarding): document ADC for Vertex AI, warn against placeholder creds

niti-go:
## Summary

The Vertex AI section of the onboarding guide only documented the service-account **key-file** path and showed a placeholder `VERTEX_CREDENTIALS=/path/to/service-account.json`. Two problems:

1. **No mention of ADC.** PDD authenticates to Vertex via Application Default Credentials when `VERTEX_CREDENTIALS` / `GOOGLE_APPLICATION_CREDENTIALS` is unset and a project is configured (`pdd/llm_invoke.py`). Anyone with `gcloud` can use Vertex without creating a key file — the guide never said so.
2. **The placeholder can break setup.** A non-empty `VERTEX_CREDENTIALS` (even the example placeholder) is treated as a real credential by `resolve_api_key_from_env` and disables the ADC fallback, so Vertex auth fails on a missing file. Following the doc literally could break Vertex.

## Changes (docs-only)

- Lead with **ADC** as the simplest path: `gcloud auth application-default login`, give the account the **Vertex AI User** role, set project/location, and leave `VERTEX_CREDENTIALS` unset.
- Keep the service-account key-file method as a clearly-labeled fallback.
- Add a warning not to set `VERTEX_CREDENTIALS` to a placeholder path.
- Document the full `gcloud` auth sequence for ADC (not just `application-default login`): `gcloud auth login`, `gcloud auth application-default login`, and `gcloud config set project YOUR_GCP_PROJECT_ID`.

## `PDD_PATH` clarification (`Final Project Configuration`)

A follow-up commit fixes a separate onboarding trap I hit on my own machine: `PDD_PATH` was easy to set to the repo root, which makes `pdd setup` abort with `Completion script not found` and breaks resource lookups.

- Clarified that `PDD_PATH` must point at the **inner `pdd/` package directory** — the one containing `cli.py` and the `pdd_completion.*` scripts. Because the repository and the package share the name `pdd`, the correct path ends in `…/pdd/pdd`, **not** the repo root.
- Updated the `pwd` hint ("should end in `/pdd/pdd`") and added a `ls "$PDD_PATH/cli.py"` existence check to the verify step so a wrong path is caught immediately.

## Why it matters

New users can set up Vertex with a login they likely already have, instead of being told to create and download a service-account key — and they avoid both the placeholder trap that silently breaks Vertex auth and the `PDD_PATH` trap that breaks `pdd setup`.

## Verification

- Confirmed PDD reaches Vertex AI via ADC alone (no key file) at both `global` and `us-central1` using `vertex_ai/gemini-2.5-flash`.
- Confirmed `pdd setup` runs past the completion step into the wizard when `PDD_PATH` points at the package directory (and aborts with `Completion script not found` when it points at the repo root).

---
*Opened as a draft for review.*


gltanaka (changes requested):
This update is needed, but I do not think it should merge as-is.

Required changes:

1. Align the service-account key-file path with the current runtime variable. The current Vertex catalog rows use `GOOGLE_APPLICATION_CREDENTIALS|VERTEXAI_PROJECT|VERTEXAI_LOCATION`, and `pdd/provider_manager.py`'s Vertex setup saves `GOOGLE_APPLICATION_CREDENTIALS`. This PR keeps documenting `VERTEX_CREDENTIALS` for the key-file path, which means a user following the service-account fallback can still end up with a value current Vertex invocation does not consume. Please either document `GOOGLE_APPLICATION_CREDENTIALS` as the service-account file variable, or add/verify an explicit runtime alias from `VERTEX_CREDENTIALS` to `GOOGLE_APPLICATION_CREDENTIALS` before documenting it as supported.

2. Update `.env.example` in the same PR. The onboarding guide points readers there, but `.env.example` still has `VERTEX_CREDENTIALS=/path/to/service-account.json`, so the placeholder trap this PR is trying to remove remains in the canonical example file. Prefer commented ADC-first examples, with a commented `GOOGLE_APPLICATION_CREDENTIALS=/absolute/path/to/service-account.json` only for the explicit key-file fallback.

3. Tighten the PDD_PATH wording to the concrete failure mode. The inner package dir clarification is useful because `pdd setup`/completion installation looks for `pdd_completion.*` and `cli.py` under `PDD_PATH`. The broader statement that the repo-root setting "breaks resource lookups" is too broad for a repo checkout where prompts/data are reachable through root paths/symlinks; if a specific resource lookup fails, please name that exact lookup.

Once those are fixed, the ADC-first direction looks worthwhile to merge.

niti-go:
Thanks @gltanaka — all three points are addressed. Latest commits: `ff76c3ab7` (initial response) and `8e9bcf149` (completes point 2). Summary with the runtime references I verified against the code:

**1. Service-account variable aligned with the runtime.** The key-file path now documents `GOOGLE_APPLICATION_CREDENTIALS` (not `VERTEX_CREDENTIALS`) in both `docs/ONBOARDING.md` and `.env.example`. That matches what current Vertex invocation actually consumes:
- catalog rows — `pdd/data/llm_model.csv` Vertex rows use `GOOGLE_APPLICATION_CREDENTIALS|VERTEXAI_PROJECT|VERTEXAI_LOCATION` (and `generate_model_catalog.py` emits exactly that);
- `provider_manager.py` — `COMPLEX_AUTH_PROVIDERS["Google Vertex AI"]` saves `GOOGLE_APPLICATION_CREDENTIALS` / `VERTEXAI_PROJECT` / `VERTEXAI_LOCATION`;
- `model_tester.py` — `pdd setup`'s connection test flags a `GOOGLE_APPLICATION_CREDENTIALS` path that isn't a real file ("Path set but file not found").

  ADC-first is correct for this catalog: the multi-credential resolver in `llm_invoke.py` succeeds on project+location alone and lets ADC authenticate when `GOOGLE_APPLICATION_CREDENTIALS` is unset. `VERTEX_PROJECT` / `VERTEX_LOCATION` are labeled "legacy aliases" because `llm_invoke.py` still reads them as fallbacks for project/location.

**2. `.env.example` fixed — no active placeholders.** Dropped the `VERTEX_CREDENTIALS=/path/to/service-account.json` placeholder; the Vertex block is now ADC-first and fully commented — `GOOGLE_APPLICATION_CREDENTIALS` is a commented key-file fallback, and (in `8e9bcf149`) `VERTEXAI_PROJECT` / `VERTEXAI_LOCATION` are commented too. They had been left uncommented, so a verbatim `cp .env.example .env` set `VERTEXAI_PROJECT=your-gcp-project-id`; the multi-cred path then treats Vertex as "configured" and fails the real call — the same placeholder-trap class this PR removes. The ONBOARDING `.env` snippet was made consistent.

**3. PDD_PATH wording tightened to the concrete failure.** Replaced the broad "breaks resource lookups" with: pointing `PDD_PATH` at the repo root makes `pdd setup` fail with "Completion script not found" because the completion installer looks for `pdd_completion.*` under `PDD_PATH`. Also added `ls "$PDD_PATH/cli.py"` to the verify step so a wrong path is caught immediately.

No merge conflicts — `mergeable=MERGEABLE`, and the branch still merges cleanly into `upstream/main` (which hasn't touched these two files). Ready for re-review.

niti-go:
Addressed the requested review changes in the latest head `8e9bcf1`:

- Updated the Vertex key-file path docs to use `GOOGLE_APPLICATION_CREDENTIALS`, matching the current catalog rows and `pdd/provider_manager.py` setup behavior.
- Updated `.env.example` to remove the `VERTEX_CREDENTIALS=/path/to/service-account.json` placeholder trap and make the Vertex examples ADC-first/commented. The service-account fallback is now a commented `GOOGLE_APPLICATION_CREDENTIALS=/absolute/path/to/service-account.json` example.
- Tightened the `PDD_PATH` wording to the concrete `pdd setup` completion-script failure: the completion installer looks for `pdd_completion.*` under `PDD_PATH`. I also kept the `cli.py` verification check and fixed later `PDD_PATH` examples to end in `/pdd/pdd`.
- Commented optional `VERTEXAI_PROJECT` / `VERTEXAI_LOCATION` placeholders in the examples so a verbatim copy does not activate fake Vertex project values.

Verification:

- Local: `git diff --check`, `python -c "import pdd; print('pdd import ok')"`, and `python -m py_compile pdd/*.py` all pass. The import emitted LiteLLM's remote cost-map warning and fell back to the local backup.
- GitHub Actions on `8e9bcf1`: `Run Unit Tests`, `Public CLI Regression`, `Package Preprocess Smoke`, and `heal` all passed; `auto-heal` is neutral/skipped.
- Merge/conflict check: branch is up to date with `upstream/main` at `22e4010bf`; no local merge conflicts found.

Ready for re-review. GitHub still shows `CHANGES_REQUESTED` / `BLOCKED` only because the prior review state has not been cleared yet.

niti-go:
Other claims I independently confirmed correct:
- ADC path works: for the multi-credential Vertex rows, _ensure_api_key (llm_invoke.py:3547-3560) returns success when GOOGLE_APPLICATION_CREDENTIALS is unset but project + location resolve — and litellm then uses google.auth ADC. Location is satisfied by the CSV's pinned global, so only VERTEXAI_PROJECT is strictly required, which the doc tells users to set.
- Placeholder warning is accurate: a set GOOGLE_APPLICATION_CREDENTIALS takes precedence over ADC; model_tester.py:130 even rejects it when it's not a real file.
- "Legacy aliases also work": _vertex_project_value() reads VERTEX_PROJECT, and _vertex_location_value() reads VERTEX_LOCATION (llm_invoke.py:3079/3091).
- …/pdd/pdd path examples and the ls "$PDD_PATH/cli.py" verify step are internally consistent throughout.

gltanaka (approved):
Approved. The updated PR addresses the requested changes:

- Vertex service-account docs now use `GOOGLE_APPLICATION_CREDENTIALS`, matching the current Vertex catalog/setup path.
- `.env.example` is updated and the optional Vertex values are commented, so copying the example no longer activates placeholder Vertex credentials/project values.
- The `PDD_PATH` warning is now scoped to the concrete `pdd setup` completion-script failure instead of the broader resource-lookup claim.

This is still docs/example-only, CI is passing, and the branch is mergeable. I think this is ready to merge.

PR #1726, opened 2026-06-23
Fix CI validation stale head diagnostics

RyanTanuki:
Fixes the CI validation stale-head failure mode observed from promptdriven/pdd_cloud#2076.\n\nChanges:\n- Poll the live PR head when no explicit expected-head override is provided.\n- Continue reading live-head checks while an explicit expected head is stale, so terminal states like ACTION_REQUIRED fail immediately.\n- Return a stale-head diagnostic with expected SHA, live PR SHA, PR number, and last known checks instead of a generic timeout.\n\nDevelopment evidence:\n- python -m pytest tests/test_ci_validation.py


RyanTanuki:
https://github.com/promptdriven/pdd_cloud/issues/2076

Issue #1857, opened 2026-07-06
Goal: E2E verification that user-story regression oracles (EPIC #1816) are complete, useful, and merge-ready


sohni-tagirisa:
## Goal

Independently verify, end-to-end, that the user-story regression feature built under EPIC #1816 **works as intended, is genuinely useful, and is fully functional** — producing a clear GO / NO-GO merge-readiness verdict so the epic branch can be confidently opened as an integration PR to `main`.

This goal is written for an autonomous Claude Fable agent. The deliverable is *confidence*, backed by evidence: every claim in the verdict must be demonstrated by an actual command run, test executed, or artifact produced — not inferred from reading code.

## Context

- Upstream epic: #1816 (user stories as executable regression oracles).
- All implementation sub-PRs are merged into the epic branch `codex/issue-1816-story-regression-epic`: #1713 (`@pytest.mark.story` marker + traceability API), #1774 (`pdd story add/list/link` CLI), #1775 (cross-dev-unit story support), #1800 (backfill EPIC, absorbing #1801–#1806), #1818 (docs), #1822 (`pdd test --from-story` deterministic regression generation).
- **No integration PR to `main` exists yet.** This verification is the gate before opening one.

## Verification plan (do all of it)

Work in an isolated `git worktree` checked out at the tip of `codex/issue-1816-story-regression-epic`. Never touch `main`.

### 1. Full user journey (the happy path, as a real user)
Walk the exact pipeline promised in #1816, using real commands against a real (sample or fixture) prompt/dev-unit setup:
1. Author a story from an issue: `pdd story add` → confirm a `user_stories/story__*.md` is created with `## Oracle`, `## Acceptance Criteria`, `## Negative Cases` sections.
2. Link it: `pdd story link` to one prompt/dev unit; confirm `<!-- pdd-story-prompts: ... -->` linkage and `pdd story list` shows regression status.
3. Repeat with a **cross-dev-unit** story (≥2 dev units) — confirm traceability attributes all linked units without double-counting.
4. Generate tests: `pdd test --from-story` → confirm the generated pytest test is deterministic (no LLM/cloud calls), asserts the `## Oracle` outcomes, rejects the `## Negative Cases`, and carries `@pytest.mark.story` + `story_id` tags.
5. Run the lane: `pytest -m story` (and `make regression-stories` / `regression_public.sh` wiring if present) — confirm it passes with **no LLM or cloud credentials available** (unset them to prove it).
[...]

PR #1866, opened 2026-07-06
fix(gate): reconcile duplicate story claims instead of first-wins collapse (#1857 G6)

sohni-tagirisa:
Closes gap **G6** from the #1857 verification report — the root cause behind **G1** (fixed symptomatically in #1860).

## Problem
`discover_story_markers` kept one marker per `story_id` (first by file path, line), **silently dropping sibling claims**. A story legitimately claimed by several tests (documented behavior) could be reported **stale** because a stale duplicate sorted before a fresh one — exactly how 4 top-flow seeds shipped stale.

## Fix
- New `discover_story_markers_all()` keeps **every** claim in deterministic (file, line) order.
- `classify_story()` reconciles: **OK if any claim records an acceptable hash**; stale only when none do (detail notes the claim count); hashless-stale only when no claim has a hash. Accepts both the legacy single-marker mapping and the new list mapping.
- `evaluate_stories()` uses the all-claims discovery.
- `discover_story_markers()` remains as a documented first-wins compat view.
- The gate now agrees with the lightweight coverage evaluator (`evaluate_story_regression`), which already had any-match semantics.

## Evidence
- New tests: all-claims discovery order; **stale-first + fresh-second duplicate → OK** (the G1 scenario, red before this fix); two-stale duplicate → stale with "2 claiming tests" detail.
- `tests/test_story_regression_gate.py`: **34 passed** (31 existing + 3 new).
- Repo-level gate at the epic tip unchanged: 9 ok / 0 stale / 23 missing.

Part of #1857 / EPIC #1816. Targets the epic branch, never `main`.

### DIFF
--- a/pdd/story_regression_gate.py
+++ b/pdd/story_regression_gate.py
@@ -18,7 +18,8 @@ with and without secrets.
 Public API
 ----------
 story_id_from_path(path) -> str
-discover_story_markers(tests_dir) -> dict[str, StoryMarker]
+discover_story_markers_all(tests_dir) -> dict[str, list[StoryMarker]]
+discover_story_markers(tests_dir) -> dict[str, StoryMarker]  (first-wins compat view)
 classify_story(story_path, markers, *, story_text=None) -> StoryRegressionResult
 evaluate_stories(*, stories_dir, tests_dir, only_story_ids) -> list[StoryRegressionResult]
 changed_story_ids(project_root, *, base_ref) -> Optional[set[str]]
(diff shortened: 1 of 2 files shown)


sohni-tagirisa:
Closing as a duplicate of #1864 (earlier, same root-cause fix: keep every duplicate story claim and classify OK when any claim is fresh). Design nuance from this PR that #1864 may want: when a story is stale across multiple claims, reporting the claim count in `detail` (e.g. "…across 2 claiming tests") helps diagnose which sibling drifted.

PR #1867, opened 2026-07-06
feat(story): wire the behavioral (## Entry Point) generator into `pdd test --from-story`

sohni-tagirisa:
## What
`pdd test --from-story` (via `generate_story_regression_test`) only ever ran the **text-pinning** generator. The **behavioral** generator in `pdd/story_test_generator.py` — which the docs describe (`## Entry Point`, `## Seams`, assertions over `result`) — was **wired to nothing**, so:
- **0 of 32** shipped contracts declared an `## Entry Point`, and
- generated tests could catch story/contract *text* drift but not runtime *behavior* regressions — short of EPIC #1816's "executable oracle" headline.

(Root of pdd#1857 gap **G2**.)

## Change
Route by contract inside `generate_story_regression_test` (CLI untouched):
- **Behavioral** — when the contract declares a machine-readable `## Entry Point`, delegate to `story_test_generator.generate_story_test`: it imports the callable, applies `## Seams`, and asserts `## Oracle` / `## Negative Cases` as expressions over `result`.
- **Text-pin (fallback)** — contracts without an `## Entry Point` keep the exact current behavior. No shipped story changes output (0/32 have an Entry Point today).

Docs (`generating_user_stories.md` Step 3) rewritten to describe both modes and the fallback accurately (it previously claimed the `## Entry Point` was required).

## Verified end-to-end (no credentials)
With an Entry-Point contract + a real callable, `generate_story_regression_test` now emits a behavioral test that:
- calls the entry point (`result = story_result`), tagged `@pytest.mark.story(story_id=STORY_ID, story_hash=STORY_HASH)`;
- **passes** on correct code, and **goes red when the callable's behavior is mutated** (`subtotal + TAX_RATE` → `subtotal + 1`) — a real behavioral oracle.
- Entry-point-less stories (e.g. `pdd_change`) still text-pin unchanged.

## Tests
- `test_generate_routes_to_behavioral_generator_when_entry_point_present`
- `test_generate_text_pins_when_no_entry_point`
- generator + coverage + gate suites: 45 passed; `pytest -m story` green.

Found during pdd#1857 verification (gap G2). Targets the epic branch.

### DIFF
--- a/docs/generating_user_stories.md
+++ b/docs/generating_user_stories.md
@@ -134,7 +134,9 @@ pdd test --from-story user_stories/story__checkout_total.md \
   --output tests/story_regression/test_story_checkout_total.py
 ```
 
-The contract must include a machine-readable `## Entry Point` section:
+`pdd test --from-story` generates in one of two modes, chosen from the contract:
+
+**Behavioral (preferred).** If the contract declares a machine-readable `## Entry Point`, the generated test imports the callable, invokes it, and asserts the `## Oracle` / `## Negative Cases` bullets as Python expressions over `result` — a true executable oracle that goes red when the *behavior* regresses.
 
 ```markdown
 ## Entry Point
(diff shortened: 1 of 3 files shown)


sohni-tagirisa:
Cross-check from a parallel verification pass that landed on the same design (wire `story_test_generator` into `--from-story` when the contract declares `## Entry Point`, text-pin fallback otherwise) — two residual spots this PR doesn't touch, worth absorbing here or as a tiny follow-up:

1. **README.md line ~719** still overclaims: *"The contract must declare `## Entry Point` and assertion bullets…"* — after this PR it should describe both modes (behavioral when Entry Point is declared, text-pin fallback otherwise).
2. **`tests/test_story_test_generator.py::test_pdd_test_from_story_cli_writes_generated_test`** only asserts `"STORY_HASH" in output`, which both generator shapes satisfy — that's exactly why the missing wiring went unnoticed. Strengthening it to assert the behavioral shape (`ENTRY_MODULE`, `def story_result(monkeypatch):`, a `test_story_checkout_total_R1_oracle_1` function) pins this PR's wiring against regression.

PR #1871, opened 2026-07-06
fix(change): use JSON step artifacts as source of truth

### LINKED ISSUE #1850: pdd change: JSON-as-source-of-truth to kill the marker-parsing bug class (single in-place PR)
_Follow-up to #1841. Scopes the "stop parsing LLM prose for markers" idea into a single, minimal, in-place change to `pdd change` (not a `change-v2` rewrite), per the conclusion in #1841's later comments: keep the orchestrator and all 13 steps, only swap the reasoning-layer I/O contract._

## Goal
Eliminate the class of bugs caused by parsing the LLM's prose for markers (#865, #868, #1080, #964, #880, #882, #571, #467) by having each step write a `NN_name.json` (the orchestrator's decision input) and `NN_name.md` (the comment body), and having the orchestrator read the JSON field instead of regexing prose.

## Non-goals (explicitly unchanged)
The 13-step structure, orchestrator control flow, worktree/branch-holder concurrency, workflow-state persistence/resume, clean-restart, architecture.json sanitization, the pre-checkup gate (12.5), the user-story step (12.6), sync-order computation, comment sanitization/truncation, step timeouts, provider-failure abort, and **issue-steering**. Only the **input format** for the orchestrator's existing decisions flips from prose → JSON.

## Decisions locked
- In-place on `pdd change` (no `change-v2`, no CLI-harness gating).
- **Single PR.** All surfaces migrate together. JSON-first, with the existing prose parsers kept as a **permanent in-code fallback** (not a temporary crutch) for when a provider emits malformed JSON. No dead code deleted here; optional parser removal is a future cleanup once the fallback rate is measured ~0.
- `/goal` is an optional prompt enhancement where the harness supports it, never a correctness dependency — the orchestrator's parse-and-re-run is the real guarantee, because `pdd change` also runs raw-API providers.

## Mid-flight steering is unchanged and still works
Steering is *input-side*: every step runs `run_agentic_task(..., steers=_issue_step_steers() or None)`, and `_issue_step_steers` → `drain_step_steers` pulls comments a user adds *while `pdd change` is running* and injects them into the next step's instruction; `apply_clarification_steers_on_resume` covers the resume case. This migration only changes the *output-side* (how the orchestrator reads a step's decision after it runs). None of the `steers=` call sites, `drain_step_steers`, the steer cursor, or the `issue_updated_at` refresh are touched — a comment added mid-run is still folded into the very next step exactly as today.

---

## Anchor map (current code, `pdd/agentic_change_orchestrator.py` unless noted)

| Marker surface | Function (line) | Parses today |
|---|---|---|
| Hard-stop decisions | `_check_hard_stop` (1206) | `STOP_CONDITION:`, `"duplicate of #"`, `"already implemented"`, `ROUTE_REDIRECT:`/`ROUTE_RATIONALE:`, `"no dev units found"`, `"architectural"`, `"fail:"` |
| Review-loop exit | `_review_loop_no_issues` (90) | 11 fuzzy "no issues" variants |
| Changed files | `_parse_changed_files` (824) / `_extract_marker_paths` (753) | `FILES_MODIFIED:`/`FILES_CREATED:`/`DIRECT_EDITS:`/`ARCHITECTURE_FILES_MODIFIED:`/`ASSOCIATED_DOCS_MODIFIED:` |
| Manual review | `_collect_manual_review_lines` (696) | `MANUAL_REVIEW:` lines |
| Doc-sync contract | `_verify_doc_sync_contract` (1500) | `ASSOCIATED_DOCS_{MODIFIED,CONFLICTS,UNCHANGED}:` |
| Dev-unit direct edits | `_parse_direct_edit_candidates` (1146) | a markdown table |
| Comment body | `extract_step_report` (agentic_common.py:8691) | `<step_report>…</step_report>` |
| PR URL | inline `re.search` (2905) | `https://…/pull/\d+` |
| Git-status fallback (KEEP) | `_detect_worktree_changes` (1169) | actual worktree `git status` |

Step→name map (`steps_config`, 2110): 1 duplicate · 2 docs · 3 research · 4 clarify · 5 docs_change · 6 devunits · 7 architecture · 8 analyze · 9 implement · 10 architecture_update. Then review loop 11 identify_issues / 12 fix_issues (≤ `MAX_REVIEW_ITERATIONS`), 13 create_pr. Clarification steps = `{4, 7}`.

---

## Single PR — ordered work items
Implementation *order*, not separate PRs. Build plumbing first, then migrate each surface. Every surface keeps its prose parser as fallback, so the PR is internally safe to build/test surface-by-surface.

### Item A — Shared artifact plumbing

**A.1 `artifacts_dir` in `context`** — inject a per-issue, gitignored dir the same way `issue_content` is injected.
- Steps 1–8 run in `cwd` (no worktree yet) → `cwd/.pdd/change/issue-<N>/`.
- Steps 9–13 run in the worktree (`current_work_dir = worktree_path`) → `worktree_path/.pdd/change/issue-<N>/`.

Recompute per phase so it tracks cwd. `.pdd/**` is already gitignored so step 13's `git add -A` never stages it. Keyed by issue number so concurrent runs on different issues never collide.

```python
def _artifacts_dir_for(work_dir: Path, issue_number: int) -> Path:
    d = work_dir / ".pdd" / "change" / f"issue-{issue_number}"
    d.mkdir(parents=True, exist_ok=True)
    return d
# in the loop, before building each step prompt:
artifacts_dir = _artifacts_dir_for(current_work_dir, issue_number)
context["artifacts_dir"] = str(artifacts_dir)
```

- **Write-root guard:** the file-tree snapshot guard refuses agent writes outside allowed roots. `artifacts_dir` must be in the allowed set (trivial for 1–8 under `cwd`; for 9–13 verify the guard treats `.pdd/` inside the worktree as writable).
- **Fresh-run reset:** on a non-resume run (and on `clean_restart`), delete the issue's artifacts dir up front so stale files aren't mistaken for this run. Never delete on resume.

**A.2 Read helpers** (next to the existing parsers)
```python
def _read_step_json(artifacts_dir: Path, filename: str) -> Optional[dict]:
    """Return parsed NN_name.json, or None on missing/unreadable/unparseable. Never raises."""
    p = Path(artifacts_dir) / filename
    try:
        return json.loads(p.read_text(encoding="utf-8"))
    except Exception:
        return None

def _read_step_md(artifacts_dir: Path, filename: str) -> Optional[str]:
    p = Path(artifacts_dir) / filename
    try:
        text = p.read_text(encoding="utf-8").strip()
        return text or None
    except Exception:
        return None
```

**A.3 Bounded re-run on invalid JSON** — when a migrated step finishes and its JSON is `None` (missing/invalid), re-run the step once via the existing retry mechanism before falling back to the prose parser. A per-step `json_retry_done` guard prevents looping.

Item A is pure plumbing — the reads exist but drive no decision yet.

### Item B — Migrate every marker surface (all in this PR, JSON-first, parser fallback)
Each surface is two edits: (a) add `Write {artifacts_dir}/NN_name.json` + `.md` instructions to that step's `.prompt` (leave existing markers so the fallback still has something to parse); (b) orchestrator reads JSON first, existing parser is the `if None:` fallback.

**B.1 — Step 9 implement (biggest bug surface).** `09_implement.json`:
```json
{
  "status": "changed | no_change_needed",
  "files_modified": ["prompts/foo_python.prompt", "README.md"],
  "files_created": ["prompts/bar_python.prompt"],
  "direct_edits": ["path/to/file"],
  "manual_review": [{"path": "docs/x.md", "reason": "originating prompt not confirmed"}]
}
```
Orchestrator (step 9 block ~2426): if JSON present → `changed_files = files_modified + files_created + direct_edits` (dedup, order-preserving); populate `context["files_created"|"files_modified"|"direct_edits"]`; feed `manual_review` into PR-body aggregation. **Keep `_detect_worktree_changes` git fallback** for the empty case. If JSON absent → existing `_parse_changed_files` path. Also update the resume block (~2018) to prefer JSON.

**B.2 — Steps 11/12 review loop.** `11_review.json` / `12_fix.json`:
```json
{ "status": "clean | fixed_issues", "manual_review": [{"path": "...", "reason": "..."}] }
```
Review loop (~2632): replace `if _review_loop_no_issues(s11_output):` with `status == "clean"`; fall back to `_review_loop_no_issues` when JSON missing. Iteration count, `previous_fixes`, step-comment posting unchanged.

**B.3 — Hard-stop steps (1, 2, 4, 6, 7).** Each writes JSON with a single decision field:
```json
{ "status": "proceed | duplicate | already_done | needs_clarification | route_bug | no_dev_units | needs_decision",
  "duplicates": ["#123"], "questions": ["..."], "rationale": "...",
  "direct_edit_candidates": ["path"] }
```
Add `_hard_stop_from_json(step_num, artifacts_dir, name)` mapping `status` → the exact stop-reason strings `_check_hard_stop` returns today, so all downstream stop handling (clarification re-run 4/7, direct-edit-continue for step 6, state saves, comment posts) is untouched. JSON first, fall back to `_check_hard_stop`. Step 6 `direct_edit_candidates` from JSON (fallback `_parse_direct_edit_candidates`).

**B.4 — Step 10 architecture / doc-sync contract.** `10_architecture_update.json`:
```json
{
  "architecture_files_modified": ["architecture.json"],
  "associated_docs_modified": ["docs/a.md"],
  "associated_docs_conflicts": ["docs/b.md"],
  "associated_docs_unchanged": ["docs/c.md"]
}
```
Feed `_verify_doc_sync_contract`'s three buckets from JSON arrays instead of `_extract_marker_paths`. Silent-drop/overlap detection, `ORCHESTRATOR_POSTCHECK_WARNINGS`, `PDD_STRICT_DOC_SYNC` abort, and `_scope_architecture_to_changed_files`/`_sanitize_architecture_*` all stay. Conflicts → `manual_review` synthesis (~2890) reads the JSON array.

**B.5 — Step 13 create PR.** `13_create_pr.json`:
```json
{ "status": "pr_created | pr_updated | blocked", "pr_url": "...", "pr_number": 42, "summary": "..." }
```
Replace inline `re.search(r"https://github.com/\S+/pull/\d+", s13_output)` (2905) with `pr_url` from JSON; fall back to regex. `blocked` maps to existing failure return. The PR *body* is still posted directly via `gh` by the step-13 agent (as today) — only the orchestrator's read-back becomes JSON.

**B.6 — Comment bodies from `.md`.** Everywhere the orchestrator does `extract_step_report(step_output)` → `post_step_comment_once(body=...)`, prefer `_read_step_md(artifacts_dir, "NN_name.md")`; fall back to `extract_step_report`, then the existing placeholder. `_sanitize_comment_body` (redaction + 25k truncation) is unchanged — the `.md` gets the same sanitization. Applies to the step loop (~2585), the 11/12 iteration posts, and step 13.

---

## Fallback deletion is deliberately NOT in this PR
The prose parsers stay as the `if None:` fallback. If any provider emits malformed JSON, the step degrades to today's behavior instead of failing. A future optional cleanup PR can delete them once the instrumented fallback rate is ~0.

## Testing (all in this PR)
- **Unit:** `_read_step_json` (missing/malformed/valid), `_hard_stop_from_json` parity with `_check_hard_stop`, changed-files-from-JSON parity with `_parse_changed_files`.
- **Fallback:** per surface, omit the JSON and assert the old parser still drives the decision (zero-regression).
- **Golden e2e:** reuse existing `tests/test_e2e_*` change fixtures; assert identical decisions with JSON present vs. absent.
- **Re-run-on-invalid:** malformed JSON triggers exactly one re-run then falls back.
- **Steering regression:** a comment added mid-run still reaches the next step's instruction (guards `drain_step_steers` is undisturbed).

## Build order within the PR
Item A (plumbing) → Step 9 → review loop → hard-stops → doc-sync → PR URL → comment bodies. All in the same PR.

## Open items to verify while building Item A
1. Write-root guard admits `worktree/.pdd/change/issue-<N>/` for steps 9–13.
2. `/goal` prefix helps but isn't required — instrument the re-run+fallback rate on raw-API providers.
3. `pdd_agent_context.md` injection (already exists) can ride along per prompt but is orthogonal — sequence separately if desired.

niti-go:
Closes #1850
Closes promptdriven/pdd_cloud#2644

## Summary

`pdd change` runs a 13-step agentic workflow. Before this PR, the orchestrator decided what happened next by reading the agent's normal prose and looking for marker text like `FILES_MODIFIED:`, `STOP_CONDITION:`, `ASSOCIATED_DOCS_MODIFIED:`, or `No Issues Found`.

That was fragile. If the model did the right work but phrased its summary differently, omitted one marker, or put the marker in an unexpected place, the user could see the workflow stop, loop, skip changed files, miss manual-review notes, or fail to find the PR URL even though the underlying step had succeeded.

This PR makes the step-to-step contract structured:

- Each decision-bearing step now writes a small JSON file into a gitignored per-issue artifact directory, for example `.pdd/change/issue-1850/09_implement.json`.
- Each step can also write a Markdown file, for example `09_implement.md`, used as the human-facing issue comment body.
- The orchestrator reads the JSON first for machine decisions.
- The old prose parsers remain as permanent fallbacks, so malformed or missing JSON degrades to the previous behavior instead of hard failing.
- If a step was expected to write JSON but did not, the orchestrator reruns that step once to give the agent a chance to repair the missing artifact.

## User Experience Before And After

Before this fix, a user could run `pdd change` on an issue and watch an agent successfully edit files, but then the workflow could still stop at Step 9 because the orchestrator did not find the exact expected prose marker.

Example before:

- Step 9 edits `prompts/foo_python.prompt`.
- The agent says something like "I updated the prompt and docs" but does not emit a parseable `FILES_MODIFIED:` line.
- The orchestrator cannot reliably tell what changed.
- The user sees a failure such as `Stopped at step 9: Implementation produced no file changes`, even though files may have changed or the agent's prose contained enough human-readable information.

After this fix:

- Step 9 writes `09_implement.json` with `status`, `files_modified`, `files_created`, `direct_edits`, and `manual_review`.
- The orchestrator reads those fields directly instead of guessing from prose.
- Step 10 receives the correct changed-file list, doc-sync discovery runs on the right files, and the workflow continues toward review and PR creation.
- If JSON is missing or broken, the step gets one repair rerun. If it still fails to write JSON, the orchestrator falls back to the original prose output so legacy marker behavior still works.

The same pattern now covers the other decision points:

- Steps 1, 2, 4, 6, and 7: hard-stop decisions such as duplicate, already implemented, needs clarification, no dev units, or architecture decision.
- Step 9: implementation status, changed files, direct edits, and manual-review items.
- Step 10: architecture files and associated-doc sync buckets.
- Steps 11 and 12: review-loop clean/fix status and manual-review items.
- Step 13: PR creation/update/blocked status and PR URL.
- Step comments: prefer the step `.md` artifact, with `<step_report>` as fallback.

## Important Compatibility Fix In This Revision

The first version of the PR introduced a regression in the JSON repair path: if a step's first response had valid legacy prose markers but forgot the JSON artifact, and the one-time JSON repair rerun still forgot JSON, the retry output replaced the original output. That meant the fallback parser saw only the generic retry text and lost the useful first response.

This revision fixes that:

- If the JSON repair rerun writes valid JSON, the rerun output is used.
- If JSON is still invalid after the repair rerun, the orchestrator falls back using the first successful output.
- If the original attempt failed, the retry output remains the fallback source.

That preserves the intended guarantee: missing JSON should degrade to the old prose-marker behavior, not create a new failure mode.

## About Hidden Comments / Invisible Step State

There is already a related mechanism in PDD: agentic workflows persist resumable state in hidden GitHub comments using a JSON block such as `PDD_WORKFLOW_STATE`. Other workflows also pass prior `step_outputs` into later prompt context, and visible step comments are built from `<step_report>` blocks.

That mechanism is different from this PR:

- Hidden workflow-state comments are cross-machine resume storage. They remember progress after interruption.
- `step_outputs` are mostly raw text from prior steps that later prompts can read.
- `<step_report>` is for human-facing comments.
- This PR adds a local, gitignored, per-step machine contract that the orchestrator reads immediately after each step to decide control flow.

The hidden issue state cannot cleanly solve this specific bug by feeding the orchestrator's next decision:

- The failing boundary is immediate: Step 9 finishes, and the orchestrator must decide right then whether implementation changed files, needs review, should continue to doc-sync, or should stop. Hidden comments are better for persisted resume checkpoints after progress has been recorded, not as the primary handoff inside the currently running process.
- To use hidden comments as the decision input, the step would still need to produce the same structured fields somewhere first. At that point the local JSON artifact is the simpler source of truth, and a hidden comment would just be a slower remote copy of it.
- Hidden comments require GitHub issue I/O. `pdd change` also needs to behave predictably in local worktrees, retries, tests, and provider/CLI paths where relying on a freshly written issue comment would add latency, permissions failure modes, race conditions, and fork behavior differences.
- Hidden comments are shared issue-thread state. Immediate step artifacts are scoped to one run's `.pdd/change/issue-<N>/` directory, so clean restarts, retries, and concurrent local attempts can reason about their own files without reading stale or unrelated issue state.
- Writing the decision state into the issue thread before the orchestrator has validated it would make malformed model output part of the persisted workflow state. With local artifacts, the orchestrator can validate JSON, rerun once for repair, and only then decide what user-visible comments or resume state should exist.

So hidden comments can still be useful for resume/history, and future work could store a summarized copy of validated artifact state there. But they are the wrong primary fix for this issue. This PR fixes the bug at the actual handoff point: each step writes local structured JSON, the orchestrator validates that JSON immediately, and only falls back to prose when the structured contract is unavailable.

## Testing

Local validation run on this branch:

- `pytest -q tests/test_e2e_change_cli_json_artifacts.py`
  - Result: `1 passed`
- `pytest -q tests/test_agentic_change_orchestrator.py`
  - Result: `221 passed`
- `pytest -q tests/test_e2e_issue_373_step5_keyerror.py`
  - Result: `11 passed`
- `pytest -q tests/test_e2e_change_cli_json_artifacts.py tests/test_e2e_issue_373_step5_keyerror.py`
  - Result: `12 passed`
- `pytest -q tests/test_e2e_issue_319_json_braces.py::TestIssue319JsonBracesE2E::test_e2e_change_with_json_in_issue_body tests/test_e2e_issue_319_json_braces.py::TestIssue319JsonBracesE2E::test_e2e_change_with_json_in_step_outputs tests/test_e2e_issue_319_json_braces.py::TestIssue319RegressionTests::test_orchestrator_context_substitution_order tests/test_track_cost.py::test_output_cost_path_via_param`
  - Result: `4 passed`
- `pytest -q tests/test_agentic_change_orchestrator.py -k "json_retry_preserves_initial or invalid_json_reruns_once_before_hard_stop or review_loop_exits_on_clean_json_even_when_prose_says_issues or step9_no_change_needed_json_stops_successfully_without_pr or orchestrator_happy_path or step13_pr_url_prefers_json_and_blocked_status"`
  - Result: `8 passed, 213 deselected`
- `pytest -q tests/test_e2e_issue_319_json_braces.py::TestIssue319JsonBracesE2E::test_e2e_change_with_json_in_issue_body tests/test_e2e_issue_319_json_braces.py::TestIssue319JsonBracesE2E::test_e2e_change_with_json_in_step_outputs tests/test_e2e_issue_319_json_braces.py::TestIssue319RegressionTests::test_orchestrator_context_substitution_order tests/test_e2e_issue_373_step5_keyerror.py::TestIssue373OrchestratorE2EIntegration::test_orchestrator_context_missing_key_error_message tests/test_track_cost.py::test_output_cost_path_via_param`
  - Result: `5 passed`
- `python -m py_compile pdd/agentic_change_orchestrator.py`
  - Result: passed
- `python -m json.tool architecture.json`
  - Result: passed
- `git diff --check`
  - Result: passed

High-level scenarios covered:

- The actual `pdd change` Click CLI command is exercised in `tests/test_e2e_change_cli_json_artifacts.py`.
- The test invokes the real command path: `pdd change --no-github-state --no-prompt-checkup https://github.com/owner/repo/issues/1871`.
- The test uses the real `run_agentic_change` function and the real `run_agentic_change_orchestrator`.
- Only external boundaries are mocked: GitHub issue fetch/comments, worktree setup, GitHub state/comment posting, and the LLM/agent call via `run_agentic_task`.
- The mocked issue is effectively `https://github.com/owner/repo/issues/1871`, with fake GitHub API JSON for title/body/comments.
- The mocked agent writes real local step artifacts during the CLI run, including `09_implement.json`, `11_review.json`, and `13_create_pr.json`.
- Step 9 returns prose with no `FILES_MODIFIED:` marker, but writes `09_implement.json` saying `prompts/structured_change_python.prompt` changed. The CLI succeeds and reports that file.
- Step 11 returns prose saying issues were found, but writes `11_review.json` with `status: clean`. The orchestrator trusts JSON, skips Step 12, and proceeds to Step 13.
- JSON artifact readers safely handle valid, missing, malformed, empty, and non-dict artifacts.
- Hard-stop JSON statuses map to the same stop behavior as the old prose parser.
- Step 9 JSON changed-file fields take precedence over legacy prose markers.
- Step 9 `no_change_needed` JSON exits successfully without creating a PR.
- Step 9 invalid/missing JSON triggers exactly one repair rerun.
- Step 9 falls back to the first successful legacy prose output when the JSON repair rerun still fails.
- Step 11 review-loop JSON can mark a review clean even when prose says issues were found.
- Step 11 falls back to the first successful `No Issues Found` prose output when the JSON repair rerun still fails, so Step 12 is skipped correctly.
- Step 13 prefers JSON PR URL/status when present.
- Step 13 falls back to the first successful prose PR URL when the JSON repair rerun still fails.
- Existing JSON-in-issue and JSON-in-step-output e2e regressions still pass; template JSON braces are not reinterpreted as format keys.
- The CI-reported Step 9 failure case now passes locally.

GitHub validation:

- Unit Tests / `Run Unit Tests`: passed in 19m7s.
- Unit Tests / `Public CLI Regression`: passed in 2m44s.
- Unit Tests / `Package Preprocess Smoke`: passed in 1m42s.
- `heal` after `/gcbrun`: passed.
- `auto-heal`: neutral/skipped because this is a fork PR and that helper cannot push healing commits back across forks; the actual `heal` job passed.

## Mergeability / Remaining Risk

The validation is enough for this PR to be mergeable. GitHub reports the PR merge state as clean after the latest push and `/gcbrun`.

The main remaining edge cases are provider-behavior edge cases:

- A provider may write syntactically valid JSON with semantically incomplete data. The validators now check required statuses and list fields, but they do not prove the model listed every real changed file. The existing git-status fallback still helps Step 9's empty-file case.
- A provider may write both JSON and contradictory prose. This PR intentionally trusts JSON first because that is the new machine contract.
- Future cleanup should not delete the prose parsers until fallback usage is measured near zero across providers.

The PR keeps the old parsers, keeps issue steering unchanged, and adds fallback-preservation tests for the JSON repair path, so the rollout risk is bounded.


niti-go:
This PR also closes promptdriven/pdd_cloud#2644.

gltanaka (changes requested):
Thanks for the detailed PR and the added coverage. The direction is needed: #1850 is a real recurring bug class, and moving the orchestrator's decision inputs from prose markers to JSON artifacts is the right shape. CI is green and I also ran the focused/local validation successfully:

- `pytest -q tests/test_e2e_change_cli_json_artifacts.py` -> passed
- `pytest -q tests/test_agentic_change_orchestrator.py` -> 221 passed
- `pytest -q tests/test_e2e_issue_373_step5_keyerror.py` -> 11 passed
- `python -m py_compile pdd/agentic_change_orchestrator.py` -> passed
- `python -m json.tool architecture.json >/dev/null` -> passed
- `git diff --check d3bbdf1...codex/pr-1871-review-head` -> passed

I do not think this should merge yet because the new artifact contract can read stale valid JSON from an earlier attempt and treat it as the current step's decision.

The review loop reuses the same `11_review.json` path each iteration. In `pdd/agentic_change_orchestrator.py`, the loop creates/uses the same artifacts dir, runs Step 11, then immediately reads `11_review.json` and only retries/falls back when `_valid_step_json(11, s11_json)` is false. If iteration 1 wrote valid `{"status": "issues_found"}` and iteration 2 forgets to overwrite the file while its current prose says `No Issues Found`, the old valid file is still present. The orchestrator sees valid JSON, skips the JSON-repair retry, ignores the current clean prose fallback, and keeps running Step 12 until max iterations.

I reproduced that locally with a mocked run: `step11_iter1` wrote `11_review.json` as `issues_found`; `step11_iter2` returned `No Issues Found` but did not write JSON. The labels executed were:

```text
step11_iter1, step12_iter1, step11_iter2, step12_iter2, step11_iter3, step12_iter3, step11_iter4, step12_iter4, step11_iter5, step12_iter5, step13
```

Expected behavior is that iteration 2 either retries for missing current JSON and then falls back to iteration 2's clean prose, or otherwise exits the review loop. This same stale-read pattern can also affect rerun/resume paths for stop-capable steps, for example a Step 4 clarification stop saves `last_completed_step = 3` and preserves artifacts for resume; if the resumed Step 4 attempt forgets JSON, the old `04_clarify.json` can keep forcing the prior stop instead of using the current output.

Please clear or version each step's expected artifacts before launching that step attempt, or otherwise validate that the JSON read after the agent returns was produced by the current attempt. For the review loop, either per-iteration artifact filenames or deleting `11_review.json` / `12_fix.json` before each iteration would address the immediate issue. The same guard should apply to JSON-retried and resumable decision steps. Please add regression coverage for at least the review-loop stale `11_review.json` case, and ideally the Step 4/7 clarification resume case too, so the fallback guarantee is locked down.

niti-go:
Date: 7/7

niti-go:
Addressed the review by making structured artifacts the trusted source of truth across retries/resumes/review loops.

What changed:
- Clear each step's JSON/Markdown artifacts before every outer attempt, JSON repair attempt, and internal provider retry via `before_attempt`, including routing fallback/escalation paths.
- Persist only accepted decision JSON in workflow state and prune/ignore stale JSON when steps are rewound or marked `FAILED:`.
- Recompute Step 9 manual-review lines from accepted Step 9 JSON, and keep review-loop manual-review lines separate so stale prose/old state cannot leak into Step 13.
- Harden Step 13 as side-effectful: single provider attempt, no generic JSON retry/provider fallback, deterministic recovery only via exact PR validation, and no posting of unaccepted Step 13 markdown.
- Validate PR URLs/recovery/existing-PR shortcuts against exact repo, OPEN state, base branch, head branch, non-empty PR head SHA, and remote/local SHA expectations; existing fallback branches are scanned and validated without comparing against the caller's unrelated checkout HEAD.
- Updated prompt files and `architecture.json` to match the code contract, including `run_agentic_task` signature sync.

Validation:
- `PDD_PATH=/Users/Niti/CS-Projects/PromptDriven/pdd-worktree-1850/pdd PDD_AGENTIC_PROVIDER=anthropic,google pytest -q tests/test_agentic_common.py tests/test_agentic_common_issue_813_anthropic_api_key_oauth_shadow.py tests/test_agentic_common_worktree.py tests/test_agentic_change_orchestrator.py tests/test_e2e_change_cli_json_artifacts.py tests/test_e2e_issue_373_step5_keyerror.py --maxfail=1` -> 905 passed, 3 warnings.
- `python -m py_compile ...`, `python -m json.tool architecture.json`, and `git diff --check` all pass.
- Repeated xhigh adversarial Codex subagent review loop is clean: final pass reported no substantial issues remain.

niti-go:
Review follow-up completed.

What changed:
- Made structured JSON artifacts the accepted source of truth and clear per-step JSON/Markdown before every artifact-producing attempt so stale files cannot drive retries, resumes, or review-loop iterations.
- Persist accepted step JSON in workflow state and hydrate from that state on resume instead of rereading fixed artifact paths from disk.
- Hardened Step 13 PR handling: single side-effecting attempt, exact repo/base/head/SHA validation, deterministic exact-head recovery, and fail-closed behavior for wrong/stale PR URLs.
- Disabled explicit JSON repair reruns and provider retry/fallback for side-effecting Steps 9, 10, 12, and 13 to avoid duplicate worktree mutations; kept bounded repair behavior for non-mutating decision steps such as Step 11.
- Synchronized code, tests, prompts, and architecture metadata, including the existing-PR guard contract.
- Hardened the full-suite cost-tracking test isolation issue that CI surfaced.

Validation:
- Local targeted suite: 1013 passed.
- Track-cost regression suite: 29 passed; focused auth-bridge interaction: passed.
- Final adversarial subagent pass: no substantial issues remain.
- Required CI is green on head af113d7f: Run Unit Tests, Package Preprocess Smoke, Public CLI Regression, and heal passed; auto-heal skipped neutrally.

niti-go:
@gltanaka I addressed the stale artifact/source-of-truth review on the latest head af113d7f and requested re-review. CI is green, the project card is back in In Review, and the follow-up summary is here: https://github.com/promptdriven/pdd/pull/1871#issuecomment-4910094995

In summary, you had pointed out a safety problem: the workflow could accidentally reuse an old file
  from a previous attempt and treat it like the result of the current attempt. So if an earlier
  review step said “issues found,” and a later step forgot to write a fresh result, the workflow
  might keep believing “issues found” even if the current run actually said “no issues found.” In
  plain terms: stale notes from an old run could steer the current run.

  We fixed that by making the workflow much stricter about what it trusts:

  - Before each step runs, it clears that step’s old result files so stale files cannot be
    reused.

  - It only trusts structured result files that were accepted for the current workflow state.
  - When resuming after a pause or retry, it uses the saved accepted state instead of blindly
    rereading old files from disk.

  - For steps that actually change files or create/update PRs, we prevented automatic reruns that
    could accidentally apply the same change twice.

  - We made PR creation safer by validating that the PR is the right repo, branch, base branch,
    and commit before treating it as successful.

  - We updated the tests, prompts, and architecture metadata so the code and documentation all
    describe the same behavior.

  The end result: the workflow is less likely to be confused by leftovers from earlier attempts,
  less likely to duplicate work during retries, and safer when creating or updating PRs.

gltanaka (approved):
Thanks for the update. I re-reviewed the latest head and the earlier stale-artifact concern looks addressed.

The change now treats accepted JSON step outputs as the source of truth, clears step artifacts before each attempt/retry, avoids JSON-repair reruns for mutating steps, and validates/recover-checks PR creation against the current head. That directly covers the failure mode where an old `11_review.json` could override a newer Step 11 prose result and incorrectly drive Step 12.

Validation I checked:
- Current CI is green for unit tests, package preprocess smoke, and public CLI regression.
- Local targeted stale-artifact/side-effect tests passed.
- Full `tests/test_agentic_change_orchestrator.py` passed.
- CLI-level e2e artifact tests and related issue regression tests passed.
- `agentic_common`, `ci_validation`, cost tracking, py_compile, JSON validation, and diff whitespace checks passed.

I do not see a remaining merge blocker. A live end-to-end run against real agents/GitHub side effects would be useful as a canary, but I do not think it is reasonable to require it before merge given the external cost/side effects and the mocked CLI e2e coverage around the relevant orchestration paths.

gltanaka (approved):
Re-approving the latest head after the merge-from-main update. The new commit is only the integration merge from current `main`; CI is green on `1f49cd4`, and the previously reviewed JSON-artifact/source-of-truth changes remain mergeable.

PR #1851, opened 2026-07-06
Epic #1209 v2: repo-bloat/context-window experiment — design revision, instrumentation, distractor scaling, runner, paper

sohni-tagirisa:
## What this epic is

Turns the merged repo-bloat pilot **design** (#1401, #1419) into a **runnable experiment + paper draft** for #1209, incorporating review feedback:

1. **SWE-bench excluded** from the core experiment (related-work only).
2. **Context snapshots** become the primary instrument (API-boundary recording proxy capturing the byte-exact context window per model request) — no longer relying only on file-level read tracing.
3. **Iteration-level token bloat** is a new pre-registered hypothesis (H2): does irrelevant-context share accumulate gradually across agent iterations within a run?
4. **PDD is the primary target repo** (pinned commit, dependency-sliced cores, seeded bugs); Codex CLI stays **unforked** — instrumented around, at the API boundary, to preserve external validity.
5. **Formal distractor-file definition** + deterministic generator (regrow / mutate / template modes) scaling to **50x and beyond**, with a breakdown probe to locate where performance collapses.
6. **Task performance vs. distractor scale** reported per size step (hidden-pass rate, tokens/time-to-success, failure classes).

## Sub-PRs (target this branch — merge order 1 → 2 → 3 → 4 → 5)

| # | Branch | Contents |
|---|--------|----------|
| 1 | `research/1209-v2-design-revision` | design.md v2 + README refresh |
| 2 | `research/1209-v2-context-instrumentation` | `harness/context_snapshots/` — proxy, schema, session parser, iteration analyzer + tests |
| 3 | `research/1209-v2-distractor-generator` | `harness/distractors/` — DEFINITION.md, 3-mode generator, token-budgeted manifests + tests |
| 4 | `research/1209-v2-runner` | `harness/runner/` — variant builder, orchestration, metrics, report, mock end-to-end (stacked on 2+3) |
| 5 | `research/1209-v2-paper` | `paper/paper.md` — full draft matching v2 methodology |

See `research/repo-bloat-benchmark/epic-1209-v2.md` for the full decision table and rationale.

**Do not merge this PR until all sub-PRs land and the epic is reviewed.**

Closes nothing yet; tracks #1209.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1901, opened 2026-07-07
Research #1209 v2 (12): resolve the 4 pilot blockers + adversarial hardening

sohni-tagirisa:
Resolves the four remaining pilot-readiness blockers from the [audit verdict](https://github.com/promptdriven/pdd/pull/1851#issuecomment-4908219734), then hardens each against **two adversarial review agents**. Stacked on #1897 → #1896 → #1891 → #1890 (this PR is based on #1897's branch, so its diff is the blocker-resolution delta only). Targets the epic via that stack, never `main`. **Zero model tokens; no billed API calls.**

## The 4 blockers — resolved
1. **Scenarios 2 & 3 authored** (`pdd-failure-classification` — timeout/syntax precedence inversion; `pdd-language-extensions` — duplicate-CSV-row last-write-wins), each per the #1897 template with committed, locked ladder manifests 1x–50x. All three pilot scenarios' preflights PASS at every step. The three seed-novelty classifications span a deliberate recall-risk gradient (see `scenarios/README.md`).
2. **Pre-registration frozen**: `pilot_arm_codex.json` (codex-cli 0.142.4, gpt-5.1-codex-max, medium effort, 272k context window) + `register_env` CLI emitting the machine-bound fingerprint + `PREREGISTRATION.md` rationale.
3. **Kernel-level egress tier** (`harness/runner/container/`): three network namespaces so the agent container has **no route to the gateway or the outside**; `egress_check.py` verifies it empirically per role.
4. **First-real-run calibration** (`first_run_calibration.py`): GO/NO-GO over the one paid calibration run's artifacts; performs no network I/O itself.

## Adversarial review — confirmed findings, fixed
Two agents attacked the resolutions. What survived scrutiny and was fixed:
- **BLOCKER (scenarios):** scenario 3's hidden verifier was gameable — an agent could delete the duplicate CSV row (the CSV was inside `allowed_edit_globs`) or hardcode `yaml=yml` and pass. Rewrote it to load the agent's fixed module against its **own crafted CSV** (two duplicated languages, one absent from `task.md`) through the public API; narrowed `allowed_edit_globs` to `src/pdd/*.py`. Both vectors now demonstrably FAIL.
- **Cross-scenario leak:** each distractor pool shipped sibling scenarios' *solved* target modules (incl. a denylisted oracle line). Replaced all pools with neutral non-target modules; a regression test now forbids any target module or any scenario's denylist string in any agent-visible tree.
- **Verifier determinism:** visible/hidden runs inherited the outer repo's `pytest.ini` and `PYTEST_ADDOPTS`. Added a local `pytest.ini` per `hidden/`+`core/` and scrub `PYTEST_*` from the verifier env.
- **Cell resilience:** a nonzero agent exit now records a failed run (`agent_error`) and the cell **continues** instead of aborting every later trial.
- **Container recording-bypass:** the agent previously shared the runner netns and could reach the gateway directly; the 3-netns redesign closes it at the kernel level (not just via the env guard).
- **register_env / calibration integrity:** `--skip-binary-check` now yields a non-authoritative record; calibration requires a parsed response per snapshot and rejects dev-only registrations; the env fingerprint now covers the egress-guard constants; denylist matching is whitespace/case-normalized.

## Tests run (all zero-billing)
- Full harness suite: **144 passed, 1 skipped** (stdlib+pytest, loopback only).
- All three scenarios' `preflight` (§4.1.2 + §6.6): PASS at every ladder step.
- Live routing probe on `codex-cli 0.142.4`: **CONFIRMED** (11/11).
- Gaming-vector checks: both degenerate fixes on scenario 3 now FAIL the hidden verifier.

## Intentionally NOT run (would bill or needs a Docker host)
- Any real-provider request (the one paid calibration request is the documented first step once a key is authorized).
- The 90-run pilot.
- `egress_check` on a live Docker host — the container tier is *implemented, field-verification pending*; both role checks are mandatory before a pilot session (stated in `container/README.md`).

With this merged, the epic is **pilot-ready pending three operational steps** (install pinned CLI + register; run both egress checks on the Docker host; one calibration request → GO), none of which are code work.


gltanaka (changes requested):
Requesting changes. The PR has been updated since the prior #1851 audit and the scenario side is in much better shape: I locally ran `python -m pytest research/repo-bloat-benchmark/harness -q` (`144 passed, 1 skipped`) and independently ran the full six-step oracle preflight for all three committed scenarios (`1x/2x/5x/10x/20x/50x`, all PASS). I do not think the paid calibration request, the full 90-run pilot, or paid provider e2e should be required before merging this code slice.

I do think the no-billing Docker/container path is reasonable to require before merge because this PR claims to resolve the prior hard-tier egress blocker, and the current implementation is not wired end-to-end:

1. **The hard-tier network isolation is not actually applied by the runner.** `ExperimentRunner` still launches the command arm with `subprocess.run(...)` in the current process namespace (`research/repo-bloat-benchmark/harness/runner/runner.py:271`). In the compose topology, the current process is the `runner` service, which is attached to `rb-proxy-egress` and can reach the gateway (`container/docker-compose.yml:58`), while the isolated `agent` service is separate (`container/docker-compose.yml:83`). A normal `agent_command: ["codex", ...]` therefore still runs in the runner namespace and can bypass the recording proxy at the kernel level. Please add a first-class launcher/path that runs the Codex process inside the isolated `agent` service/container, plus a no-billing compose/integration check proving the agent process cannot reach the gateway directly and can reach the recording proxy.

2. **The recording proxy endpoint is not usable from the separate agent container as written.** `RB_PROXY_HOST=0.0.0.0` is loaded as `RunConfig.proxy_host` (`harness/runner/cli.py:48`), passed as the proxy bind host (`harness/runner/runner.py:217`), and `RecordingProxy.start()` returns that same bind wildcard as the base URL (`harness/context_snapshots/proxy.py:228`). That URL is then written into the frozen agent config/env (`harness/runner/env_freeze.py:246`). A process in the separate agent container needs an advertised URL like `http://runner:<actual-port>`, not `http://0.0.0.0:<port>`. Split bind host from advertised host/base URL and cover the container case with a regression test.

3. **The documented egress check does not validate the pilot proxy path.** `egress_check.py` defaults to probing `runner:8080` (`harness/runner/container/egress_check.py:38`), but the actual `RecordingProxy` binds port `0` and uses the chosen ephemeral port (`harness/context_snapshots/proxy.py:228`). The compose file does not start any listener on `runner:8080`; the runner image default command is just the egress checker. Please make the check use the same live proxy endpoint that the frozen agent receives, or start an explicit no-billing dummy/live proxy for the check.

4. **Dispatching into `agent` would not see the frozen home/report volume yet.** `FrozenEnvironment` writes `config.toml` under `report_dir/codex_home` and passes that path as `CODEX_HOME` (`harness/runner/env_freeze.py:207`, `:244`). In compose, the runner mounts `rb-reports` at `/bench/reports`, but the agent service only mounts the source tree (`container/docker-compose.yml:74`, `:92`). Once the launcher runs inside `agent`, the generated `CODEX_HOME` will not be visible unless the agent shares the same run/report volume or the home is created in an agent-visible location.

5. **First-run calibration can currently GO on a failed Codex process.** This PR intentionally records nonzero agent exits as `agent_error` instead of aborting (`harness/runner/runner.py:293`; recorded at `harness/runner/metrics.py:271`), but `first_run_calibration.py` only checks snapshots/model/usage/fingerprint/evidence flags (`harness/runner/first_run_calibration.py:68`). A run with valid API snapshots and `agent_error: true` can still return GO. Please require `agent_error` to be false, and ideally verify `agent_process.json.returncode == 0`, with a regression test.

6. **Scenario metadata has incorrect core paths for the new pytest configs.** The variant builder copies the contents of each scenario's `core/` directory directly into the materialized workdir (`harness/runner/variant_builder.py:56`), so `core/pytest.ini` becomes `pytest.ini`. The three `core_files.txt` files list `core/pytest.ini` instead (`scenarios/pdd-find-section/core_files.txt:7`, `scenarios/pdd-failure-classification/core_files.txt:5`, `scenarios/pdd-language-extensions/core_files.txt:6`). Please correct those entries and add a small regression that compares `core_files.txt` against the actual files under each scenario's `core/` tree.

Once the container runner path is genuinely end-to-end, a zero-billing Docker compose verification is the right e2e bar for this PR. The paid provider calibration and full pilot can remain documented pre-pilot operational gates.

Issue #1903, opened 2026-07-07
PDD tests must land where the runner looks and auto-adopt existing tests (never fork, never overwrite, never block)


Serhan-Asad:
## Problem

PDD can write a module's test to a path the project's **test runner does not collect**, and when a test for that module already exists elsewhere, PDD writes a **second, parallel test** instead of maintaining the real one. Result: `pdd generate`/`change`/`sync` report "tests pass" (they ran/updated PDD's own copy) while the project's actual CI test for the same module goes stale and fails. PDD is structurally blind to the tests that gate the PR.

This hits any project whose test layout is idiomatic for its ecosystem (co-located `__test__/*.test.tsx` for jest/Next.js, etc.) rather than PDD's flat `tests/test_<name>` default.

## Live evidence (promptdriven/pdd_cloud, 2026-07-07)

- `pdd change` on the Contributions page → PR #2987; its verifier reported tests green.
- CI `pr-tests` **failed** on the only real, co-located suite: `frontend/src/app/contributions/__test__/page.test.tsx`.
- PDD had maintained a shadow copy at repo-root `tests/test_frontend_src_app_contributions_page.tsx`; jest `rootDir` is `frontend/`, so **jest never runs root `tests/`**. PDD passed its shadow, never saw the real test the change had broken.
- Scope: 50 real co-located jest tests under `frontend/src/app/**/__test__/`, vs 6 orphan PDD tests parked in root `tests/`. Systemic, not one file.

## Why current behavior exists (do not naively "fix" this)

The current design is intentional, and a blind change would break things:
1. Default `test_dir = "tests/"` (`agentic_change_orchestrator.py:1620`) is **pytest-idiomatic** — correct for Python (PDD's primary language); it only misfires on jest/TS co-location.
2. PDD **regenerates** tests from the prompt, so it writes to a PDD-controlled path *by design* to avoid clobbering human-authored tests it doesn't own.
3. `TestChurnError` (#1012) already blocks rewriting an existing test by >40% and routes through the `PDD_REPAIR_DIRECTIVE` self-repair loop — the system is deliberately built NOT to silently destroy human tests.
4. `_sibling_test_paths` (#876) already detects co-located sibling tests, but only for Python (`test_{stem}.py`) and only for mock-detection — not for choosing the write path, and not for TS.

**Hard constraint for any fix: PDD must never silently overwrite a human-authored test, and must never hand work back to the user by failing/blocking a command.** The value proposition is that the commands take care of everything.

[...]

Issue #876, opened 2026-05-09
Support compressed marketplace few-shot example context


gltanaka:
Parent: #1411

## Goal

Compress marketplace few-shot/example context before injecting it into PDD prompts, preserving behavioral mold while avoiding full example dumps.

## Prototype links

- Full/compressed few-shot experiment: https://github.com/promptdriven/pdd_cloud/commit/512e722a4c95d01a4828d2d69ab14a6ed34141e8
- Contract-source preservation fix: https://github.com/promptdriven/pdd_cloud/commit/f54f2ef6a7a71a10e8dd003a244414cd3f7fe075
- Experiment doc: https://github.com/promptdriven/pdd_cloud/blob/benchmark/live-prompt-test-traces/docs/grounded-pdd-attempt-2026-05-04.md

## Evidence

Local Qwen without few-shot failed. Full worker-app few-shot passed but used 36,694 traced input tokens. Contract-preserving compressed few-shot also passed and used 27,143 traced input tokens, a 26% reduction versus full few-shot grounding.

## Acceptance criteria

- Example compression keeps prompt intent, selected test contracts, code slices, and contract-bearing symbols.
- Compression is deterministic and records source hashes/token counts.
- Compression can fall back to full examples when confidence is low or required slices cannot be resolved.
- Regression tests prove compressed examples preserve externally referenced helpers and patch targets.

Parent epic: #873. Adjacent to #794 (grounding budget — ranks/caps/dedupes); this issue is about *compressing* the few-shot content itself.

---
*Migrated from [gltanaka/pdd#1414](https://github.com/gltanaka/pdd/issues/1414) (originally filed 2026-05-08 by `gltanaka`). gltanaka/pdd is deprecated.*

gltanaka:
Cross-referenced with promptdriven/pdd_cloud#1387 — cloud #1387 is the cloud-side counterpart that adds safe budgeting, deterministic compression, and replay regression for marketplace few-shot in `generateTest`.

Issue #342, opened 2026-01-20
pdd test should generate tests that isolate local code under test


jamesdlevine:
# `pdd test` should generate tests that isolate local code under test

## Summary

Tests generated by `pdd test` don't set up the Python path to prioritize the local repository code. If the package is also installed in site-packages, tests import and run against the **installed version** rather than the **local code under test**. This breaks `pdd fix`, CI workflows, and any other situation where tests are run against local modifications.

## Problem

Currently, generated tests simply import the module:
```python
from foo.module import SomeClass
```

This relies on Python's default import resolution, which finds site-packages before the local directory. The generated tests should ensure the repository root is first in `sys.path`:

```python
import sys
from pathlib import Path

# Ensure local repo code takes precedence over installed packages
_repo_root = Path(__file__).resolve().parent.parent  # adjust depth as needed
sys.path.insert(0, str(_repo_root))

from foo.module import SomeClass
```

## Why This Matters

By fixing test generation, the isolation is **baked into the test file itself**. This means:
- Tests work correctly regardless of how they're invoked (pytest, make, CI, IDE, `pdd fix`)
- No need to modify every PDD command that runs tests
- No reliance on external PYTHONPATH configuration
- Self-documenting: the test file explicitly shows what code it's testing

## Reproduction

1. Have a project with package `foo` installed in site-packages
2. Run `pdd test` to generate tests for `./foo/module.py`
3. Generated test imports `from foo.module import ...`
4. Run pytest - imports from site-packages, not local code
5. Modify `./foo/module.py` locally
6. Run pytest again - still tests the old site-packages version

**Verification:**
```python
# Add to generated test to see the problem
from foo.module import SomeClass
print(f"Testing code from: {SomeClass.__module__}")
# Prints: foo.module (site-packages)
# Should print path to local ./foo/module.py
```

## Real-World Example

From a `pdd fix` run on issue #333:

The local file was correctly modified:
- `./pdd/agentic_bug_orchestrator.py` - contains the fix

But tests imported from site-packages:
- `/home/james/miniconda3/.../site-packages/pdd/agentic_bug_orchestrator.py` - no fix

Result: 5 fix iterations + 2 agentic fallback attempts all "failed" despite generating correct fixes. The tests were validating the wrong code.

## Impact

- **`pdd fix` broken**: Correct fixes fail validation because tests run against installed code
- **Silent failure mode**: Users see "fix failed" but the fix was actually correct
[...]

PR #1914, opened 2026-07-08
fix(test): adopt the runner-collected co-located test as PDD's canonical test path (#1903)

Serhan-Asad:
## What & why

**Part of #1903** (scoped — does not fully close it; see "Deliberately out of scope" for the greenfield / never-block / evidence follow-ups). PDD derived a module's canonical **test-output path** purely from `.pddrc test_output_path`/defaults (`tests/test_{name}{ext}`), **blind to the project's actual test runner**. On a jest/Next.js project (jest `rootDir=frontend/`) the real test is co-located (`frontend/src/app/contributions/__test__/page.test.tsx`), but PDD derived and then wrote/verified a repo-root **shadow** (`tests/test_..._page.tsx`) jest never collects. Result: `pdd generate`/`change`/`sync` report "tests pass" (PDD ran its own shadow) while the real test the change broke goes stale — a false-green. (Secondary symptom fixed too: a `.tsx` module derived a `.ts` test extension.)

## The fix — adopt the runner-collected co-located test

When a module already has a **single, unambiguous, existing co-located test**, PDD now **adopts that file as its canonical test path** so generate/change/sync target the real test CI runs instead of maintaining a shadow. This changes only the *derived path*, never the write contract or the churn safety net.

- **`pdd/content_selector.py`** — pure detector `find_collocated_test(code_file)`: Python via the existing `_sibling_test_paths`; JS/TS jest/vitest/Next.js `{stem}.test|spec.{ts,tsx,js,jsx,mjs,cjs}` beside the module and under `__test__/`, `__tests__/`, `tests/`. Existing files only, module excluded, resolved-dedup, **single match else `None`** (0 or >1 → no retargeting), **suffix-gated** (unsupported languages → no adoption), never raises. Plus the adopter `resolve_test_output_path(...)` and the provenance predicate `_pins_test_output_location(...)`/`configured_test_output_pinned(...)`.
- **`pdd/sync_determine_operation.py`** — `get_pdd_file_paths` is a thin wrapper over the original derivation (`_get_pdd_file_paths_uncollocated`) that adopts across **every** return branch (arch.json / .pddrc / fallback), covering `sync` and `change`.
- **`pdd/cmd_test_main.py`** — after `construct_paths`, adopts on the single `output_file_paths['output']` source of truth (covers direct `pdd test`); keeps the native/cloud generation destination in sync with the adopted write target; and, in sync's `merge`-into-existing case, pins the write to `existing_tests[0]` so `construct_paths`' `force=False` dedup can't divert the regenerated test into a numbered `page.test_1.tsx` shadow beside the real one.
- **`pdd/code_generator_main.py`** — `_is_test_output_path` now also recognizes `.test/.spec.mjs/.cjs` and the singular `__test__` dir, so an adopted human test is always covered by the `TestChurnError` guard.
- **`pdd/evidence_manifest.py`** — `resolve_test_output_paths` applies the same adoption after `construct_paths`, so `pdd test --evidence` records the adopted co-located test (the file actually written/run), not the shadow.

### Explicit-vs-default provenance (the subtle part)

Adoption only fires when PDD was using its **derived default** path. The user's pin is honored — CLI `--output`, `PDD_TEST_OUTPUT_PATH`, or an explicit `.pddrc test_output_path`/`outputs.test.path` (detected by **presence, not truthiness**, so an explicit `""` root-level output is respected). Crucially, provenance is read from the **raw `.pddrc` context defaults**, resolved from the **same context that produced the path** (prompt-side/`prompts_dir`, `--context`, or the arch branch's own code-path context) — **never** from `construct_paths`' `resolved_config`, which injects a generated-default `test_output_path` and would always read as pinned.

## Deliberately out of scope (kept minimal; documented follow-ups)

Per the design review, this is the smallest correct core. It does **not**: parse jest/pytest runner config (roots/testMatch/rootDir/testpaths); do agentic adopt-and-merge; implement the issue's "never-block/never-fail" semantics (which conflict with the existing `TestChurnError` safety contract); change Python defaults. These remain **tracked follow-ups to #1903** (this PR is intentionally partial and does not close the issue): **(A) greenfield** runner-config detection (jest `roots`/`testMatch`/`rootDir`, pytest `testpaths`) so a *fresh* project gets the co-located location from the start — this PR only adopts an *existing* co-located test — tracked in **#1916**; **(B) never-block-on-churn** — open the PR and flag the test for review instead of raising `TestChurnError`, deferred because it conflicts with the current safety contract (silently overwriting a human test is worse) and needs its own design + review — tracked in **#1917**. (The `pdd test --evidence` manifest resolver is **aligned in this PR** — `evidence_manifest.resolve_test_output_paths` now applies the same adoption, so the manifest records the file PDD actually writes, not the shadow; **#1918** is addressed here.) A minor Python symlink-loop-first-candidate edge in `_sibling_test_paths` (#876) remains (standard `src/foo.py → tests/test_foo.py` unaffected; not the #1903 JS/TS path).

## Safety / no-regression

- Standard Python flow is a **no-op**: `src/foo.py` → `tests/test_foo.py` (detector returns exactly the derived path → unchanged). Verified through the real `get_pdd_file_paths`.
- Honors `--output`, env, and explicit `.pddrc` pins (incl. `prompts_dir`/`--context`-selected and arch.json code-path contexts).
- Ambiguity guard: >1 co-located test → no change. Unsupported language → no change.
- Does **not** bypass `_verify_test_churn`/`TestChurnError` — writes now diff against the real test content on the adopted path (the safety net working; never a silent overwrite of a human test outside the guard).

## Tests

New `tests/test_issue_1903_adopt_collocated_test.py` (TDD; every regression verified fails-before/passes-after). Integration tests drive the **real** `construct_paths` / a real `architecture.json` (a mock hides the `resolved_config` injection this fix routes around). Coverage: detector correctness (incl. `page.tsx`→`__test__/page.test.tsx`, 0/>1 → None, cross-language safety, bad paths); adopter gating; `get_pdd_file_paths` end-to-end (default jest → adopts; explicit `test_output_path`/`outputs.test.path`/env/`prompts_dir`-anchored/arch-code-path → not adopted); `cmd_test_main` (retarget when unpinned, honor `--output`/pin, prompts_dir-context pin, generation destination matches the adopted path in both self-retarget and sync-explicit-output paths); churn-guard coverage for `.mjs/.cjs`/`__test__`; Python-default no-op.

Full targeted + regression-trap suites green (`test_issue_1903_adopt_collocated_test`, `test_content_selector`, `test_sync_determine_operation`, `test_cmd_test_main`, `test_code_generator_main`, `test_evidence_manifest`, `test_generate_output_paths`(+regression), `test_construct_paths`, `test_agentic_change_orchestrator`), plus broader touched-area suites (`test_sync_orchestration`, `test_one_session_sync`, `test_agentic_test_generate`, `test_generate_test`, `test_sync_main`, `test_update_main`, `test_increase_tests`) — all pass (1 pre-existing skip).

## Relationship to other work

Companion pdd_cloud config fix #3024 reconciles the repo's explicit-`tests/` orphan configs; this PR fixes the tool's default-path blindness. Independent of the parallel "warn-only" exploration for #1903 (which detects+warns but does not retarget); this PR retargets, eliminating the false-green rather than only surfacing it.


Serhan-Asad:
## CodeQL failure analysis — real finding, fix available to cherry-pick

The failing **CodeQL** check is not infra noise: the Analyze jobs pass, and the 2s "fail" is the aggregated result-poster reporting **4 new high-severity `py/path-injection` alerts (#339–#342, CWE-022)** introduced by this PR:

- `pdd/content_selector.py` — `find_collocated_test` (candidate derivation), `resolve_test_output_path` (adoption)
- `pdd/sync_determine_operation.py` — `_adopt_collocated_test` (×2)

**Why it matters:** the caller-supplied code-file path is issue-influenced in the hosted/agentic flow, and the adopted path becomes the generated-test **write target** (`result['test']`/`['test_files']`). A traversal-shaped input could retarget test generation outside the repo.

**Fix (validated on the E2E stack branch):** containment — candidates that resolve outside the working tree are rejected in `find_collocated_test`, and both adoption sinks re-assert containment before retargeting; traversal inputs degrade safely to no-adoption. Includes 4 containment regression tests; 5 existing tests updated to run with `cwd=`project (matching real CLI invocation).

Cherry-pick from the stack branch: `promptdriven/pdd@fd107724a` ("Contain collocated-test adoption to the working tree").

Suites after the fix: `test_issue_1903_adopt_collocated_test.py` 45 passed · `test_issue_1900_surface_contract.py`+`test_ci_validation.py` 142 passed · `test_sync_orchestration.py` 215 passed · zero new ruff findings vs main baseline.

Serhan-Asad:
Audit 2026-07-21: closing this PR as superseded by #1998, the canonical PDD staging campaign bundle containing or replacing this slice. The branch and review history remain recoverable here; reopen only if #1998 drops this behavior.

PR #1889, opened 2026-07-07
EPIC: user stories as executable regression oracles in the PDD CLI (#1816)

sohni-tagirisa:
## EPIC integration PR for #1816 — user stories as executable regression oracles

Closes #1816.

This is the integration PR for the epic branch `codex/issue-1816-story-regression-epic` (101 commits, all landed via reviewed sub-PRs). It brings **user stories as regression tests into the PDD CLI**, independent of the pdd_cloud GitHub-app work.

### What's in it
- **`pdd story` CLI** (`add`, `link`, `list`, …) for managing user stories in the regression suite (#1774, #1772)
- **`@pytest.mark.story` marker + traceability API** — stories claim tests via `story_id` markers (#1713)
- **`pdd test --from-story`** — deterministic regression-test generation from a story: behavioral test when the story has an `## Entry Point` contract (#1867), text-pin fallback otherwise; malformed stories are rejected instead of emitting vacuous tests (#1863)
- **Regression gate** — `evaluate_stories()` classifies stories ok/stale/missing; warn mode (non-blocking, collect-first) with strict mode available (#1805, #1712)
- **Seeded stories for top PDD flows** + cross-dev-unit story coverage (#1800, #1775)
- **Verification fixes from the #1857 E2E pass**: explicit `--prompt` authoritative in `story link` (no LLM call, #1859), stale top-flow seed hashes refreshed (#1860), ClickException error reporting (#1861), stale duplicate markers can't shadow fresh claims (#1864)

### Verification (issue #1857 — 🟢 GO posted)
- `pytest -m story`: 16/16 pass; story feature suite 128 passed / 1 skipped
- Gate at tip: **9 ok / 0 stale / 23 missing** (missing = collect-first, expected)
- Oracle-mutation red/green verified; zero-credential offline run verified

### Notes for the maintainer
- Known conflict vs `main` in `pdd/prompts/commands/__init___python.prompt` — a reconcile sub-PR against this branch will resolve it.
- Do not squash sub-PR history; branch is merge-commit structured.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1699, opened 2026-06-22
Story regression foundation: @pytest.mark.story marker + story→test traceability API


sohni-tagirisa:
Part of EPIC #1698 — User stories as executable regression oracles.

## Summary

Establish the foundation that lets a `user_stories/story__*.md` file be backed by **runnable, traceable** pytest regression tests. No test generation yet (that is sub-issue 2) — this issue defines the model, the marker, and the discovery/traceability API.

## Scope

- Add a `@pytest.mark.story(story_id=...)` marker (register in `pytest.ini` alongside `e2e`/`integration`/`slow`/`real`/`private_prompt`).
- Define a `story_id` convention derived from the `story__<name>.md` filename, stable across renames where practical.
- Traceability API in `pdd/user_story_tests.py` (or a new `pdd/story_regression.py`): given a story file, resolve the set of regression tests that claim it (via marker), and given a test, resolve its story.
- Extend the existing coverage matrix (`pdd checkup coverage`, `docs/coverage_contracts.md`) to surface a new dimension: **does this story have an executable regression test?** (distinct from the existing `story-only` / `test-only` rule statuses, which today key off `test_R<n>` naming).

## Out of scope

- Generating the tests (sub-issue 2).
- The Make/CI lane (sub-issue 3) and the gate (sub-issue 4).

## Acceptance criteria

- `pytest -m story` selects exactly the story-backed regression tests and nothing else.
- Given a story file, the API returns its linked regression tests; given a test node id, it returns the owning story.
- `pdd checkup coverage <prompt>` shows, per story, whether an executable regression test exists.
- Unit tests cover the marker registration, the id convention, and the bidirectional resolution.

sohni-tagirisa:
Resolved in EPIC branch codex/issue-1698-story-regression-epic. The story marker + traceability foundation is integrated; sub PR #1713 is closed as already included. Verified via make regression-stories and targeted story/coverage tests on the epic branch.

PR #1945, opened 2026-07-08
fix(story): unify story-hash freshness semantics across gate evaluators

sohni-tagirisa:
Ships a focused correctness sub-PR for the story-regression epic (references PR #1889).

## Root cause — "two of everything"
The epic ships **two hash primitives, two evaluators, two AST scanners** for story-hash freshness, with silently divergent semantics. An adversarial reviewer confirmed all three bugs with reproductions. This PR unifies them so they cannot diverge again.

## Bug 1 (P1) — the only CLI-wired gate uses incompatible hash semantics (wrong in BOTH directions)
`evaluate_story_regression` (the ONLY evaluator wired to a CLI: `pdd checkup coverage --story-regression-gate`) compared **only** `story_bundle_hash` against **only** module-level `STORY_HASH`-constant regex matches, ignoring marker `story_hash=` kwargs — while the full gate `_classify_story_from_markers` accepts content-hash ∪ bundle-hash from kwargs OR constants.
- **Direction A (false STALE):** a test recording the content hash (behavioral generator / documented kwarg form) reported stale when a contract made the bundle hash differ.
- **Direction B (false PASS):** a bogus `story_hash="0000…"` kwarg was never hash-checked and passed.

**Fix:** both evaluators now share `_acceptable_story_hashes` (content ∪ bundle), and `evaluate_story_regression` gathers recorded hashes via the gate's own AST scan (`_recorded_story_hashes`) — resolving marker kwargs AND module constants. Legacy hashless markers stay "present" (#1699 compat); a marker WITH a non-matching hash is stale. The dead `_recorded_hash`/`_HASH_ASSIGN_RE` regex path is removed.

## Bug 2 (P1) — metadata-only relink falsely stales bundle-recording tests
`story_bundle_hash` hashed RAW story text including the `<!-- pdd-story-prompts … -->` / `<!-- pdd-story-dev-units … -->` comments, so `pdd story link` / `story add --update` flipped the bundle hash → false stale.

**Fix:** hash `_normalized_story_for_hash(story) + normalized(contract)` so a metadata-only relink is invariant; the contract stays in the hash. The text-pin generator records `story_bundle_hash` and its inline `_bundle_hash()` reuses the canonical primitive so recorded value and gate check can never drift.

## Bug 3 (P2) — static fallback can't resolve constant-name story_ids
`_scan_story_markers_static` only accepted literal `story_id`s, missing generated `story_id=STORY_ID` / `PDD_STORY_ID` module constants when pytest collection yields zero markers.

**Fix:** the static fallback resolves module-level string constants, mirroring the gate's AST resolver.

## Refreshed seed bundle-hash constants
Only where the normalization changed the value (sync/fix/change already recorded the unchanged content hash):

| story | old | new |
|---|---|---|
| pdd_update | cf0bc4a2189f81bd | af77d101cb8515bd |
| pdd_connect | 941636e0fbab930c | 263eed0bff2eb6f9 |
| pdd_test | 254412e41b69ae19 | 8d7c10cc1b2ca908 |
| pdd_generate | e0e44229295fc640 | e23eb1e50140fe63 |
| pdd_checkup | 213cefce7c9bb9de | c182aa91b22551eb |
| pdd_bug | e04067b53306ba0f | 19e588bbffc02cf1 |
| top_flows pdd_generate | 3660bcd6d6d7e8a8 | e23eb1e50140fe63 |
| top_flows pdd_update | cf0bc4a2189f81bd | af77d101cb8515bd |

## Verification
- New tests red→green (Bug 1 A/B, Bug 2 relink, Bug 3 constant fallback).
- `pytest -m story`: **15 passed**.
- `evaluate_stories` over the repo: **0 stale**, 10 seed stories ok.

### DIFF
--- a/pdd/story_regression.py
+++ b/pdd/story_regression.py
@@ -98,15 +98,48 @@ def _relative_nodeid(item: pytest.Item, root: Path) -> str:
         return item.nodeid.lstrip("/")
 
 
-def _literal_story_id(node: ast.AST) -> Optional[str]:
-    """Return a story id from a literal decorator argument, if present."""
+def _module_string_constants(tree: ast.AST) -> Dict[str, str]:
+    """Return top-level ``NAME = "literal"`` string constants in a module.
+
+    Mirrors the gate's own AST constant resolver so the static fallback can
+    resolve ``story_id=STORY_ID`` / ``story_id=PDD_STORY_ID`` module constants
+    that generated tests emit (pdd#1889 Bug 3), not only bare string literals.
+    """
+    constants: Dict[str, str] = {}
+    for node in getattr(tree, "body", []):
+        if isinstance(node, ast.Assign):
+            targets = list(node.targets)
+            value = node.value
+        elif isinstance(node, ast.AnnAssign):
+            targets = [node.target]
+            value = node.value
+        else:
+            continue
+        if not (isinstance(value, ast.Constant) and isinstance(value.value, str)):
+            continue
+        for target in targets:
+            if isinstance(target, ast.Name):
+                constants[target.id] = value.value
+    return constants
+
+
+def _literal_story_id(
+    node: ast.AST, constants: Optional[Dict[str, str]] = None
+) -> Optional[str]:
+    """Return a story id from a decorator argument (literal or module constant)."""
     if isinstance(node, ast.Constant) and node.value is not None:
         return _normalize_story_ref(node.value)
+    if constants is not None and isinstance(node, ast.Name):
+        resolved = constants.get(node.id)
+        if resolved is not None:
+            return _normalize_story_ref(resolved)
     return None
 
 
-def _story_id_from_decorator(decorator: ast.AST) -> Optional[str]:
-    """Resolve ``@pytest.mark.story`` decorator literals without executing code."""
+def _story_id_from_decorator(
+    decorator: ast.AST, constants: Optional[Dict[str, str]] = None
+) -> Optional[str]:
+    """Resolve a ``@pytest.mark.story`` decorator's story id without executing code."""
     if not isinstance(decorator, ast.Call):
         return None
 
@@ -124,16 +157,19 @@ def _story_id_from_decorator(decorator: ast.AST) -> Optional[str]:
 
     for keyword in decorator.keywords:
         if keyword.arg == "story_id":
-            return _literal_story_id(keyword.value)
+            return _literal_story_id(keyword.value, constants)
 
     if decorator.args:
-        return _literal_story_id(decorator.args[0])
+        return _literal_story_id(decorator.args[0], constants)
 
     return None
 
 
 def _scan_story_markers_static(tests_dir: Path) -> StoryTestMap:
-    """Best-effort static fallback for literal ``@pytest.mark.story`` decorators."""
+    """Best-effort static fallback for ``@pytest.mark.story`` decorators.
+
+    Resolves both string-literal and module-constant ``story_id`` forms.
+    """
     story_to_tests: Dict[str, Set[str]] = {}
     test_to_stories: Dict[str, Set[str]] = {}
     root = tests_dir.resolve()
@@ -149,6 +185,7 @@ def _scan_story_markers_static(tests_dir: Path) -> StoryTestMap:
         except (ValueError, OSError):
             rel_path = path.name
 
+        constants = _module_string_constants(tree)
         for node in ast.walk(tree):
             if not isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
                 continue
@@ -157,7 +194,7 @@ def _scan_story_markers_static(tests_dir: Path) -> StoryTestMap:
             story_ids = {
                 sid
                 for decorator in node.decorator_list
-                for sid in [_story_id_from_decorator(decorator)]
+                for sid in [_story_id_from_decorator(decorator, constants)]
                 if sid
             }
             if not story_ids:
--- a/pdd/story_regression_gate.py
+++ b/pdd/story_regression_gate.py
@@ -33,7 +33,6 @@ from __future__ import annotations
 
 import ast
 import logging
-import re
 import shutil
 import subprocess
 from dataclasses import dataclass
@@ -78,10 +77,6 @@ _GATE_MODES = frozenset({"off", "warn", "strict"})
 _DEFAULT_MODE = "warn"
 _STRICT_FAIL_EXIT = 2
 _DEFAULT_TESTS_DIR = "tests"
-_HASH_ASSIGN_RE = re.compile(
-    r"^(?:PDD_)?STORY_HASH\s*=\s*(?P<value>['\"].*?['\"])",
-    re.MULTILINE,
-)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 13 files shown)


sohni-tagirisa:
## Pre-fix investigation (the "two of everything")

**Two hash primitives**
- `user_story_tests._story_content_hash` → `sha256(_normalized_story_for_hash(story))` — story prose only, strips `pdd-story-prompts`/`pdd-story-dev-units` comments + incidental whitespace. **Metadata-insensitive.**
- `story_test_generation.story_bundle_hash` → (before) `sha256(RAW story + "\n\n" + RAW contract)`. Included metadata comments → **metadata-sensitive** (Bug 2).

**Two evaluators**
- `_classify_story_from_markers` (full gate, `evaluate_stories`): accepts marker `story_hash=` kwarg OR module constant, over content ∪ bundle. Permissive/correct.
- `evaluate_story_regression` (wired to `pdd checkup coverage --story-regression-gate`): compared **only** bundle hash vs **only** `STORY_HASH`-constant regex — ignored kwargs (Bug 1).

**Two AST scanners**
- Gate `_module_string_constants`/`_string_value` resolve module constants.
- `story_regression._literal_story_id` accepted only `ast.Constant` → static fallback missed `story_id=CONST` (Bug 3).

**Committed-hash audit** — the seed tests already mixed forms: `pdd_sync/fix/change` recorded the **content** hash, while `pdd_update/connect/test/generate/checkup/bug` recorded the **old bundle** hash, and `top_flows pdd_generate` was already stale (`3660…` matched neither). Normalizing the bundle hash refreshes the bundle-form constants (content-form ones are unchanged and stay accepted after Bug 1).

## Red → green

Red (4 new tests fail on `HEAD~1`):
```
FAILED test_story_regression_gate.py::test_evaluate_story_regression_accepts_recorded_content_hash
FAILED test_story_regression_gate.py::test_evaluate_story_regression_stale_on_bogus_marker_kwarg_hash
FAILED test_story_regression_gate.py::test_metadata_only_relink_does_not_flip_bundle_hash
FAILED test_story_regression.py::TestStaticFallback::test_constant_story_id_markers_survive_empty_pytest_collection
```
Green after fix: `test_story_regression_gate.py`, `test_story_regression.py`, `test_story_test_generation.py`, `test_story_test_generator.py` → **87 passed**.

## E2E evidence
Wired evaluator on a story+contract (so bundle ≠ content) with a `pdd-story-prompts` comment:
```
A content_hash != bundle_hash: 1ab132798b7397de vs 96b4714503e40dcc
A (content-hash recorded)  -> story-regression-present   (was false-STALE)
metadata relink bundle stable: True
metadata-relink            -> story-regression-present   (no false stale)
B (bogus kwarg "0000…")    -> story-regression-stale     (was false-PASS)
```
`pdd checkup coverage --story-regression-gate warn` runs clean (the shown `story-regression-missing` entries are stories with no seed test — pre-existing gaps, unrelated). `evaluate_stories` over the repo: **0 stale**, 10 seed stories ok. `pytest -m story`: **15 passed**.

## How the evaluators are unified
Shared helper `_acceptable_story_hashes(story_path)` (content ∪ bundle) is called by BOTH `_classify_story_from_markers` and `evaluate_story_regression`; the latter now collects recorded hashes with the gate's own `_markers_in_source` AST scan (via `_recorded_story_hashes`), so kwarg/constant resolution is single-sourced. Bundle normalization lives in `story_test_generation._normalized_bundle`; the generated text-pin's inline `_bundle_hash()` imports `story_bundle_hash` so it cannot drift from the recorded constant.

`git log origin/codex/issue-1816-story-regression-epic..HEAD --oneline`:
```
1568f65aa fix(story): unify story-hash freshness semantics across gate evaluators (pdd#1889)
b95fc015f test(story): pin unified story-hash freshness semantics (pdd#1889)
```

sohni-tagirisa:
## Adversarial review — PR #1945 @ bed470cbe

**Verdict:** MERGE
**Findings:** 1 (0 P1, 1 P2)

The three fixes are sound and each is backed by a non-vacuous test (reverting each source hunk turns exactly one new test red). The full gate on the real repo classifies all 34 stories with **0 stale** (10 ok / 24 missing), and the round-trip generate→evaluate→present holds.

### P2 — `[P2] evaluate_story_regression collapses recorded hashes to found[0] — pdd/story_regression_gate.py:815 — CONFIRMED`
`_recorded_story_hashes()` returns the *list* of every hash a file records for a story, but the caller keeps only `found[0]` (`recorded[nodeid] = found[0]`). If a single test file has two markers for the same `story_id` with divergent hashes and the **stale one appears first in source order**, the wired evaluator reports `stale` even though a fresh sibling marker exists — and it mislabels every nodeid's `recorded_hashes` with the first hash. The full gate (`_classify_story_from_markers`, any-fresh-wins) is correct here, so the two evaluators still diverge in this narrow case — the exact class of bug this PR set out to kill.
Repro: one file, marker A `story_hash="deadbeef…"` then marker B `story_hash=<fresh>` for the same story → wired=`stale`; reverse the order → wired=`present`.
Not blocking: generated templates always emit identical `PDD_STORY_HASH` for both functions, and seed files use distinct story_ids per marker, so the live repo is unaffected (0 stale). **Fix:** accept if `any(h in acceptable for h in found)` across all recorded hashes (store the list, not `[0]`).

### Verified OK
- **Full gate not regressed:** `evaluate_stories` → 0 stale; classification matrix reproduced exactly as claimed — content-hash marker→ok/present, bundle-hash→ok/present, bogus-hash→stale/stale, no-hash→**full=stale but wired=present** (#1699 legacy preserved), deleted story→absent.
- **Bug 2 normalization (2a–2d), all reproduced:** (2a) prose-only edit → bundle **AND** content hash change; (2b) `pdd-story-prompts`/`dev-units` metadata-only edit → hash **unchanged**; (2c) contract-prose edit → bundle **changes**, content unchanged, contract metadata/whitespace edit → unchanged; (2d) trailing-whitespace/blank-line/trailing-newline prose edit → hash **unchanged**, and consistently so at both generation and gate time (both go through `_normalized_story_for_hash`). No masking of a real prose change observed.
- **Generation↔checking symmetry:** generated `_bundle_hash()` imports the canonical `story_bundle_hash`; `generate_story_regression_test` records `story_bundle_hash(story_path)`; round-trip generate→`evaluate_story_regression`=present and full gate=ok; recorded hash == gate-recomputed hash.
- **No missed pins:** all 10 seed pins match under the content∪bundle union (7 refreshed to bundle hashes; fix/sync/change/story_gate_identity match via the content-hash arm — untouched but valid); coverage-contracts fixture `95d7a1ca…` already refreshed in `bed470cbe`.
- **Bug 3 fallback:** static resolver handles `story_id=STORY_ID`, `story_id=PDD_STORY_ID`, and bare literals; skips unresolvable names; does not crash on a syntax-error test file. The duplicated `_module_string_constants` in `story_regression.py` is semantically identical to the gate's (both accept `Assign`/`AnnAssign`, `Name` targets, string `Constant` only) — accepted R0801 mirror, low drift risk.
- **Suites:** `test_story_regression_gate.py` + `test_story_regression.py` + `test_story_test_generation.py` = 81 passed; `pytest -m story` = 15 passed; coverage-contracts story-regression class = 7 passed. (Full `test_coverage_contracts.py` is slow >2min; ran the affected subset green.)

sohni-tagirisa:
## Implementer response to adversarial review

Verdict was MERGE (0 P1, 1 P2). Thanks — the 4 normalization cases and the full-gate/round-trip verification are exactly the invariants this PR needs to hold.

- **P2 (`found[0]` collapse → same-file stale-shadows-fresh divergence)** — FIXED in `1b7ca47c2`. Since this is the very divergence class the PR eliminates, it shouldn't ship. `evaluate_story_regression` now decides freshness over the union of ALL recorded hashes (`all_recorded & acceptable`) instead of the per-file first hash, matching the full gate's any-fresh-wins rule. The `recorded_hashes` display dict is unchanged. Added `test_evaluate_story_regression_fresh_wins_over_earlier_stale_in_same_file` (RED before the fix: a stale marker listed before a fresh one in one file reported stale; GREEN after → present). Full gate suite 40/40.

Also refreshed one pinned bundle hash in `tests/test_coverage_contracts.py` (bed470cbe) that the normalization changed — status stays `story-regression-present`, only the echoed hash value moved.

Merging into the epic branch.

PR #1947, opened 2026-07-09
fix(story): harden story gate discovery, exit safety, and CI summary (pdd#1889 P2s)

sohni-tagirisa:
## Summary

Robustness hardening for the story-regression gate, story-test generation, core-dump handling, and the CI story summary. Fixes seven confirmed P2s from **#1889** via strict TDD (red tests first, then implementation).

### Fixes (each with a red→green test)

- **G-F3 — non-UTF8 test file crashed the whole gate.** `discover_story_markers` / `discover_all_story_markers` guarded `read_text` with `except OSError` only; a `UnicodeDecodeError` (a `ValueError`) escaped and crashed the gate. Now caught alongside `OSError`, matching the sibling scanner in `pdd/story_regression.py`. The bad file is skipped like a syntax error.
- **G-F4 — class- and module-level markers were invisible to AST discovery.** `_markers_in_source` only inspected function decorators. Now also collects `@pytest.mark.story` on test **classes** and module/class-level `pytestmark = pytest.mark.story(...)` (single or list), reusing the existing kwarg/module-constant resolution. Ends the false-`missing` divergence from pytest collection.
- **G-F5 — recursive vs flat discovery diverged.** The gate + orphan check used a non-recursive `glob` while the coverage path uses `rglob`, so a nested `user_stories/<sub>/story__baz.md` was both evaluated by coverage AND warned "references a story with no story__baz.md on disk" in the same run. `discover_story_files` now uses `rglob` — one shared helper for gate, `discover_story_ids` (orphan detection), and coverage lookup. Flat default layout unaffected.
- **G-F7 — `run_story_regression_gate` ignored `PDD_USER_STORIES_DIR` when `project_root != cwd`.** It hardcoded `root/user_stories`, so a custom stories dir was silently skipped and strict mode passed vacuously over zero stories. Now resolves via `_resolve_stories_dir(None, root=root)`.
- **C-F7 — an empty `## Entry Point` silently degraded to a text-pin.** `generate_story_regression_test` routed on section truthiness; an empty Entry Point fell through to text-pinning with no warning. Now routes on heading **presence**, so an empty/partial Entry Point surfaces the existing validation error.
- **C-F8 — deliberate user errors wrote/advertised a "bug" core dump.** A `click.ClickException` (usage mistake) still wrote `.pdd/core_dumps/pdd-core-*.json` and printed "attach when reporting bugs". `handle_error` now tags deliberate ClickExceptions; `_write_core_dump` skips the snapshot when a run's only error(s) are deliberate. Genuinely unexpected faults still snapshot.
- **V-F4 — the CI Story Regression Summary miscounted.** `conftest._story_id` never looked up `story_id=` (the kwarg every story test uses), so it counted files; setup-phase skips were dropped (`@pytest.mark.skip` stories vanished); a skip-only story reported `PASS`. Now resolves `story_id=`, counts setup-phase skips, and renders skip-only as `SKIP`.

### Verification
- Red→green: all target tests fail on the pre-fix tree and pass after.
- `pytest -m story`: **15 passed**, 10 stories — and the summary now shows real story IDs (`pdd_bug`, `pdd_change`, …) instead of file paths (V-F4).
- `test_story_regression_gate.py`, `test_story_regression.py`, `test_coverage_contracts.py`, `test_story_test_generation.py`: all green.
- E2E via real paths: latin-1 file no crash; class+`pytestmark` discovered; nested story discovered with no self-contradiction; strict gate honors the env var (exit 2, not vacuous 0); empty Entry Point errors + writes no test; `pdd story link` on a missing file writes no core dump while a genuine crash still does.
- pylint on touched modules: only pre-existing warnings (9.92/10).

Refs #1889.

### DIFF
--- a/pdd/core/dump.py
+++ b/pdd/core/dump.py
@@ -178,6 +178,15 @@ def _write_core_dump(
     if not ctx.obj.get("core_dump"):
         return
 
+    # Skip the auto-snapshot when the run's only error(s) are deliberate,
+    # user-facing click errors (usage mistakes). Framing a usage mistake as a
+    # reportable "bug" snapshot — and littering .pdd/core_dumps for it — is
+    # noise; genuinely unexpected faults (no ``deliberate`` tag) still snapshot
+    # so real bugs stay debuggable (pdd#1889 C-F8).
+    recorded_errors = get_core_dump_errors()
+    if recorded_errors and all(e.get("deliberate") for e in recorded_errors):
+        return
+
     try:
         core_dump_dir = Path.cwd() / ".pdd" / "core_dumps"
         core_dump_dir.mkdir(parents=True, exist_ok=True)
--- a/pdd/core/errors.py
+++ b/pdd/core/errors.py
@@ -171,6 +171,11 @@ def handle_error(exception: Exception, command_name: str, quiet: bool):
         "traceback": "".join(
             traceback.format_exception(type(exception), exception, exception.__traceback__)
         ),
+        # A click.ClickException is a deliberate, user-facing error (a usage
+        # mistake, e.g. `pdd story link` on a missing file), not a reportable
+        # crash. Tag it so the core-dump writer can skip the auto-snapshot for
+        # it while still snapshotting genuinely unexpected faults (pdd#1889 C-F8).
+        "deliberate": isinstance(exception, click.ClickException),
     }
 
     # For KeyboardInterrupt, enrich with agentic progress context if available
--- a/pdd/story_regression_gate.py
+++ b/pdd/story_regression_gate.py
@@ -46,9 +46,9 @@ from .path_resolution import find_project_root_from_path
 from .story_regression import StoryTestMap, build_story_map
 from .story_test_generation import story_bundle_hash
 from .user_story_tests import (
-    DEFAULT_STORIES_DIR,
     STORY_PREFIX,
     STORY_SUFFIX,
+    _resolve_stories_dir,
     _story_content_hash,
     discover_story_files,
     story_id,
@@ -238,8 +238,59 @@ def _marker_from_decorator(
     )
 
 
+def _pytestmark_markers(
+    body: Sequence[ast.AST],
+    *,
+    test_name: str,
+    test_file: str,
+    constants: Dict[str, str],
+) -> List[StoryMarker]:
+    """Collect story markers declared via a ``pytestmark = pytest.mark.story(...)``
+    assignment in *body* (a module or class body).
+
+    ``pytestmark`` may be a single mark or a list/tuple of marks; a story mark in
+    either form claims every test under the enclosing scope, exactly as pytest
+    applies it at collection time.
+    """
+    found: List[StoryMarker] = []
+    for node in body:
+        if isinstance(node, ast.Assign):
+            targets: Sequence[ast.AST] = node.targets
+            value: Optional[ast.AST] = node.value
+        elif isinstance(node, ast.AnnAssign):
+            targets = [node.target]
+            value = node.value
+        else:
+            continue
+        if value is None:
+            continue
+        if not any(
+            isinstance(t, ast.Name) and t.id == "pytestmark" for t in targets
+        ):
+            continue
+        elts = value.elts if isinstance(value, (ast.List, ast.Tuple)) else [value]
+        for elt in elts:
+            marker = _marker_from_decorator(
+                elt,
+                test_name=test_name,
+                test_file=test_file,
+                lineno=getattr(elt, "lineno", getattr(node, "lineno", 1)),
+                constants=constants,
+            )
+            if marker is not None:
+                found.append(marker)
+    return found
+
+
 def _markers_in_source(source: str, test_file: str) -> List[StoryMarker]:
-    """Parse *source* and return every story marker it declares."""
+    """Parse *source* and return every story marker it declares.
+
+    Covers ``@pytest.mark.story`` decorators on a test function OR class (a class
+    marker propagates to its methods in pytest), plus module-level and
+    class-level ``pytestmark = pytest.mark.story(...)`` assignments. Missing any
+    of these would make the gate report a story ``missing`` while pytest
+    collection (and the coverage path) still see it — an evaluator divergence.
+    """
     try:
         tree = ast.parse(source)
     except SyntaxError:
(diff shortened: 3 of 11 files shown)


sohni-tagirisa:
## Pre-fix investigation

**Sibling / anti-pattern occurrences noted:**
- **G-F3:** `pdd/story_regression.py:180` already catches `(OSError, SyntaxError, UnicodeDecodeError)` in `_scan_story_markers_static` — the convention we matched. Sibling unguarded/OSError-only `read_text` calls exist for *story* files (`story_regression_gate.py:333, 361`; `coverage_contracts.py:673, 705`) but those read story markdown, not test sources, and are out of scope for this test-file crash.
- **G-F4:** `pdd/story_regression.py`'s `_scan_story_markers_static` also only walks `FunctionDef/AsyncFunctionDef` — it shares the same class/pytestmark blind spot, but this PR's scope is the gate's `_markers_in_source` (the one that drives `missing`/`stale`). Noted as a follow-up candidate.
- **G-F5:** the divergence was `discover_story_files` (non-recursive `glob`, used by the gate via `evaluate_stories` and by `discover_story_ids` for orphan detection) vs `coverage_contracts.py`'s own `rglob` at `_linked_story_ids` (671), `story_lookup` (926), and `_cross_unit_story_partners` (703). Fixing the single shared helper (`discover_story_files` → `rglob`) aligns all three consumers; the coverage-side rglobs were already recursive.

**C-F8 snapshot trace:** `handle_error` (`pdd/core/errors.py`) records the error into the module-level `_core_dump_errors` buffer but does **not** write. The snapshot is written by `_write_core_dump` (`pdd/core/dump.py`), invoked from `pdd/core/cli.py` at the end of a run (result path ~1107, exception path ~501, SystemExit path ~420). Commands like `generate` catch their own exceptions and call `handle_error` themselves (`pdd/commands/generate.py:675`), so a ClickException never reaches `CustomGroup.invoke`'s handler — it's swallowed and the dump is written by the end-of-run result path. The one common choke point across all paths is `_write_core_dump` reading `get_core_dump_errors()`. Fix: tag deliberate ClickExceptions in `handle_error` (`deliberate: isinstance(exception, click.ClickException)`) and have `_write_core_dump` skip when the run's only recorded error(s) are deliberate. `UsageError` (a ClickException subclass) is re-raised earlier and handled natively by click, so it never hit this path anyway.

## Red → Green

Pre-fix, the 9 new assertions failed (the C-F8 *control* — unexpected exception still dumps — passed pre-fix, as intended). Post-fix all pass. Per-fix test:
- G-F3: `test_discover_skips_non_utf8_file_without_crashing`, `test_evaluate_survives_non_utf8_test_file`
- G-F4: `test_discover_finds_class_level_story_marker`, `test_discover_finds_module_level_pytestmark_story`
- G-F5: `test_evaluate_discovers_nested_story` (gate), `test_nested_story_is_supported_not_self_contradicted` (coverage — evaluated AND no orphan warning)
- G-F7: `test_gate_honors_stories_dir_env_var_when_root_differs`
- C-F7: `test_empty_entry_point_errors_instead_of_silent_text_pin`
- C-F8: `test_click_exception_writes_no_core_dump` + `test_unexpected_exception_still_writes_core_dump` (control)
- V-F4: `tests/test_story_summary_conftest.py` (`_story_id` resolves `story_id=`; setup-phase skip counted; skip-only → SKIP)

## E2E (real paths)
```
G-F3 (latin-1, no crash): {'refund': 'story-regression-ok'}
G-F4 clsflow=True modflow=True
G-F5 baz discovered=True
G-F7 ok=False exit=2 n_results=1        # env-var honored, strict fails (not vacuous 0)
C-F7 raised ValueError ('Story contract must include ## Entry Poi...'); wrote_test_files=[]
C-F8 pdd story link <missing> --prompt foo.prompt -> "Error: Story file not found"; core_dumps: none
C-F8 control (RuntimeError crash) -> core_dumps: ['pdd-core-...json']   # unexpected fault still dumps
```
`pytest -m story`: **15 passed**, 10 stories; summary now keys on real story IDs (`pdd_bug`, `pdd_change`, …) — the V-F4 fix visible end-to-end.

sohni-tagirisa:
## Adversarial review — PR #1947 @ e746f758e
**Verdict:** MERGE
**Findings:** 1 (0 P1, 1 P2)

### P2
- `[P2] V-F4 hook wiring not integration-tested — tests/conftest.py:576,605 — the setup-phase-skip counting (`if _story_report_counts(report)`) and skip-only→SKIP rendering (`status = _story_summary_status(counts)`) are only exercised via direct unit tests of the pure helpers. Reverting the *wiring* in the hook (while leaving the helpers intact) fails NO test, so a future refactor could silently unwire either behavior. — Fix: add one end-to-end test that runs a `@pytest.mark.skip`-decorated story through the real hook and asserts the terminal summary renders `[SKIP]`. — CONFIRMED (reverted wiring → `test_report_counts_includes_setup_phase_skip` and `test_summary_status_skip_only_is_skip_not_pass` still pass).

### Verified OK
- **G-F3** revert → red (2 tests). BOM/binary edge: a UTF-8-with-BOM file`s marker is silently dropped via `ast.parse` SyntaxError on U+FEFF, but this is PRE-EXISTING and identical to the sibling scanner in `story_regression.py` (line 179-180, same utf-8 read + SyntaxError swallow) → no evaluator divergence; the "don`t crash" requirement is met.
- **G-F4** class marker, module `pytestmark` (single), list-form `[mark.story(...), mark.slow]`, tuple-form, and class-level `pytestmark` all resolve `story_id`/`story_hash` via the same `_marker_from_decorator` path (kwargs + module constants). A same-`story_id` class+method pair yields two markers but that is benign (dict setdefault first-wins; `discover_all` intentionally keeps all claims). Revert → red (2 tests).
- **G-F5** rglob: real-repo gate unchanged at **10 ok / 24 missing (34 total)**; `user_stories/` rglob finds the same 34 flat `story__*.md` (no `story__` files under `contracts/`; pattern still `story__*.md`). Revert → red (2 tests, gate + coverage orphan-check).
- **G-F7** honors `PDD_USER_STORIES_DIR` when root≠cwd. Revert → red.
- **C-F7** empty AND whitespace-only `## Entry Point` both raise `ValueError(... Entry Point ...)`; no-heading still falls back to text-pin. Revert → red.
- **C-F8** guard correct for all cases: only-deliberate → no dump; only-unexpected → dump; **mixed → dump** (the `all(...)` guard); legacy record with no `deliberate` key → `.get()` → no KeyError → dump; empty → dump. Both new tests assert `mock_main.assert_called_once()` (non-vacuous). The 7 pre-existing `test_core_dump`/`test_core_errors` failures are environmental (CLI ctx/flag propagation) and unchanged in nature by this hunk. Revert → red.
- **V-F4 non-story safety:** the hook is marker-gated, so non-story tests are untouched — a mixed non-story + story run exits 0 with no hook exceptions and non-story tests never appear in the summary. `story_id=` now resolves (real IDs `pdd_bug`, `pdd_change`, … in the live `-m story` summary). Revert of `_story_id` → red.
- Full suites green: `-m story` (15 passed), `test_story_regression_gate`, `test_story_regression`, `test_coverage_contracts`, `test_story_test_generation`, new `test_story_summary_conftest` — 213+ passed, only the known 7 environmental core_dump/core_errors failures remain.

MERGE — no P1 findings; the single P2 is a test-completeness gap, not a product defect.

sohni-tagirisa:
## Implementer response to adversarial review

Verdict was MERGE (0 P1, 1 P2). Thanks for the class/list/tuple `pytestmark` matrix, the C-F8 `all(...)`/legacy-record check, and confirming V-F4 is marker-gated (no effect on non-story tests' reporting or exit codes).

- **P2 (V-F4 hook wiring not directly revert-tested)** — ACCEPTED. The setup-phase-skip counting and skip-only→SKIP rendering are covered at the helper level and verified end-to-end, and the wiring is a thin marker-gated hook. Noted as a test-completeness follow-up; not holding the merge for it since the behavior is empirically correct and non-story tests are untouched.

Merging into the epic branch `codex/issue-1816-story-regression-epic`.

PR #1961, opened 2026-07-09
Bundle: run-scoped provider permanent-failure fall-through (#1936)

### LINKED ISSUE #1936: Agentic provider selection must fall through on permanent auth errors instead of failing the task (dead codex kills sync despite healthy anthropic/google)
## Problem

Inside a single pdd CLI run, the agentic phases (one-session sync, conformance-repair regeneration, agentic step tasks) select a provider, and when that provider fails with a **permanent auth error**, the task fails immediately — the CLI does not fall through to the next available provider. With codex auth dead pool-wide, every sync leg dies in 1–3 seconds at the agentic step even though healthy Anthropic/Google credentials are available in the same run.

The cloud executor's credential waterfall (promptdriven/pdd_cloud#3062) cannot compensate: earlier non-agentic phases bill real cost on platform keys, which correctly marks the run as "real work" and suppresses credential rotation; and a rotated retry would re-stage codex and repeat.

## Live evidence (2026-07-08, staging E2E, promptdriven/test_repo)

- test_repo#4233 `pdd sync` (frontend/app/contributions/page + frontend/page): identification phases billed $0.15/$0.16 fine, then
  `One-session sync Failed (0m01s, $0.0000)` / `All agent providers failed: openai: Codex CLI authentication failed: the stored ChatGPT/Codex login token could not be refreshed`.
- test_repo#4232 `pdd sync` (secrets_client): architecture-conformance repair loop engaged correctly ("retrying generation with repair directive"), then the repair retry died the same way — `Provider openai reported a permanent error (auth); skipping retries`.
- In both, "All agent providers failed" listed ONLY openai — anthropic/google were never attempted, despite valid credentials in the environment.

## Expected behavior

1. When a provider returns a **permanent** auth error (the CLI already classifies it: "permanent error (auth); skipping retries"), the agentic runner falls through to the next provider in its list (default order or `PDD_AGENTIC_PROVIDER`) instead of failing the task. "All agent providers failed" should mean all providers were actually attempted.
2. Once a provider has failed permanently in a run, later agentic calls in the same run skip it (no per-step re-burn: a 12-step workflow should not pay the dead-provider timeout 12 times).
3. Codex auto-detection (`_has_codex_auth_file` / `PDD_CODEX_AUTH_AVAILABLE`) must not pin the run to openai-only when that auth is dead — staged-auth presence is a hint, not a commitment.

## Acceptance criteria

- Regression test: agentic task with provider order [openai, anthropic]; openai returns a permanent auth error → task completes via anthropic; the failure summary records openai as permanently failed.
- One-session sync and the conformance-repair retry path both inherit the fall-through (the two live failure shapes above).
- A run where ALL providers genuinely fail still reports "All agent providers failed:" with per-provider reasons.
- No behavior change when the first provider works.

## Context

Found by the staging E2E hardening campaign (chains: change→sync→checkup). Sibling cloud-layer fix: promptdriven/pdd_cloud#3062 (waterfall rotation on the generic permanent-auth stdout line). Short-term staging mitigation available (`PDD_AGENTIC_PROVIDER=anthropic,google`), but the fall-through is the universal fix.

Serhan-Asad:
## Bundle: run-scoped provider permanent-failure fall-through

Off current `main`, adds run-scoped fall-through so a provider family that has reported a **permanent** failure once during a run is skipped for the rest of that run instead of being retried into the same wall.

### Design

A per-run registry in `agentic_common` records provider families that have hit a permanent (non-transient) failure — auth/credential-limit and other terminal classifications — and short-circuits further dispatch to that family for the remainder of the run, falling through to the next eligible provider. The registry is run-scoped (not process-global), so it never leaks across independent runs, and transient failures are deliberately excluded so normal retry/backoff is unaffected.

### Root-cause correction

The initial framing treated this as a routing/registry gap; the corrected root cause — that the permanent-failure signal was observed but not made authoritative for subsequent dispatch within the run — is documented at:
https://github.com/promptdriven/pdd/issues/1936#issuecomment-4927918151

Closes #1936

### Validation

`tests/test_agentic_common.py` — 525 passed on the bundle branch.


Serhan-Asad:
## Staging E2E evidence (2026-07-09, staging1, stack/e2e-staging-v2 = 288d05316f)

Validated on every lane, all day, against the genuinely-dead codex fleet:

- First credential attempts (codex) failed in **seconds for ~$0**: steps after the first permanent auth error fail-fast with the new registry message — `All agent providers failed permanently earlier this run: openai: auth` — instead of re-burning the auth timeout on each of 12–13 workflow steps (observed on test_repo#4248/#4249/#4250/#4251).
- The registry's env mirror propagated through re-entrant steps (no step re-attempted the dead provider within an attempt).
- Root-cause correction documented on the issue: https://github.com/promptdriven/pdd/issues/1936#issuecomment-4927918151 (filter was never bypassed; the gaps were cross-call memory + the staged-auth pin — both covered here).

Serhan-Asad:
Audit 2026-07-21: closing this PR as superseded by #1998, the canonical PDD staging campaign bundle containing or replacing this slice. The branch and review history remain recoverable here; reopen only if #1998 drops this behavior.

PR #1975, opened 2026-07-10
Prevent include parsing across inline documentation tokens

### LINKED ISSUE #1974: Prevent include parser from spanning inline documentation tokens
pdd update can fail with Errno 36 when extract_includes_from_file pairs an inline documentation <include> token with a much later real closing tag. The DOTALL body match captures hundreds of lines; agentic_update then passes that capture to Path(). Require body-form matching not to cross another <include opener while preserving bare, attributed, self-closing, and include-many forms. Reproduced on pdd/prompts/agentic_change_orchestrator_python.prompt while resolving promptdriven/pdd#1620.

RyanTanuki:
## Summary

Fixes #1974 by preventing the body-form include regex from crossing a nested <include opener.

## Root cause

An inline documentation <include> token could pair with a much later genuine closing tag. The parser returned hundreds of lines as one include path, and agentic_update._compute_include_allowlist raised OSError: [Errno 36] File name too long when constructing Path(...).

## Changes

- Temper the body-form regex so a match cannot cross another include opener.
- Preserve bare, attributed-body, self-closing, and include-many parsing.
- Add the exact inline-token-before-real-include regression.
- Update the owning prompt contract and run deterministic PDD metadata sync.

## Validation

Per the requested workflow, no local test command was run; PR CI is the validation gate.


RyanTanuki:
Added commit `115dbaedc` to record the PDD-generated `sync_order` fingerprint.

Root cause of the first auto-heal failure:
- The parser prompt and generated code were intentionally changed together.
- Deterministic metadata sync generated `.pdd/meta/sync_order_python.json`, but the repository-wide metadata ignore rule left this previously untracked module fingerprint out of the PR.
- Without a tracked current fingerprint, auto-heal correctly reported an unresolved prompt/code conflict.

The fingerprint is now explicitly tracked as the reconciliation record. The other reported drift, `agentic_change_orchestrator`, pre-exists this branch and is repaired by PR #1620; this branch will merge updated `main` after #1620 lands.

PR #1981, opened 2026-07-10
Epic: agentic-sync default-readiness fixes (#1979 exit codes, #1980 scope reconciliation)

sohni-tagirisa:
## Epic: agentic-sync default-readiness fixes

Integration branch for the two remaining correctness blockers identified by the agentic-sync E2E validation (epic #1868) before the agentic-vs-regular default sync evaluation can be trusted:

| Sub-PR | Issue | Fix |
|---|---|---|
| #1982 (merged) | #1979 | `pdd sync <basename>` exits non-zero when `Overall status: Failed` |
| #1983 (in review) | #1980 | Branch-diff fast path reconciles diff-detected scope with the issue's explicitly requested modules |

## Why these two

Both bugs make failure look like success, which poisons any exit-code- or status-based comparison of regular vs agentic sync:

- **#1979** — CI gates, shell chains, and benchmarks read a failed single-module sync as success. The agentic runner already carries a stdout-scraping workaround (`agentic_sync_runner.py` sticky capture); direct invocations have nothing.
- **#1980** — an issue-driven agentic sync can silently drop explicitly requested modules and still report Success. The GitHub App uses exactly this issue-URL route.

## Process

Each fix ships as a TDD sub-PR (failing test committed and pushed before the fix) into this branch, with an adversarial Codex review loop per `docs/runbooks/pr-loop-process.md`.

**Do not merge this epic into `main` without maintainer review.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1883, opened 2026-07-07
pdd sync: consume PDD_CHANGED_MODULES as authoritative module-identification fast path


DianaTao:
## Summary

`pdd_cloud` already computes a deterministic changed-module list for cloud `pdd sync` jobs and exports it as `PDD_CHANGED_MODULES`, but the PDD CLI does not consume that environment variable directly. Today the value is only appended to `PDD_USER_FEEDBACK` as an advisory LLM hint.

That leaves cloud sync vulnerable to the same module-identification failure modes even when cloud has already computed the correct module list: oversized identify-module prompts, stale feedback text, or LLM/tooling failures can still decide that there are no modules or choose the wrong module.

PDD should treat `PDD_CHANGED_MODULES` as an authoritative module-identification fast path before falling back to the LLM identify-modules step.

## Current behavior

Cloud side:

- `pdd_cloud` extracts changed prompt modules from the checked-out branch diff in `extensions/github_pdd_app/src/workers/pdd_executor/env_setup.py`.
- It preserves directory context under `prompts/` so nested modules remain distinguishable.
- It exports a comma-separated list, for example:

```text
PDD_CHANGED_MODULES=src/services/foo,frontend/app/dashboard/page
```

- It also appends a textual hint to `PDD_USER_FEEDBACK`:

```text
IMPORTANT: The following modules were changed on the branch and MUST be synced: ...
```

PDD side:

- The PDD CLI has no direct reader for `PDD_CHANGED_MODULES`.
- Agentic sync still relies on PDD's own git diff path or `agentic_sync_identify_modules_LLM`.
- Therefore the deterministic cloud module list is advisory only. If PDD's own diff path misses the module, the LLM prompt overflows, or stale feedback conflicts with the current diff, the sync can still fail or sync the wrong target.

### 1. Input-too-large module identification collapsed into "no modules"

Related:

- promptdriven/pdd_cloud#2196
- promptdriven/pdd_cloud#2234
- promptdriven/pdd_cloud#2257
- promptdriven/pdd#1838
- promptdriven/pdd#1679

Observed cloud failure:

```text
input_error_code: input_too_large
max_chars: 1048576
actual_chars: 1096015
```

The same run later surfaced as:

```text
LLM identified no modules to sync
```

The PDD-side prompt-size guard has since been implemented in promptdriven/pdd#1679, but this incident shows why the identify-modules LLM should not be required when cloud has already provided a deterministic changed-module list.

[...]

PR #1868, opened 2026-07-06
Epic: agentic sync E2E validation & force-local reliability

sohni-tagirisa:
Tracking epic for non-billable end-to-end validation of **agentic issue sync** (`pdd sync <issue-url>`) and the reliability fixes that validation surfaced. Related: #1464 (agentic-by-default), pdd_cloud epic PR promptdriven/pdd_cloud#2722 (app-side agentic sync integration).

## Why

The GitHub app already shells out to `pdd sync <issue-url>` (the agentic route) as its sync mechanism, and pdd_cloud#2716 proposes making agentic sync the app's first-class default. Before that migration, the CLI flow needed an end-to-end reliability assessment that does **not** consume provider credits: no test previously exercised the real chain *CLI → `run_agentic_sync` → `AsyncSyncRunner` → child `pdd sync` subprocesses* (unit suites patch `run_agentic_task`/`AsyncSyncRunner` in-process; the deepest CLI test, `test_e2e_issue_557_codex_ndjson.py`, stops at the runner boundary).

## Validation performed (hermetic, zero billable calls)

Sandbox: `env -i` (no inherited keys), scratch `HOME` (hides `~/.codex/auth.json`, `~/.gemini` OAuth, litellm key files), restricted `PATH` with shims for `claude`/`gh`/`pdd` plus tripwire shims for `codex`/`gemini`/`agy`/`opencode` that fail loudly if invoked, `PDD_FORCE_LOCAL=1`, `PDD_AGENTIC_PROVIDER=anthropic`, `PDD_FORCE=1`, `PDD_AUTO_UPDATE=false`.

Verified end to end against `origin/main`: dry-run pipeline (issue fetch → LLM module identification → dependency graph → real child dry-run validation → fingerprint filter → report), write-mode failure safety (failed module blocks dependents, resumable state file, clean tree, exit 1), identification-failure error comment, deterministic branch-diff detection (zero provider calls), resume-skip + state cleanup, and budget forwarding to children.

## Defects found

1. **`PDD_FORCE_LOCAL=1` (env-only) did not stop per-step cloud dispatch** — `ctx.obj["local"]` came only from the `--local` argv flag; outside `CI=1` the generate step launched a real interactive GitHub device-flow prompt and hung the sync child indefinitely. *(fixed in this epic)*
2. **`--local` was dropped for issue-sync children** — `run_agentic_sync` never forwarded it (unlike `run_global_sync`). *(fixed in this epic)*
[...]

PR #1983, opened 2026-07-10
fix(sync): reconcile branch-diff fast-path scope with issue's explicitly requested modules (#1980)

### LINKED ISSUE #1980: agentic sync: branch-diff fast path silently under-scopes vs issue's explicitly requested modules
## Summary

The agentic issue-sync branch-diff fast path silently **under-scopes** the sync relative to the modules the issue explicitly requests. When the current work branch has a diff, `_detect_modules_from_branch_diff()` wins unconditionally and the issue text is never consulted.

Reproduced 2026-07-07 in the non-billable E2E harness (epic PR #1868 validation): the fixture issue explicitly asked to sync `greeter` **and** `textutil`; the work branch's diff only touched `greeter`. The fast path synced only `greeter`, dropped the `greeter → textutil` dependency edge with just a yellow warning, and reported **Success**. The LLM-identify path (taken when the work branch equals main) scopes both modules correctly from the same issue.

## Where

- `pdd/agentic_sync.py:152` — `_detect_modules_from_branch_diff()`.
- `pdd/agentic_sync.py:~2736` — step 7 of the issue-sync flow: `branch_modules = _detect_modules_from_branch_diff(project_root)`; when non-empty, `modules_to_sync = branch_modules` with no cross-check against the issue content. Note the issue `title`/`body` (and the `architecture` module inventory) are already in scope at this point.

## Why it matters

- An issue-driven sync that reports Success while ignoring explicitly requested modules is silent corruption of the operation's contract — the caller (including the GitHub App, which uses exactly this issue-URL route) believes the issue is fully handled.
- This is one of the two remaining blockers for the agentic-sync default GO/NO-GO evaluation.

## Suggested fix shape

Keep the fast path deterministic and free (0 LLM calls), but reconcile before committing to it:

1. Deterministically extract explicitly-named modules from the issue title/body by matching against the known module inventory (architecture.json entries / discovered prompt basenames).
2. If the issue explicitly names modules that are missing from the diff-detected set, either (a) union them into `modules_to_sync` (still deterministic), or (b) fall through to the LLM identification step for authoritative scoping.
3. Never report Success on a scope that silently dropped explicitly requested modules — at minimum the mismatch must be a hard decision point, not a yellow warning.

## Acceptance criteria

- [ ] Given an issue explicitly naming modules A and B and a work-branch diff touching only A, the sync scope includes both A and B (or the run clearly fails/asks, never silent Success on A only).
- [ ] Fast path still avoids LLM calls when the diff-detected scope is consistent with the issue text.
- [ ] Dependency edges between requested modules are preserved.
- [ ] Regression test covering the under-scope reproduction; existing branch-diff fast-path tests keep passing.

Refs: epic PR #1868 (validation that found this), #1883 (PDD_CHANGED_MODULES fast path), #1396 (runtime-LLM-only boundary).

sohni-tagirisa:
Fixes #1980. Part of epic PR #1981.

## Root cause

Step 7 of the issue-sync flow (`run_agentic_sync`, `pdd/agentic_sync.py`) committed to `modules_to_sync = branch_modules` unconditionally whenever `_detect_modules_from_branch_diff()` returned a non-empty set. The issue `title`/`body` and the loaded `architecture` inventory were already in scope at that point but never consulted, so the deterministic fast path silently under-scoped relative to what the issue explicitly requested. Repro (found by the #1868 E2E validation): issue explicitly asks to sync `greeter` **and** `textutil`; work-branch diff touches only greeter → only greeter synced, the greeter→textutil dependency edge dropped with a yellow warning, Success reported. The LLM-identify path scoped both correctly from the same issue.

## Chosen semantics: deterministic union (still 0 LLM calls)

Before committing to the fast path, deterministically extract explicitly-named known modules from the issue title/body (`_extract_issue_named_modules`) and **union** any that the diff missed into the scope, with a clear log of which modules came from the diff vs the issue. Union was chosen over falling through to LLM identification because:

- it preserves the fast path's core contract (deterministic, free, reproducible) — a mismatch is exactly the case where the issue's explicit request is unambiguous, so no LLM judgment is needed;
- falling through would spend an LLM call to rediscover information we already hold deterministically, and would make scope non-reproducible between runs;
- issue-only modules go through the same downstream pipeline as every other scope source (`_normalize_modules_for_sync`, `_filter_invalid_basenames`, `_drop_ambiguous_basenames` (#1677), dep-graph expansion), so nothing bypasses existing validation.

When the diff-detected scope already covers every issue-named module, behavior is byte-for-byte unchanged: same scope, same output, no LLM calls.

## Precision/recall trade-off in extraction

Only tokens that unambiguously resolve against the known architecture inventory count as "explicitly named" — precision over recall, because a false positive here inflates a write-mode sync scope:

1. **Backticked inline-code tokens** (`` `greeter` ``, `` `prompts/greeter_python.prompt` ``) resolving to a known basename after `.prompt`/language-suffix stripping.
2. **Prompt-file path tokens** anywhere in the text (the FILES_MODIFIED-style lists the GitHub App writes) — this is the exact shape of the #1868 repro fixture.
3. **Bare word-boundary mentions**, but *only* for basenames whose final component contains an underscore (e.g. `ci_validation`): such names cannot be ordinary prose words. A single-word module named e.g. `python` is never pulled in by prose ("written in python"); requesting it requires backticks or its prompt path. Case-sensitive; ambiguous tails (multiple architecture entries) are skipped, consistent with #1677.

Accepted recall loss (documented on purpose): an issue that names a single-word module in plain prose without backticks or a prompt path does not trigger the union — behavior is then no worse than today. Extraction matches only title/body (not comments), mirroring where the issue's *request* lives; comments routinely quote logs naming unrelated modules.

## Hard boundaries preserved

- `_branch_diff_is_runtime_llm_only` (#1396): untouched arm; runtime `*_LLM` templates never resolve during extraction (both the inventory and `extract_module_from_include` exclude them), covered by `test_runtime_llm_templates_never_extracted`.
- `PDD_CHANGED_MODULES` (#1883): arm untouched — see sibling note in comments.

## Test plan (TDD: red pushed first, commit 854ae1369)

- `tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation` — 7 helper unit tests (backticks, prompt paths, prose guard for a `python`-named module, underscored-name prose match, longer-token non-match, `*_LLM` exclusion, no-architecture no-op) + 3 `run_agentic_sync` integration tests (union repro asserting both modules reach dry-run and the runner with `run_agentic_task` never called; superset-unchanged; prose-non-inflation).
- `tests/test_agentic_sync_mocked_e2e.py::test_branch_diff_scope_reconciled_with_issue_named_modules` — hermetic real-CLI repro: work branch touches only greeter, fixture issue names both prompts → both child dry-run validations run, zero provider calls.
- Full `tests/test_agentic_sync.py` (247 passed) and full mocked E2E suite (5 passed); `pylint pdd/agentic_sync.py` introduces no new findings vs the epic baseline (verified by diffing pylint output; the resume E2E test's state fixture now marks both fixture modules succeeded because the fixture issue explicitly names both prompts).

### DIFF
--- a/pdd/agentic_sync.py
+++ b/pdd/agentic_sync.py
@@ -21,11 +21,13 @@ from typing import Any, Dict, List, NamedTuple, Optional, Tuple
 from rich.console import Console
 
 from .agentic_change import _check_gh_cli, _escape_format_braces, _parse_issue_url, _run_gh_command
+from .agentic_change_orchestrator import _extract_marker_paths
 from .agentic_common import build_agentic_task_instruction, run_agentic_task
 from .agentic_sync_runner import (
     AsyncSyncRunner,
     _architecture_entry_aliases,
     _basename_from_architecture_filename,
+    _basename_from_architecture_filepath,
     _find_pdd_executable,
     build_dep_graph_from_architecture_data,
 )
@@ -230,8 +232,107 @@ def _detect_modules_from_branch_diff(project_root: Path) -> List[str]:
 
 # Backticked inline-code spans (single-line, non-empty) in issue markdown.
 _ISSUE_BACKTICK_TOKEN_RE = re.compile(r"`([^`\n]+)`")
-# Prompt-file path tokens (FILES_MODIFIED lists, prose references).
-_ISSUE_PROMPT_PATH_RE = re.compile(r"[A-Za-z0-9_.\-][A-Za-z0-9_./\-]*\.prompt\b")
+# Prompt-file path tokens (FILES_MODIFIED lists, prose references). Captures
+# full non-whitespace paths so supported route-segment characters like the
+# brackets in ``app/users/[id]/page_TypeScriptReact.prompt`` survive
+# (PR #1983 round 3, P1#2); surrounding markdown punctuation is trimmed
+# afterwards by ``_prompt_path_tokens``.
+_ISSUE_PROMPT_PATH_RE = re.compile(r"\S+\.prompt\b")
+# Leading punctuation markdown wraps around a path (backticks, quotes, parens,
+# emphasis, list/link leftovers). The ``\.prompt`` anchor already excludes
+# trailing punctuation. ``[`` is handled separately: it is markdown link syntax
+# only when no matching ``]`` follows (otherwise it starts a dynamic route
+# segment such as ``[id]/page_...``).
+_PROMPT_PATH_LEAD_TRIM = "`\"'(<*,;:!]}>"
+
+
+def _prompt_path_tokens(text: str) -> List[str]:
+    """Extract prompt-file path tokens from text, trimming markdown wrapping."""
+    tokens: List[str] = []
+    for raw in _ISSUE_PROMPT_PATH_RE.findall(text):
+        token = raw.lstrip(_PROMPT_PATH_LEAD_TRIM)
+        while token.startswith("[") and "]" not in token:
+            token = token[1:]
+        if token:
+            tokens.append(token)
+    return tokens
+
+
+def _strip_fenced_code_blocks(text: str) -> str:
+    """Remove fenced code blocks (``` or ~~~) from markdown text.
+
+    Fenced content is quoted logs/snippets, not a request (PR #1983 round 3,
+    P2). An unclosed fence conservatively drops the remainder of the text —
+    under-matching noise is safe, while over-matching inflates a write-mode
+    sync scope.
+    """
+    kept: List[str] = []
+    fence_marker: Optional[str] = None
+    for line in text.splitlines():
+        stripped = line.lstrip()
+        if fence_marker is None and (
+            stripped.startswith("```") or stripped.startswith("~~~")
+        ):
+            fence_marker = stripped[:3]
+            continue
+        if fence_marker is not None and stripped.startswith(fence_marker):
+            fence_marker = None
+            continue
+        if fence_marker is None:
+            kept.append(line)
+    return "\n".join(kept)
+
+
+def _architecture_identity_index(
+    architecture: Optional[List[Dict[str, Any]]],
+) -> Tuple[List[str], set, Dict[str, List[str]]]:
+    """Index REAL per-entry module identities, duplicates preserved.
+
+    ``_architecture_module_basenames`` dedupes by derived basename, which
+    collapses distinct modules that share a prompt FILENAME — the normal
+    Next.js shape: several ``page_TypeScriptReact.prompt`` entries differing
+    only by route dir / ``filepath``. Extraction and coverage must instead see
+    one identity per distinct entry so a shared tail is detectably ambiguous
+    (PR #1983 round 3, P1#1).
+
+    Returns ``(identities, known_keys, tail_to_identities)``:
+
+    * ``identities`` — one primary identity per distinct entry (filename
+      derivation with directories preserved, falling back to the
+      filepath-derived key); duplicates KEPT.
+    * ``known_keys`` — every exact-match alias (filename- and
+      filepath-derived, via ``_architecture_entry_aliases``).
+    * ``tail_to_identities`` — final path component -> per-entry identities;
+      a tail listing more than one entry is ambiguous.
+
+    Runtime ``*_LLM.prompt`` entries are excluded entirely. True duplicate
+    entries (same primary identity AND same filepath) are counted once.
+    """
+    identities: List[str] = []
+    known_keys: set = set()
+    tail_to_identities: Dict[str, List[str]] = {}
+    seen_entries: set = set()
+    for entry in architecture or []:
+        if not isinstance(entry, dict):
+            continue
+        filename = str(entry.get("filename", "") or "")
+        if _is_runtime_llm_template(filename):
+            continue
+        filepath = entry.get("filepath")
+        filepath_str = filepath if isinstance(filepath, str) else ""
+        primary = _basename_from_architecture_filename(filename)
+        if not primary and filepath_str:
+            primary = _basename_from_architecture_filepath(filepath_str)
+        if not primary:
+            continue
+        entry_key = (primary, filepath_str)
+        if entry_key in seen_entries:
+            continue
+        seen_entries.add(entry_key)
+        identities.append(primary)
+        known_keys.update(_architecture_entry_aliases(entry))
+        tail_to_identities.setdefault(primary.rsplit("/", 1)[-1], []).append(primary)
+    return identities, known_keys, tail_to_identities
 
 
 def _module_key_from_prompt_path(path_token: str) -> Optional[str]:
@@ -271,40 +372,26 @@ def _module_key_from_prompt_path(path_token: str) -> Optional[str]:
     return f"{prefix}{basename}"
 
 
-def _issue_candidate_tokens(text: str) -> List[str]:
-    """Collect high-precision explicit-module tokens from issue text.
-
-    Two token classes (see ``_extract_issue_named_modules``):
-    backticked single-word inline-code spans, and prompt-file path tokens
-    (FILES_MODIFIED-style lists). Prompt paths convert to path-qualified
-    module keys via ``_module_key_from_prompt_path``; runtime ``*_LLM.prompt``
-    templates are rejected there and again in the downstream resolver.
-    """
-    tokens: List[str] = []
-    for raw in _ISSUE_BACKTICK_TOKEN_RE.findall(text):
-        token = raw.strip()
-        if token and not any(ch.isspace() for ch in token):
-            tokens.append(token)
-    for path_token in _ISSUE_PROMPT_PATH_RE.findall(text):
-        module_key = _module_key_from_prompt_path(path_token)
-        if module_key:
-            tokens.append(module_key)
-    return tokens
-
-
 def _resolve_issue_module_token(
     token: str,
-    known_set: set,
-    tail_to_modules: Dict[str, List[str]],
+    known_keys: set,
+    tail_to_identities: Dict[str, List[str]],
 ) -> Optional[str]:
     """Resolve one candidate token to a known module key, or ``None``.
 
     Tries the token verbatim, then with ``.prompt`` / language suffixes
-    stripped. A candidate that is not an exact known key may still resolve
-    through its final path component when that tail is UNAMBIGUOUS (exactly
-    one known module) and the qualification is consistent (one side is a
-    path suffix of the other, or the candidate is bare). Runtime ``*_LLM``
-    templates never resolve (#1396 boundary).
+    stripped. Rules over the REAL identity index (duplicates preserved):
+
+    * An exact alias match resolves — but a BARE candidate whose tail lists
+      more than one distinct entry has no deterministic referent (two Next.js
+      ``page`` modules answer to the same bare name) and is skipped; a
+      path-qualified alias stays exact.
+    * A non-exact candidate resolves through its final path component only
+      when that tail lists exactly ONE entry and the qualification is
+      consistent (one side is a path suffix of the other, or the candidate is
+      bare).
+
+    Runtime ``*_LLM`` templates never resolve (#1396 boundary).
     """
     if _is_runtime_llm_template(token):
         return None
@@ -315,9 +402,11 @@ def _resolve_issue_module_token(
     if stripped:
         candidates.append(stripped)
     for cand in candidates:
-        if cand in known_set:
-            return cand
-        tail_matches = tail_to_modules.get(cand.rsplit("/", 1)[-1], [])
+        tail_matches = tail_to_identities.get(cand.rsplit("/", 1)[-1], [])
+        if cand in known_keys:
+            if "/" in cand or len(tail_matches) == 1:
+                return cand
+            continue
         if len(tail_matches) == 1 and (
             "/" not in cand
             or tail_matches[0].endswith(f"/{cand}")
@@ -327,21 +416,96 @@ def _resolve_issue_module_token(
     return None
 
 
-def _issue_scan_text(
-    title: str,
-    body: str,
-    comments: Optional[List[Dict[str, Any]]],
-) -> str:
-    """Join the issue title, body, and comment BODIES for module scanning.
+def _bare_mention_identities(
+    prose_text: str,
+    identities: List[str],
+    tail_to_identities: Dict[str, List[str]],
+) -> List[str]:
+    """Class-3 scan: bare word-boundary mentions of identifier-like names.
 
-    Only comment bodies are included (not author logins or structural
-    markers), so a username can never resolve to a module.
+    Only tails containing an underscore or hyphen qualify (they cannot be
+    ordinary prose words) and only when the tail lists exactly one real entry.
+    The lookarounds exclude adjacent word chars, ``/``, ``.``, and ``-`` so a
+    standalone mention matches but longer identifiers (``ci_validation_extra``,
+    ``double-check-run``), path segments (``a/ci_validation/b``), and filenames
+    (``ci_validation.py`` — file mentions go through class 2) do not.
     """
-    parts = [title or "", body or ""]
+    found: List[str] = []
+    for module in identities:
+        tail = module.rsplit("/", 1)[-1]
+        if module in found or len(tail_to_identities[tail]) != 1:
+            continue
+        if "_" not in tail and "-" not in tail:
+            continue
+        if re.search(rf"(?<![\w./\-]){re.escape(tail)}(?![\w./\-])", prose_text):
+            found.append(module)
+    return found
+
+
+def _trusted_comment_prompt_paths(
+    comments: Optional[List[Dict[str, Any]]],
+) -> List[str]:
+    """Collect prompt paths from trusted FILES_MODIFIED marker payloads only.
+
+    Issue comments are largely workflow noise (step logs, retries), so free
+    text in them must never widen a write-mode sync scope (PR #1983 round 3,
+    P2). The only trusted signal is a ``FILES_MODIFIED:`` marker payload — the
+    exact format the change orchestrator emits — parsed with the same
+    ``_extract_marker_paths`` walker that produces/consumes those markers.
+    Fenced code blocks are stripped FIRST so a marker quoted inside a log
+    fence is treated as a log, not a directive. Only ``*.prompt`` payload
+    entries are returned; code/doc paths are not module requests.
+    """
+    paths: List[str] = []
     for comment in comments or []:
-        if isinstance(comment, dict):
-            parts.append(str(comment.get("body", "") or ""))
-    return "\n".join(parts)
+        if not isinstance(comment, dict):
+            continue
+        comment_body = str(comment.get("body", "") or "")
+        payload = _extract_marker_paths(
+            "FILES_MODIFIED", _strip_fenced_code_blocks(comment_body)
+        )
+        paths.extend(p for p in payload if p.endswith(".prompt"))
+    return paths
+
+
+def _explicit_token_modules(
+    text: str,
+    comments: Optional[List[Dict[str, Any]]],
+    known_keys: set,
+    tail_to_identities: Dict[str, List[str]],
+) -> List[str]:
+    """Resolve class-1 and class-2 tokens to module keys, in mention order.
+
+    Class 1: backticked single-word tokens from the title/body, resolved via
+    ``_resolve_issue_module_token``. Class 2: prompt-file paths from the
+    title/body plus trusted FILES_MODIFIED marker payloads from the filtered
+    comments. A path-qualified derived key whose tail exists in the inventory
+    is accepted directly: the path itself is the disambiguation, and the key
+    form matches what ``_detect_modules_from_branch_diff`` schedules for the
+    same file (#1675). Bare derived keys must still resolve uniquely.
+    """
+    modules: List[str] = []
+
+    def _push(module: Optional[str]) -> None:
+        if module and module not in modules:
+            modules.append(module)
+
+    for raw in _ISSUE_BACKTICK_TOKEN_RE.findall(text):
+        token = raw.strip()
+        if token and not any(ch.isspace() for ch in token):
+            _push(_resolve_issue_module_token(token, known_keys, tail_to_identities))
+
+    for path_token in _prompt_path_tokens(text) + _trusted_comment_prompt_paths(comments):
+        module_key = _module_key_from_prompt_path(path_token)
+        if not module_key:
+            continue
+        if "/" in module_key and module_key.rsplit("/", 1)[-1] in tail_to_identities:
+            _push(module_key)
+        else:
+            _push(
+                _resolve_issue_module_token(module_key, known_keys, tail_to_identities)
+            )
+    return modules
 
 
 def _extract_issue_named_modules(
@@ -354,11 +518,10 @@ def _extract_issue_named_modules(
 
     Issue #1980: the branch-diff fast path must not silently drop modules the
     issue explicitly requests. This helper matches the issue title/body — plus
-    the FILTERED comment bodies (the same signal-bearing set the identify path
-    consumes; FILES_MODIFIED markers arrive as comments, PR #1983 review P1a)
-    — against the KNOWN module inventory (architecture basenames, runtime
-    ``*_LLM`` templates already excluded by ``_architecture_module_basenames``)
-    using three high-precision, zero-LLM token classes:
+    trusted marker payloads from the FILTERED comments (FILES_MODIFIED markers
+    arrive as comments, PR #1983 review P1a) — against the REAL module
+    identity index (runtime ``*_LLM`` templates excluded) using three
+    high-precision, zero-LLM token classes:
 
     1. Backticked inline-code tokens (e.g. ```greeter```,
        ```prompts/greeter_python.prompt```) that resolve to a known
@@ -374,48 +537,32 @@ def _extract_issue_named_modules(
        plain-prose mention of a single-word module is deliberately NOT
        treated as an explicit request; use backticks or the prompt path).
 
-    Ambiguous bare tails (mapping to more than one known module) are skipped;
-    downstream #1677 handling requires path-qualified names for those anyway.
-    Returns known-module keys in first-mention order; empty when there is no
+    Scan surfaces (PR #1983 round 3, P2): classes 1 and 2 scan the title/body;
+    class 3 scans the title/body with fenced code blocks stripped (quoted
+    logs/snippets are not requests). Comments contribute ONLY trusted
+    ``FILES_MODIFIED:`` marker payloads (class 2), never free text.
+
+    Ambiguous tails (listing more than one real entry) never resolve from
+    bare names and a path-qualified prompt-path key is accepted as a
+    scheduler key directly — the same key form the branch-diff detector emits
+    (#1675). Returns keys in first-mention order; empty when there is no
     architecture inventory to match against.
     """
-    known = _architecture_module_basenames(architecture or [])
-    if not known:
+    identities, known_keys, tail_to_identities = _architecture_identity_index(
+        architecture
+    )
+    if not identities:
         return []
-    known_set = set(known)
-    # Map each final path component to its full module keys so a bare mention
-    # of "worker_app" can resolve "extensions/foo/src/worker_app".
-    tail_to_modules: Dict[str, List[str]] = {}
-    for module in known:
-        tail_to_modules.setdefault(module.rsplit("/", 1)[-1], []).append(module)
 
-    text = _issue_scan_text(title, body, comments)
+    text = f"{title or ''}\n{body or ''}"
     named: List[str] = []
-    seen: set[str] = set()
-
-    # Classes 1 + 2: backticked tokens and prompt-file paths.
-    for token in _issue_candidate_tokens(text):
-        resolved = _resolve_issue_module_token(token, known_set, tail_to_modules)
-        if resolved and resolved not in seen:
-            named.append(resolved)
-            seen.add(resolved)
-
-    # Class 3: bare word-boundary mentions of identifier-like names.
-    for module in known:
-        tail = module.rsplit("/", 1)[-1]
-        if module in seen or len(tail_to_modules[tail]) != 1:
-            continue
-        if "_" not in tail and "-" not in tail:
-            continue
-        # No word chars, '/', '.', or '-' adjacent: matches a standalone
-        # mention but not longer identifiers (ci_validation_extra,
-        # double-check-run), path segments (a/ci_validation/b), or filenames
-        # (ci_validation.py is intentionally excluded — file mentions go
-        # through class 2).
-        if re.search(rf"(?<![\w./\-]){re.escape(tail)}(?![\w./\-])", text):
+    for module in _explicit_token_modules(
+        text, comments, known_keys, tail_to_identities
+    ) + _bare_mention_identities(
+        _strip_fenced_code_blocks(text), identities, tail_to_identities
+    ):
+        if module not in named:
             named.append(module)
-            seen.add(module)
-
     return named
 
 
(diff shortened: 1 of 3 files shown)


sohni-tagirisa:
## Pre-fix investigation summary

**Bug mechanics (verified on the epic base = current main):**
- `_detect_modules_from_branch_diff()` (`pdd/agentic_sync.py`) diffs `main...HEAD` for changed non-`_LLM` `.prompt` files and returns path-qualified module keys.
- Step 7 of `run_agentic_sync`: `if branch_modules: modules_to_sync = branch_modules` — unconditional. `title`, `body`, filtered comments and the loaded `architecture` inventory are all in scope at that line but never consulted. Downstream, dep-graph expansion only sees the single diff module, so the greeter→textutil edge is dropped with a yellow "not in scope" style warning and the run reports Success.

**Inventory / helper reuse:** the flow already loads `architecture` via `_load_architecture_json`; `_architecture_module_basenames()` yields the known syncable basenames (runtime `*_LLM` entries excluded by `_basename_from_architecture_filename`). Suffix normalization reuses `_strip_language_suffix_preserving_path` / `extract_module_from_include` — no new normalization logic was invented. No pre-existing issue-text→module matcher existed (the only issue-text consumer was the LLM identify prompt), so `_extract_issue_named_modules` is new, placed next to the diff detector.

**Prior art mined:**
- #1883 (`325d70f93`): `PDD_CHANGED_MODULES` fast path — established the pattern of a deterministic scope source feeding the shared normalization/validation pipeline, with hard failure on unresolved targets.
- #1396 (runtime-LLM-only boundary): example of a carefully-bounded deterministic decision in the same step; its arm and semantics are untouched here, and extraction can never resolve a `*_LLM` template (regression-tested).
- #1677 ambiguity handling and #1675 path-qualified keys informed the coverage check (`_issue_modules_missing_from_scope` compares full keys and tails) and the decision to skip ambiguous bare tails during extraction.

**Sibling hunt — `PDD_CHANGED_MODULES` arm (#1883):** it can in principle also under-scope vs the issue text, but it is a different contract, not the same bug shape: the env var is an explicit, authoritative operator/App directive ("consume as authoritative" per #1883), it only fires when the branch diff is empty, and #1883 already hard-fails on unresolved targets rather than silently dropping. Overriding an explicit directive with issue-text inference would break that contract, so it is deliberately left unchanged; if reconciliation is ever wanted there, it should be a separate follow-up decision on #1883's semantics.

sohni-tagirisa:
## Red test output (commit 854ae1369, before the fix)

The repro integration test fails with exactly the under-scope from #1980, and the helper unit tests fail on the missing extractor. The two passing tests are the intentional behavior-unchanged guards (superset scope, prose non-inflation), which must pass both before and after the fix.

```
  /Users/sohnitagirisa/.local/state/agent-worktrees/promptdriven__pdd/fix-1980/pdd/llm_invoke.py:1330: UserWarning: All LiteLLM cache configuration attempts failed. Caching is disabled.
    warnings.warn("All LiteLLM cache configuration attempts failed. Caching is disabled.")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_extracts_backticked_known_modules
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_extracts_prompt_file_paths
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_prose_word_matching_single_word_module_is_not_extracted
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_underscored_module_name_in_prose_is_extracted
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_underscored_name_inside_longer_token_is_not_extracted
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_runtime_llm_templates_never_extracted
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_no_architecture_extracts_nothing
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_union_issue_named_modules_missing_from_diff
=========== 8 failed, 2 passed, 237 deselected, 2 warnings in 2.48s ============
```

sohni-tagirisa:
## Mocked-E2E evidence (hermetic real-CLI run, post-fix)

The under-scope repro is exercised through the real code path (`python -m pdd --force --local sync <issue-url> --no-steer --dry-run --no-github-state` as a subprocess, all external binaries shimmed, zero billing): the fixture issue explicitly names both `prompts/greeter_python.prompt` and `prompts/textutil_python.prompt`, while the work branch commits a change to greeter only. The new test asserts (a) "Detected modules from branch diff" plus the reconciliation line "adding them to the sync scope: ['textutil']", (b) real child dry-run validation subprocesses ran for **both** `greeter` and `textutil`, and (c) `claude_calls.log` is empty — the fast path stayed at zero provider calls.

Full suite transcript (`python3 -m pytest tests/test_agentic_sync_mocked_e2e.py -vv`):

```
tests/test_agentic_sync_mocked_e2e.py::test_dry_run_pipeline_end_to_end PASSED [ 20%]
tests/test_agentic_sync_mocked_e2e.py::test_identification_failure_posts_error_comment PASSED [ 40%]
tests/test_agentic_sync_mocked_e2e.py::test_branch_diff_detection_skips_llm PASSED [ 60%]
tests/test_agentic_sync_mocked_e2e.py::test_branch_diff_scope_reconciled_with_issue_named_modules PASSED [ 80%]
tests/test_agentic_sync_mocked_e2e.py::test_resume_skips_succeeded_modules PASSED [100%]

=================== 5 passed, 2 warnings in 63.87s (0:01:03) ===================
```

(Re-run after the lint refactor: `5 passed in 68.43s`.) Unit suite: `python3 -m pytest tests/test_agentic_sync.py -q` → `247 passed` (includes the 10 new `TestBranchDiffIssueScopeReconciliation` tests). `python3 -m pylint pdd/agentic_sync.py` → no new findings vs the epic baseline (9.47/10; remaining findings in the touched line ranges are pre-existing, e.g. `_branch_diff_is_runtime_llm_only` R0911).

Note on `test_resume_skips_succeeded_modules`: its state fixture now marks both fixture modules succeeded, because the shared fixture issue explicitly names both prompts and the reconciled scope is correctly ['greeter', 'textutil']; the test's purpose (resume skips already-succeeded modules without any provider call) is unchanged.

sohni-tagirisa:
## Codex adversarial review — PR #1983 @ 09e29055b

**Verdict:** DON'T MERGE
**Findings:** 3 total (2 P1, 1 P2)

### P1 (blocking)
- [P1] Comment-only FILES_MODIFIED requests are ignored — `pdd/agentic_sync.py:2871-2908`
  The normal identify path preserves signal-bearing comments, including `FILES_MODIFIED`, but the branch-diff reconciliation extracts only from `title` and `body`. If the body only describes the issue and a later comment names `prompts/textutil_python.prompt`, the fast path still syncs only the diff modules and can report success while dropping the comment-requested module.
  **Suggested fix:** Run extraction over the same filtered issue content used for identify-modules, or pass `filtered_comments` into `_extract_issue_named_modules` and add a regression with a comment-only `FILES_MODIFIED` marker.

- [P1] Path-qualified duplicate-tail modules can be treated as already covered — `pdd/agentic_sync.py:252-380`
  Prompt-path tokens are reduced to just the filename, and `_issue_modules_missing_from_scope` treats any matching tail as covered. In a repo with `extensions/a/src/page` and `extensions/b/src/page`, a diff touching `a/page` can mask an issue request for backticked `extensions/b/src/page` or `extensions/b/prompts/src/page_python.prompt`, so the explicitly requested module is never added.
  **Suggested fix:** Preserve prompt-path qualification when converting prompt paths to module keys, and only use tail coverage when that tail is globally unambiguous; path-qualified issue modules should require exact full-key coverage.

### P2 (non-blocking)
- [P2] Bare hyphenated module names are never extracted — `pdd/agentic_sync.py:346-350`
  The bare prose matcher only considers tails containing `_`, so identifier-like basenames such as `github-app-api` or `check-run` are missed unless the issue uses backticks or a prompt path. That leaves a false-negative class parallel to underscored names.
  **Suggested fix:** Include hyphenated identifier tails in the escaped boundary matcher, or explicitly document and test that hyphenated modules require backticks/path references.

### Notes
- The runtime-only `_LLM.prompt` branch and `PDD_CHANGED_MODULES` fallback arm appear structurally unchanged.
- The mocked-E2E resume fixture update is legitimate for the current fixture because the issue body names both prompts.
- Targeted PR tests passed: `12 passed`, with only existing LiteLLM cache warnings.

*(Posted by the orchestrator on behalf of `codex exec` — sandboxed review had no network access. Orchestrator note: both P1s independently match my own review pass; confirmed FIX.)*

sohni-tagirisa:
## Implementer response — round 2

Thanks — all three findings are valid. Plan per finding:

**P1a (comment-only FILES_MODIFIED ignored) — Status: FIX.** Extraction will scan the same signal-bearing content the identify path uses: `_extract_issue_named_modules` gains a `comments` parameter and the step-7 call site passes the already-computed `filtered_comments` (post-`_filter_low_signal_comments`, so machine-noise comments stay excluded). Regression: helper test with a comment-only FILES_MODIFIED marker, plus a `run_agentic_sync` integration test where only a comment names the second module.

**P1b (duplicate-tail coverage masking) — Status: FIX.** Two changes: (i) prompt-path tokens are converted to path-qualified module keys using the same `<project-prefix>/<sub-path-under-prompts>/<stripped-basename>` derivation as `_detect_modules_from_branch_diff`, instead of being reduced to the filename; (ii) `_issue_modules_missing_from_scope` now takes the architecture inventory and only accepts tail-based coverage when the tail maps to exactly one known module — path-qualified issue keys otherwise require full-key (or consistent path-suffix) coverage. Regression: duplicate-tail inventory (`extensions/a/src/page` + `extensions/b/src/page`), diff covers a's, issue names b's path-qualified → union happens; unit tests for both the extraction qualification and the coverage predicate.

**P2 (hyphenated bare basenames never prose-matched) — Status: FIX (verified real).** Checked the derivation: nothing normalizes hyphens away — `_basename_from_architecture_filename("check-run_python.prompt")` returns `check-run` (language-suffix stripping only splits on the final `_<lang>` token), and `_filter_invalid_basenames`' non-prompt fallback stems (e.g. frontend `*.tsx` entries) make hyphenated identifiers plausible inventory members. The class-3 gate extends from `_` to `_` or `-` in the tail; the existing boundary lookarounds already exclude `-`-adjacent longer identifiers, so precision is preserved. Test: hyphenated basename mentioned in prose is extracted; embedded in a longer hyphenated token is not.

Process: red tests first as a separate `test(sync):` commit pushed before the fix, then implementation + targeted pytest + pylint, then red/green evidence comment.

sohni-tagirisa:
## Round-2 red/green evidence

**Red** (commit 11c9122e1, tests-only, pushed before the fix) — 8 failed / 13 passed in `TestBranchDiffIssueScopeReconciliation`:

```
E       TypeError: _extract_issue_named_modules() got an unexpected keyword argument 'comments'
tests/test_agentic_sync.py:5993: TypeError
>       assert named == ["extensions/b/src/page"]
E       AssertionError: assert [] == ['extensions/b/src/page']
tests/test_agentic_sync.py:6020: AssertionError
E       TypeError: _issue_modules_missing_from_scope() takes 2 positional arguments but 3 were given
tests/test_agentic_sync.py:6037: TypeError
E       TypeError: _issue_modules_missing_from_scope() takes 2 positional arguments but 3 were given
tests/test_agentic_sync.py:6049: TypeError
E       TypeError: _issue_modules_missing_from_scope() takes 2 positional arguments but 3 were given
tests/test_agentic_sync.py:6059: TypeError
>       assert named == ["check-run"]
E       AssertionError: assert [] == ['check-run']
tests/test_agentic_sync.py:6075: AssertionError
        assert success is True
        mock_agentic_task.assert_not_called()
>       assert sorted(dry_run_modules) == ["greeter", "textutil"]
E       AssertionError: assert ['greeter'] == ['greeter', 'textutil']
tests/test_agentic_sync.py:6148: AssertionError
        assert success is True
        mock_agentic_task.assert_not_called()
>       assert sorted(dry_run_modules) == [
E       AssertionError: assert ['extensions/a/src/page'] == ['extensions/...s/b/src/page']
tests/test_agentic_sync.py:6211: AssertionError
  /Users/sohnitagirisa/.local/state/agent-worktrees/promptdriven__pdd/fix-1980/pdd/llm_invoke.py:1330: UserWarning: All LiteLLM cache configuration attempts failed. Caching is disabled.
```

Failures map 1:1 to the findings: P1a (`test_extracts_modules_from_filtered_comments`, `test_union_from_comment_only_files_modified`), P1b (`test_prompt_path_tokens_preserve_path_qualification`, `test_duplicate_tail_does_not_mask_missing_issue_module`, `test_union_path_qualified_duplicate_tail_not_masked`, plus signature-change reds on the two coverage guards), P2 (`test_hyphenated_module_name_in_prose_is_extracted`). `test_backticked_path_qualified_key_resolves_exactly` and `test_hyphenated_name_inside_longer_token_is_not_extracted` pass pre-fix by design (guards).

**Green** (commit 84ae8d07c):
- `python3 -m pytest tests/test_agentic_sync.py -q` → **258 passed** (21/21 in the reconciliation class)
- `python3 -m pytest tests/test_agentic_sync_mocked_e2e.py -q` → **5 passed in 72.49s**
- `python3 -m pylint pdd/agentic_sync.py` → 9.49/10, zero findings in the touched line ranges (baseline epic tip: 9.47)

**One deliberate deviation from the suggested P1b fix:** instead of adding path-suffix coverage for path-qualified issue keys, coverage is now strictly *exact full key OR globally-unambiguous tail*. A path-suffix rule reintroduces the same masking through bare or partially-qualified keys (e.g. issue key `page` vs diff `extensions/a/src/page`), while a globally-unique tail provably refers to one module and safely bridges the key-form difference between the diff derivation (repo path, #1675) and architecture basenames. Duplicate-tail requests therefore require the exact known key — which is precisely what the extraction now produces from qualified prompt paths and backticked keys.

sohni-tagirisa:
## Codex adversarial review — PR #1983 @ 84ae8d07c

**Verdict:** DON'T MERGE
**Findings:** 3 total (2 P1, 1 P2)

### P1 (blocking)
- [P1] Duplicate filename entries still mask path-qualified issue modules — `pdd/agentic_sync.py:382-460`
  The stricter exact-or-globally-unique-tail rule is correct for the new directory-qualified test fixture, but it is not complete for existing supported architectures where distinct modules share the same `filename` basename and differ by prompt subdir/`filepath` (for example two Next.js `page_TypeScriptReact.prompt` entries). `_architecture_module_basenames()` dedupes those to one `page`, so `prompts/app/settings/page_TypeScriptReact.prompt` resolves to bare `page`, and `_issue_modules_missing_from_scope(["app/login/page"], ["page"], arch)` returns `[]`. That means a diff touching login still masks an explicit settings-page request, reopening the #1980 under-scope class.
  **Suggested fix:** Build the extraction/coverage inventory from real module identities including duplicate architecture entries and path/filepath-derived choices, and preserve path-qualified prompt-path tokens as scheduler keys instead of collapsing them to a bare known tail.

- [P1] Prompt path tokenizer drops supported route-segment characters — `pdd/agentic_sync.py:234-290`
  `_ISSUE_PROMPT_PATH_RE` only allows alnum, `_`, `.`, `/`, and `-`, so valid PDD/Next.js paths such as `prompts/app/users/[id]/page_TypeScriptReact.prompt` are tokenized as only `page_TypeScriptReact.prompt`. In duplicate `page` route projects this either extracts nothing or collapses to the wrong bare tail, so an explicit FILES_MODIFIED path for a dynamic route can still be silently omitted from the widened scope.
  **Suggested fix:** Replace the narrow prompt-path regex with a tokenizer that captures full non-whitespace `.prompt` paths while trimming markdown/list punctuation, and add a duplicate dynamic-route regression test.

### P2 (non-blocking)
- [P2] Surviving workflow comments and fenced noise can inflate scope — `pdd/agentic_sync.py:330-343`
  The new comment scan concatenates every body that survives `_filter_low_signal_comments`, but that filter intentionally only removes a few PDD sync/state prefixes. Other machine comments like `## Step 9/13...` survive, and `_issue_candidate_tokens` then treats any prompt path anywhere in those bodies (or fenced logs/code blocks in the issue text) as an explicit requested module. A stale or abandoned bot step comment containing `FILES_MODIFIED: prompts/textutil_python.prompt` can therefore add `textutil` even when the current title/body and branch diff do not request it.
  **Suggested fix:** For comments, scan only explicitly trusted marker payloads (or ignore known workflow step comments except the intended FILES_MODIFIED source), and avoid applying bare-name extraction to comment/log/code-fence text.

### Notes
- The runtime-LLM-only and `PDD_CHANGED_MODULES` arms remain structurally unchanged.
- The two-module mocked-E2E resume fixture is a legitimate consequence for the greeter/textutil fixture.

*(Round-2 re-review; posted by the orchestrator on behalf of `codex exec`.)*

sohni-tagirisa:
## Implementer response — round 3

All three findings are valid; all three are FIXING. The duplicate-`page` shape is confirmed real production data (pdd_cloud's Next.js routes, including bracketed dynamic segments), so P1s are treated as the App's normal case, not an edge.

**P1#1 (duplicate-filename entries mask path-qualified requests) — Status: FIXING.** Root cause: extraction/coverage consumed `_architecture_module_basenames`, which dedupes by derived basename, collapsing two same-filename entries into one identity. Fix: a real-identity index built per entry (duplicates preserved) — primary identity from the filename derivation, exact-match aliases additionally from `_architecture_entry_aliases` (filename + filepath-derived keys). Rules: tail-based resolution/coverage only when the tail maps to exactly one DISTINCT real identity (pre-dedup); path-qualified issue keys otherwise require exact coverage; and a prompt-path token that derives a path-qualified key via `_module_key_from_prompt_path` is accepted directly as a scheduler key (same key form the branch-diff detector already emits, #1675) provided its tail exists in the inventory. Regression: two same-tail entries, diff covers one, issue names the other via prompt path → union with the path-qualified key.

**P1#2 (tokenizer drops `[`/`]` route segments) — Status: FIXING.** `_ISSUE_PROMPT_PATH_RE` becomes a non-whitespace `.prompt` capture with leading markdown punctuation trimmed (backticks/quotes/parens; the `.prompt` anchor already excludes trailing punctuation), so `prompts/app/users/[id]/page_TypeScriptReact.prompt` keeps its full path. All downstream matching uses `re.escape`d tokens or plain string comparison, so brackets are inert. Regression: dynamic-route FILES_MODIFIED path resolves to the right path-qualified module in a duplicate-tail inventory.

**P2 (surviving workflow comments / fenced noise inflate scope) — Status: FIXING.** (a) Comments are no longer scanned as free text at all: after stripping fenced code blocks, only `FILES_MODIFIED:` marker payloads are consumed — parsed with the same `_extract_marker_paths` walker the change orchestrator produces/consumes those markers with — and only as class-2 prompt-path tokens. A marker quoted inside a log fence is treated as a log, not a directive. (b) In title/body, class-3 bare-name extraction runs on fence-stripped text (an unclosed fence conservatively drops the remainder); class-1 backticks and class-2 prompt paths stay global. Tests: fenced stale step-comment must not inflate scope; a genuine FILES_MODIFIED marker line must.

**Consciously unfixed (ACCEPTED, recall-only):** a backticked *bare* name whose tail is duplicated across real identities (e.g. `` `page` ``) still resolves to nothing — with duplicate identities there is no deterministic referent; the request needs a path-qualified key or prompt path (mirrors #1677's ambiguity contract). Also, a *stale-but-well-formed* FILES_MODIFIED marker outside a fence in a surviving comment remains indistinguishable from a genuine one by format; scope inflation there is bounded to known-inventory prompt paths and mirrors what the LLM-identify path reads from the same comment.

Process: red tests first as a separate `test(sync):` commit pushed before the fix, then implementation + targeted pytest + pylint, then evidence reply.

sohni-tagirisa:
## Round-3 red/green evidence

**Red** (commit c7186e0e7, tests-only, pushed before the fix) — 8 failed / 23 passed in `TestBranchDiffIssueScopeReconciliation`:

```
  /Users/sohnitagirisa/.local/state/agent-worktrees/promptdriven__pdd/fix-1980/pdd/llm_invoke.py:1330: UserWarning: All LiteLLM cache configuration attempts failed. Caching is disabled.
    warnings.warn("All LiteLLM cache configuration attempts failed. Caching is disabled.")
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_duplicate_filename_entries_do_not_collapse_to_one_identity
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_bare_backtick_of_duplicated_tail_resolves_to_nothing
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_duplicate_filename_tail_does_not_cover_missing_module
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_bracketed_dynamic_route_prompt_path_is_tokenized
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_comment_prompt_path_outside_marker_is_ignored
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_comment_marker_inside_log_fence_is_ignored
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_fenced_code_in_body_does_not_bare_name_match
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_union_duplicate_filename_prompt_path_not_masked
8 failed, 23 passed, 237 deselected, 2 warnings in 1.81s
```

Failure-to-finding map: P1#1 — `test_duplicate_filename_entries_do_not_collapse_to_one_identity`, `test_bare_backtick_of_duplicated_tail_resolves_to_nothing`, `test_duplicate_filename_tail_does_not_cover_missing_module`, `test_union_duplicate_filename_prompt_path_not_masked`; P1#2 — `test_bracketed_dynamic_route_prompt_path_is_tokenized`; P2 — `test_comment_prompt_path_outside_marker_is_ignored`, `test_comment_marker_inside_log_fence_is_ignored`, `test_fenced_code_in_body_does_not_bare_name_match`. The two passing new tests (`test_comment_genuine_marker_line_still_extracts`, `test_bare_name_outside_fence_still_matches`) are pre-fix guards that must stay green.

**Green** (commit 8bd157559):
- `python3 -m pytest tests/test_agentic_sync.py -q` → **268 passed** (31/31 in the reconciliation class)
- `python3 -m pytest tests/test_agentic_sync_mocked_e2e.py -q` → **5 passed in 67.10s**
- `python3 -m pylint pdd/agentic_sync.py` → 9.51/10, zero findings in the touched line ranges (epic baseline: 9.47)

**Implementation notes:** the identity index reuses the existing per-entry alias derivation (`_architecture_entry_aliases`: filename- and filepath-derived keys) rather than inventing a new one; the trusted-marker parser is the change orchestrator's own `_extract_marker_paths` (bullets, inline terminators, multi-line payloads all behave exactly as the marker producer defines); and a prompt-path-derived qualified key is scheduled in the same key form the branch-diff detector emits, which downstream validation (`_filter_invalid_basenames` tail check, #1677 ambiguity handling) already accepts. Accepted recall trade-offs stated in the round-3 response comment stand: a bare backticked name colliding with a duplicated tail resolves to nothing (needs a path), and a well-formed unfenced FILES_MODIFIED marker in a surviving comment is trusted by format.

sohni-tagirisa:
## Codex adversarial review — PR #1983 @ 8bd157559

**Verdict:** DON'T MERGE
**Findings:** 4 total (2 P1, 2 P2)

### P1 (blocking)
- [P1] Combined architecture duplicates still collapse across project scopes — `pdd/agentic_sync.py:328-334`
  `load_combined_architecture_data()` concatenates nested architecture entries without source-path metadata, but the new index dedupes on only `(primary, filepath_str)`. Two projects can legitimately both declare `filename: "src/page_python.prompt"` and `filepath: "src/page.py"`; the second entry is skipped, so `tail_to_identities["page"]` looks globally unique. Then a branch diff for `extensions/a/src/page` can be considered to cover an issue request for `extensions/b/prompts/src/page_python.prompt`, silently recreating the under-scope failure this PR is meant to prevent.
  **Suggested fix:** Preserve architecture scope in the identity key, or stop deduping architecture-only entries unless they are proven to be the same absolute output.

- [P1] Prompt-path scheduler keys can lose dependency edges — `pdd/agentic_sync.py:498-503`
  The direct path-token branch emits keys like `app/settings/page` when only the tail exists in the architecture. That key can be accepted by the child sync via filepath-suffix matching, but `_build_targeted_dep_graph()` later asks `build_dep_graph_from_architecture_data()` to find an architecture entry by aliases, where flat-filename entries expose `page` / `src/app/settings/page`, not `app/settings/page`. The result is an empty dependency list with no warning for a newly widened module that may have declared dependencies.
  **Suggested fix:** Resolve prompt-path-derived keys back to the matched architecture entry, or teach targeted dependency graph construction the same filepath-suffix matching used by sync resolution.

### P2 (non-blocking)
- [P2] Fenced issue-body tokens still widen scope — `pdd/agentic_sync.py:557-562`
  Comments strip fences before marker parsing, and class-3 body matching uses stripped text, but class-1 backticks and class-2 prompt paths scan the raw title/body. A fenced prior-run log in the issue body containing `FILES_MODIFIED: prompts/textutil_python.prompt` still adds `textutil` to the write scope.
  **Suggested fix:** Remove fenced code blocks from the issue body before class-1/class-2 extraction as well.

- [P2] Longer Markdown fences are parsed as alternating triple fences — `pdd/agentic_sync.py:261-283`
  Valid Markdown fences can be four or more backticks/ tildes. The stripper stores only the first three characters, so a four-backtick quote containing a triple-fenced snippet can leak the text between inner triple fences; that leaked text can include a marker payload or bare module name.
  **Suggested fix:** Track the full opening fence character and length, and close only on a fence of the same character with at least that length.

### Notes
- Targeted reconciliation tests and the two updated mocked-E2E tests passed locally.

*(Round-3 re-review; posted by the orchestrator on behalf of `codex exec`. Orchestrator triage: P1s go to verify-first — implementer must confirm or refute against `load_combined_architecture_data` and `_build_targeted_dep_graph` behavior before coding; both P2s are mechanical FIXes.)*

sohni-tagirisa:
## Implementer response — round 4

Branch rebased onto the updated epic base (263b4925f, post-#1982 squash) via clean cherry-pick; `gh pr diff` still shows only the three PR files. Verdicts after verify-first:

**P1#1 (cross-project identity collapse) — Status: VERIFIED, FIXING.** Confirmed against the source: `load_combined_architecture_data` (`pdd/architecture_registry.py`) concatenates every architecture file's entries with `combined.extend(extract_modules(data))` — no source-path qualification whatsoever — so two nested projects that both declare `filename: "src/page_python.prompt"` / `filepath: "src/page.py"` yield two string-identical entries, and my `(primary, filepath)` dedupe merges them into one falsely-unique tail. Fix: drop the dedupe entirely. With an unqualified combined inventory, string-identical entries are *indistinguishable*, so the only safe deterministic treatment is ambiguity (no bare resolution, no tail coverage — exact/path-qualified requests still work). Cost is recall-only, confined to a module literally double-listed with identical strings; that trade-off is stated in the code comment.

**P1#2 (dependency edges lost for prompt-path-derived keys) — Status: VERIFIED (empirically), FIXING.** Reproduced directly: `build_dep_graph_from_architecture_data(arch, ["app/settings/page", "textutil"])` returns `{'app/settings/page': [], ...}` with **zero warnings**, while the filepath alias `src/app/settings/page` resolves the declared `textutil` edge. Note the gap predates this PR — a plain branch-diff key for the same file (`prompts/app/settings/page_TypeScriptReact.prompt` → `app/settings/page`) hits the identical silent edge loss — my widening just extended its reach. Fix at graph construction (Codex's second suggested option, which heals both producers): `_build_targeted_dep_graph` pre-resolves any group target that fails alias matching to the UNIQUELY suffix-matched entry's filepath-derived alias before calling the builder, and remaps the result back to the original scheduler keys. Scheduler keys themselves stay unchanged — deliberately, because the diff detector emits the prompt-subpath form for the same file, and switching the issue-side key to a filepath alias would break exact coverage in the superset case and double-schedule one module. Regression test asserts the widened module's declared dependencies survive into the dep graph, with a uniqueness guard (ambiguous suffix → left unresolved, existing behavior).

**P2#1 (fenced title/body tokens widen scope via classes 1/2) — Status: FIXING.** Mechanical: title/body is fence-stripped once, before all three classes (comments were already stripped before marker parsing). Test: fenced prior-run log containing a FILES_MODIFIED line and an inline-backticked module name must not widen scope; unfenced tokens still do.

**P2#2 (4+ char fences leak inner triple fences) — Status: FIXING.** Mechanical: the stripper now records the opening fence character and length and closes only on a fence of the same character with >= length (CommonMark rule). Test: a ````-fenced block containing an inner ``` pair leaks nothing.

Process: red tests first as a separate `test(sync):` commit pushed before the fix, then implementation + targeted pytest + mocked E2E + pylint, then evidence reply.

sohni-tagirisa:
## Round-4 red/green evidence

**Rebase:** branch cherry-picked cleanly onto the updated epic base 263b4925f (post-#1982 squash) before this round; `gh pr diff 1983 --name-only` still shows exactly `pdd/agentic_sync.py`, `tests/test_agentic_sync.py`, `tests/test_agentic_sync_mocked_e2e.py`.

**Red** (commit 1cfe8a62f, tests-only, pushed before the fix) — 6 failed / 32 passed in `TestBranchDiffIssueScopeReconciliation`:

```
  /Users/sohnitagirisa/.local/state/agent-worktrees/promptdriven__pdd/fix-1980/pdd/llm_invoke.py:1330: UserWarning: All LiteLLM cache configuration attempts failed. Caching is disabled.
    warnings.warn("All LiteLLM cache configuration attempts failed. Caching is disabled.")
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_string_identical_cross_project_entries_stay_ambiguous_for_coverage
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_string_identical_entries_bare_backtick_resolves_to_nothing
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_prompt_path_scheduler_key_keeps_declared_dependencies
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_fenced_body_content_does_not_widen_scope_via_classes_1_and_2
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_quad_backtick_fence_does_not_leak_inner_content
FAILED tests/test_agentic_sync.py::TestBranchDiffIssueScopeReconciliation::test_quad_fence_comment_marker_does_not_extract
6 failed, 32 passed, 237 deselected, 2 warnings in 1.76s
```

Failure-to-finding map: P1#1 — `test_string_identical_cross_project_entries_stay_ambiguous_for_coverage`, `test_string_identical_entries_bare_backtick_resolves_to_nothing`; P1#2 — `test_prompt_path_scheduler_key_keeps_declared_dependencies` (the silent empty-deps repro, asserted through `_build_targeted_dep_graph` with a real tmp architecture.json); P2#1 — `test_fenced_body_content_does_not_widen_scope_via_classes_1_and_2`; P2#2 — `test_quad_backtick_fence_does_not_leak_inner_content`, `test_quad_fence_comment_marker_does_not_extract`. `test_ambiguous_suffix_target_left_unresolved_in_dep_graph` passes pre-fix by design (no-guess guard that must stay green after the fix).

**Green** (commit f680c7614):
- `python3 -m pytest tests/test_agentic_sync.py -q` → **275 passed** (38/38 in the reconciliation class)
- `python3 -m pytest tests/test_agentic_sync_mocked_e2e.py -q` → **5 passed in 70.23s**
- `python3 -m pylint pdd/agentic_sync.py` → 9.52/10; the only finding in a touched range is `_build_targeted_dep_graph` R0914 at 17/15, which pre-exists at exactly that count on the parent commit (verified by pylinting both versions)

**Notes:** dep-graph pre-resolution deliberately keeps scheduler keys unchanged (rewrites only the internal graph-construction target and remaps back), so exact coverage against branch-diff keys and child-sync acceptance are untouched — and the fix also heals the pre-existing silent edge loss for plain branch-diff keys against flat-filename entries. The identity index's no-dedupe trade-off (string-identical double-listing loses bare-name recall only) is documented in the code.

sohni-tagirisa:
## Round-4 evidence (posted by orchestrator — implementer was interrupted by a transient API outage after pushing)

Commits: red tests `1cfe8a62f` pushed before fix `f680c7614`, per the TDD discipline. Working tree clean; branch contains only intentional commits.

Verification re-run independently by the orchestrator on the pushed tip:
- `pytest tests/test_agentic_sync.py -k "Reconciliation or fence or cross_project or dep_graph or identity"` → **46 passed**
- `pytest tests/test_agentic_sync_mocked_e2e.py` → **5 passed** (64s, hermetic, zero provider calls)

Round-4 changes as planned in the response comment above: identity dedupe dropped (string-identical cross-project duplicates now resolve as ambiguous rather than falsely unique), `_build_targeted_dep_graph` pre-resolves suffix-unique targets so widened modules keep their declared dependency edges (also heals the pre-existing branch-diff-key edge loss), title/body fence-stripped once before all three token classes, and fence open/close now follows the CommonMark char+length rule.

sohni-tagirisa:
**Target completion: Wednesday, July 15, 2026.**

This PR is currently mergeable and its auto-heal checks are green. The remaining target is final maintainer review and merge readiness so the dependent epic can be refreshed.

PR #1988, opened 2026-07-11
fix(sync): add trusted Jest verification adapter

gltanaka:
## Summary
Adds a fail-closed Jest adapter to the trusted sync runner.

## Investigation
- Existing `runner.py` dispatches only `pytest`; the Jest adapter follows its base-clone identity, support-closure, dirty-support, exact execution, and secret-scrubbing conventions.
- Stable identity: `relative-test-file::ancestor title > test title`, emitted only by a temporary trusted reporter. Candidate code never supplies protected expected identities.
- Supported configuration is static JSON (`jest.config.json` or `package.json` `jest` object). The closure binds config and local setup, transform, reporter, resolver, project, and static relative-import support. Executable/dynamic Jest config (`jest.config.js/cjs/mjs/ts`) fails closed.
- pdd_cloud currently has executable Jest configs, so they are deliberately rejected until an independently reviewable static-config policy exists. Vitest and Playwright are not supported or claimed.
- Prior runner hardening: `f39dd6090` established config-closure binding and credential scrubbing; this PR follows that approach. No sibling trusted-runner dispatch sites require changes.

## Test plan
- `conda run -n pdd pytest -q tests/test_sync_core_runner.py tests/test_sync_core_runner_jest.py`
- `conda run -n pdd pylint pdd/sync_core/runner.py`
- `git diff --check`
- Real local Jest 29.7.0 fixture E2E using `/Users/gregtanaka/Developer/pdd_cloud/frontend/node_modules/jest/bin/jest.js`
- `python -m build --wheel --outdir /tmp/pdd-jest-wheel-dist`

PR #1993, opened 2026-07-11
fix(sync): add trusted Vitest verification adapter

gltanaka:
## Summary
- Add a distinct trusted `validator_id=vitest` runner adapter.
- Bind static Vitest config, setup and transform support closure; reject dynamic config, plugin/workspace/project controls, filtering/sharding/watch controls, and dirty executable modules.
- Derive protected-base identities independently and execute exact protected paths via the local Vitest binary with a sanitized environment.

## Test plan
- `conda run -n pdd pytest -q tests/test_sync_core_runner.py tests/test_sync_core_runner_jest.py tests/test_sync_core_runner_vitest.py`
- `conda run -n pdd pylint pdd/sync_core/runner.py tests/test_sync_core_runner_vitest.py`
- Real fixture through `run_profile` using `node_modules/vitest/vitest.mjs`: `PASS 1 protected Vitest tests passed`
- Wheel/source parity SHA-256 check for `pdd/sync_core/runner.py`


gltanaka:
## Pre-fix investigation

- Base verified: `origin/codex/issue-1932-jest-adapter` at `dc6d5ccd8` (`fix(sync): fail closed for unbound Jest project plugins`).
- History mined: `dc74dd60c` / `98e5e9e2b` / `dc6d5ccd8` establish the trusted JS runner convention: static config only, protected-base identity collection, direct local binary argv, sanitized environment, and fail-closed closure controls.
- Sibling hunt: no trusted Vitest adapter existed. The only related trusted-runner implementation is Jest; this PR deliberately adds a separate `vitest` adapter and does not alter pytest or Jest behavior.
- Scope decision: `pdd/sync_core/runner.py` plus dedicated `tests/test_sync_core_runner_vitest.py` only. No aliases, transaction/certificate/lifecycle work, includes/preprocess/CI changes, or `.pdd` files.

gltanaka:
## TDD evidence: red then green

**Red commit:** `fafcd298e test(sync): reproduce missing trusted Vitest adapter`

Command:
```text
conda run -n pdd pytest -q tests/test_sync_core_runner_vitest.py
```

Result before the fix: `16 failed`; the first failure was `TypeError: RunnerConfig.__init__() got an unexpected keyword argument 'vitest_command'`, reproducing the absence of a trusted Vitest adapter surface. This commit was pushed before the implementation commit.

**Green commit:** `d4b1ba671 fix(sync): add trusted Vitest verification adapter`

Commands:
```text
conda run -n pdd pytest -q tests/test_sync_core_runner.py tests/test_sync_core_runner_jest.py tests/test_sync_core_runner_vitest.py
conda run -n pdd pylint pdd/sync_core/runner.py tests/test_sync_core_runner_vitest.py
```

Result: both commands passed. The focused Vitest contract suite contains 20 passing cases covering pass/fail/skip/todo/zero/timeout/malformed output, identity mismatches, protected test/config/setup/transform mutation, dirty support, secret stripping, dynamic config, and unbound watch/shard/projects/plugins.

gltanaka:
## E2E and parity evidence

Created a disposable local Git fixture with real `vitest@4.0.18`, a static `vitest.config.json`, and `tests/widget.test.js`. Invoked the production runner path through `run_profile`; it directly ran `node_modules/vitest/vitest.mjs run` against both the protected-base clone and checked head.

Result:
```text
PASS 1 protected Vitest tests passed
```

Also built a wheel outside the repository and compared SHA-256 of source versus wheel `pdd/sync_core/runner.py`: `wheel/source runner.py parity: PASS`.

Final stack check: rebased against freshly fetched `origin/codex/issue-1932-jest-adapter`; it was already current at `dc6d5ccd8`, so no history rewrite or force-with-lease push was needed.

gltanaka:
## Refreshed stacked-base evidence

The Jest parent was refreshed by the stack maintainer. I fetched it and rebased this PR onto exact parent `92dee253589e4a071c6e0038a8c0af8330394dfa` before rerunning verification.

Rewritten stacked commits now on the PR:
- `360f48213 test(sync): reproduce missing trusted Vitest adapter`
- `4fa03324c fix(sync): add trusted Vitest verification adapter`

Post-rebase verification passed:
```text
conda run -n pdd pytest -q tests/test_sync_core_runner.py tests/test_sync_core_runner_jest.py tests/test_sync_core_runner_vitest.py
conda run -n pdd pylint pdd/sync_core/runner.py tests/test_sync_core_runner_vitest.py
```

The real local `vitest@4.0.18` fixture again returned `PASS 1 protected Vitest tests passed`; wheel/source `runner.py` SHA-256 parity also passed. Because rebase rewrote the child commits, the verified branch was updated with `git push --force-with-lease` only.

Final state: clean worktree, no unpushed commits, PR base remains `codex/issue-1932-jest-adapter`.

gltanaka:
## Codex xhigh adversarial review — PR #1993 @ bbafa28f3805db3c3251f5c41cbd12de727e6771
**Verdict:** DON'T MERGE
**Findings:** 2 total (2 P1, 0 P2)
### P1 (blocking)
- [P1] Vitest closure drops protected-test imports — pdd/sync_core/runner.py:439-471,1327-1357
  `_vitest_support_closure` explicitly discards `test_paths`, so the hashed/quarantined closure contains only the static config plus setup/transform support. `run_obligation` checks changes in `artifact_paths | support_paths`, but committed helpers imported by the protected test are absent from `support_paths`; a candidate can change `tests/helper.ts` (or a transitive local module) while leaving the protected test file and identities unchanged, and the head Vitest execution can still PASS with candidate-controlled helper behavior.
  **Suggested fix:** Start the Vitest support closure from the protected test paths, recursively include literal local imports from those tests and their helpers, and add a regression where a committed imported helper changes without touching the test file.
- [P1] The Vitest adapter trusts an unbound candidate Node toolchain — pdd/sync_core/runner.py:790-900,1492-1518
  The default command resolves `root/node_modules/vitest/vitest.mjs`, while protected-base collection deliberately runs the base clone using `command_root=root`. That means the supposedly trusted collector/executor can come from ignored candidate-worktree package contents that are not part of the protected SHA, the closure digest, or the dirty support check. A modified local Vitest CLI can emit the expected JSON identities/statuses and make forged evidence look trusted.
  **Suggested fix:** Run Vitest only from a protected, pinned external toolchain whose digest/version is included in the runner identity, or bind and verify the full local package/tool closure before collection and execution.
### P2 (non-blocking)
- None.
### Notes
- Reviewed against actual base `codex/issue-1932-jest-adapter` @ `39e26901f4053e22ebdd297e9be5d2586b4c7ff6`, not `main`.

gltanaka:
## Implementer response to Codex xhigh review

@codex Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4942748315

### P1 #1: Vitest closure drops protected-test imports
Status: **FIXING**
Plan: Add failing coverage for a changed committed helper imported by a protected Vitest test, then include recursive local imports from protected tests in the bound support closure.

### P1 #2: Vitest adapter trusts an unbound candidate Node toolchain
Status: **FIXING**
Plan: Add failing coverage for ignored local `node_modules` Vitest trust, then require the Vitest toolchain to be digest-bound/protected or reject it before accepting trusted evidence.

gltanaka:
## Implementer evidence

Fixed both xhigh P1 findings and rebased this stacked branch onto updated PR #1988.

Commits:
- `cb0df824a` test: cover Vitest trust boundary gaps
- `3bbe50030` fix: bind Vitest support closure and reject local runner

TDD red evidence before fix:
- `pytest -q tests/test_sync_core_runner_vitest.py::test_vitest_imported_test_helper_mutation_cannot_pass tests/test_sync_core_runner_vitest.py::test_default_candidate_node_modules_vitest_is_not_trusted`
- Result: 2 failed. Imported helper mutation and ignored candidate `node_modules/vitest` both incorrectly PASSed.

Green verification:
- `pytest -q tests/test_sync_core_runner_vitest.py tests/test_sync_core_runner_jest.py` -> 43 passed, 1 warning
- `pylint --disable=line-too-long,too-many-lines,too-many-branches,too-many-return-statements,too-many-locals,missing-function-docstring,too-many-arguments,too-many-positional-arguments,duplicate-code pdd/sync_core/runner.py tests/test_sync_core_runner_vitest.py tests/test_sync_core_runner_jest.py` -> 10.00/10

Trust-boundary note: default candidate-checkout `node_modules/vitest` now fails closed; trusted Vitest execution requires an explicit external command rather than candidate-local infrastructure.

Pushed: `3bbe500304c3d0db3da4c5dc6dbeabe910de5d3b`.

gltanaka:
## Codex xhigh adversarial re-review — PR #1993 @ 8b5717551e715b1ca5933a366bd88725eea0eb47

**Verdict:** DON'T MERGE
**Findings:** 3 total (3 P1, 0 P2)

### P1 (blocking)
- [P1] Explicit `vitest_command` can still run a candidate-controlled validator — `pdd/sync_core/runner.py:83`, `pdd/sync_core/runner.py:810`, `pdd/sync_core/runner.py:888`, `pdd/sync_core/runner.py:571`
  The default candidate `node_modules/vitest` path now fails closed, but `_vitest_command()` accepts any explicit argv and `runner_identity_digest()` records only a placeholder command, not the resolved command or toolchain digest. I verified from this head that an ignored candidate-local script passed as `RunnerConfig(vitest_command=(python, <root>/node_modules/vitest/vitest.py))` can emit a forged JSON PASS for both collection and execution. That still violates the prior trust-boundary fix: the trusted validator can be supplied by the candidate checkout and the final attestation does not bind it.
  **Suggested fix:** Reject explicit Vitest commands under the candidate checkout, require a protected/pinned external toolchain, and include the resolved tool identity/digest in the runner identity or signed evidence.

- [P1] Vitest support closure still misses side-effect imports — `pdd/sync_core/runner.py:337`, `pdd/sync_core/runner.py:464`, `pdd/sync_core/runner.py:1346`
  The Vitest closure relies on the shared `_local_javascript_imports()` regex, which does not match `import './helper';`. I verified a protected `tests/widget.test.ts` with that side-effect import produces a closure containing only `tests/widget.test.ts` and `vitest.config.json`; mutating `tests/helper.ts` leaves `vitest_validator_config_digest(...)` unchanged. A candidate can therefore alter committed side-effect support used by the protected test without quarantine.
  **Suggested fix:** Parse JS/TS imports with a real parser or fail closed on unsupported import syntax; add Vitest regressions for side-effect and transitive side-effect imports.

- [P1] Vitest aliases can hide mutable local code outside the closure — `pdd/sync_core/runner.py:422`, `pdd/sync_core/runner.py:464`
  `_vitest_config_references()` only binds setup and transform paths and does not reject or bind Vite/Vitest resolution aliases. I verified `vitest.config.json` with `{"resolve":{"alias":{"@":"./src"}}}` plus a protected test importing `@/helper` leaves `src/helper.ts` out of the closure; mutating that helper leaves the digest and changed-path guard unchanged. Alias targets are normal executable test dependencies, so this can produce a trusted PASS using candidate-controlled support code.
  **Suggested fix:** Fail closed on unsupported `resolve.alias`/`test.alias` forms or resolve and bind local alias targets transitively before accepting Vitest evidence.

### P2 (non-blocking)
- None.

### Notes
- Live state reviewed: open PR, actual base `53891daf84b6c9f2bf2c53fcf19f41bd6533ab4c`, head `8b5717551e715b1ca5933a366bd88725eea0eb47`; merge-base equals the base.
- Prior xhigh P1 #1 is only partially resolved: named imports are covered, but side-effect imports and alias-resolved local modules remain unbound.
- Prior xhigh P1 #2 is only partially resolved: default candidate `node_modules` is rejected, but explicit candidate-local `vitest_command` is still trusted and not bound into the signed runner identity.

gltanaka:
## Implementer response to Codex xhigh re-review

@codex Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4943055893

### P1 #1: Explicit `vitest_command` can still run a candidate-controlled validator
Status: **FIXING**
Plan: Add failing coverage for explicit candidate-local Vitest commands, then reject commands under the candidate checkout and bind accepted external command identity/digest into runner evidence.

### P1 #2: Vitest support closure still misses side-effect imports
Status: **FIXING**
Plan: Add failing coverage for direct and transitive side-effect imports, then extend shared JS import discovery or fail closed on unsupported syntax.

### P1 #3: Vitest aliases can hide mutable local code outside the closure
Status: **FIXING**
Plan: Add failing alias coverage for local `resolve.alias`/`test.alias`, then bind local alias targets transitively or fail closed on unsupported alias forms.

gltanaka:
## Implementer evidence

Fixed all three xhigh re-review P1 findings and rebased onto updated PR #1988.

Commits:
- `84a312485` test: cover remaining Vitest trust gaps
- `50415ee61` fix: harden Vitest command and aliases

TDD red evidence before fix:
- Targeted new Vitest cases for explicit candidate-local command trust, local `resolve.alias`, and side-effect import closure coverage
- Result: alias and explicit candidate command failed as expected; side-effect import was already green from the shared JS import fix inherited from #1988.

Green evidence:
- Targeted new Vitest cases -> 3 passed
- `pytest -q tests/test_sync_core_runner_vitest.py tests/test_sync_core_runner_jest.py` -> 51 passed, 1 warning

Trust-boundary note: explicit Vitest commands inside the candidate checkout now fail closed; accepted explicit command identity is folded into runner evidence; `resolve.alias`/`test.alias` fail closed instead of leaving local code outside the closure.

gltanaka:
## Codex xhigh adversarial re-review round 3 — PR #1993 @ 9d83bf3944267fd739e68fd4da2eaa2f23ba13f5

**Verdict:** DON'T MERGE
**Findings:** 1 total (1 P1, 0 P2)

### P1 (blocking)
- [P1] Hardened Vitest validation is not reachable through the production CLI — `pdd/commands/sync_core.py:347`, `pdd/commands/sync_core.py:355`, `pdd/sync_core/runner.py:973`
  The fixed adapter now rejects candidate `node_modules/vitest` and requires an explicit external `RunnerConfig.vitest_command`. The production `pdd validate` command still calls `finalize_unit(...)` with the default `RunnerConfig()` and has no `--vitest-command` or protected env equivalent. Any Vitest obligation finalized through the CLI therefore fails before executing trusted evidence (`vitest-unavailable`, or candidate `node_modules` untrusted), so the adapter is only usable by direct Python callers that bypass the product entrypoint.
  **Suggested fix:** Add protected CLI/env plumbing for an absolute external Vitest command, resolve/hash that command identity in runner evidence, reject candidate-local values, and add an end-to-end `pdd validate` regression with a Vitest obligation.

### P2 (non-blocking)
- None.

### Notes
- Live state reviewed: open PR, actual base `cd292cd0c1bad5043e5c92c7a8a9e84b11f1806e`, head `9d83bf3944267fd739e68fd4da2eaa2f23ba13f5`; merge-base equals the base.
- Prior xhigh findings verified resolved: protected-test imports and side-effect imports are in the closure, `resolve.alias`/`test.alias` fail closed, default candidate `node_modules/vitest` is rejected, and explicit candidate-local `vitest_command` is rejected and command identity is included in runner evidence.

gltanaka:
## Implementer response to Codex xhigh re-review round 3

@codex Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4943358603

### P1 #1: Hardened Vitest validation is not reachable through the production CLI
Status: **FIXING**
Plan: Add protected CLI/env plumbing for an absolute external Vitest command, reject candidate-local values before runner execution, bind accepted command identity in evidence, and cover production `pdd validate`/finalize wiring with a Vitest obligation regression.

gltanaka:
## Implementer evidence

Fixed the round-3 P1 finding for #1993 and rebased onto updated #1988.

Commits:
- `c85e11a18` test: cover Vitest validate CLI wiring
- `4bf673c12` fix: wire protected Vitest validation CLI

TDD red evidence before fix:
- `pytest -q tests/test_sync_core_reporting.py::test_validate_command_wires_protected_vitest_runner_config`
- Result: failed because `pdd validate` had no `--vitest-command` option.

Green evidence:
- Focused Vitest CLI case + Jest CLI case + Jest/Vitest adapter suites: `pytest -q tests/test_sync_core_reporting.py::test_validate_command_wires_protected_vitest_runner_config tests/test_sync_core_reporting.py::test_validate_command_wires_protected_jest_runner_config tests/test_sync_core_runner_vitest.py tests/test_sync_core_runner_jest.py` -> 54 passed, 1 warning

Trust-boundary note: production `pdd validate` now accepts protected external Vitest command configuration through CLI/env and carries it into `RunnerConfig` without giving candidate checkout paths authority.

Pushed head: `4bf673c1286220a05e53ba7f604832a883eac123`.

gltanaka:
## Codex xhigh adversarial re-review round 4 — PR #1993 @ bbf7541430735f437816a8e2f001f9ea5f08f728

**Verdict:** DON'T MERGE
**Findings:** 1 total (1 P1, 0 P2)

### P1 (blocking)
- [P1] Vitest closure misses parent-directory imports — `pdd/sync_core/runner.py:360-379`, `pdd/sync_core/runner.py:493-510`, `pdd/sync_core/runner.py:1467-1472`
  The Vitest support closure relies on `_local_javascript_imports()`, which builds `source_path.parent / PurePosixPath(item)` but does not normalize `..` segments before probing Git. A normal protected spec importing `../src/helper` is therefore treated as `tests/../src/helper.ts`, missed by `read_git_blob(...)`, and omitted from `support_paths`. Since the quarantine check only watches the incomplete closure, mutating `src/helper.ts` can alter the checked-head Vitest run without invalidating the trusted evidence. Focused probe against this head: `support_paths == ['tests/widget.test.ts', 'vitest.config.json']`, and both `vitest_validator_config_digest(...)` and `_vitest_support_digest(...)` stayed unchanged after mutating `src/helper.ts`.
  **Suggested fix:** Normalize and validate literal relative JS/TS imports before Git lookup, reject escaping paths, include normalized file/index candidates transitively, and add Vitest regressions for `import '../src/helper'` and side-effect parent imports.

### P2 (non-blocking)
- None.

### Prior xhigh verification
- Prior findings verified resolved for protected-test `./helper` imports, side-effect imports, aliases, default/explicit candidate-local `vitest_command`, command identity binding, and production `--vitest-command` CLI wiring.
- This remains a separate parent-directory import resolution gap.

### Verification
- Live base/head: `codex/issue-1932-jest-adapter` @ `13af150edb39d152d02339673f06b6e10f728b09`; head `bbf7541430735f437816a8e2f001f9ea5f08f728`; merge-base equals base.
- `conda run --no-capture-output -n pdd pytest -q tests/test_sync_core_runner_vitest.py tests/test_sync_core_reporting.py` -> 45 passed, 1 warning.
- `git diff --check 13af150edb39d152d02339673f06b6e10f728b09..bbf7541430735f437816a8e2f001f9ea5f08f728` -> clean.

gltanaka:
Rebased onto updated #1988 and pushed.

Changed: Vitest uses the shared JS support-path resolver for file/index candidates; validator digests include transitive test imports and config support candidates; added Vitest direct/side-effect/parent-dir/index/config-digest/escape regressions.

Evidence: focused #1993 stack passed: 81 passed. Top-stack suite from #1997 passed: 174 passed. Real JS smoke passed: Jest 30.4.1, Vitest 4.1.10, Playwright 1.61.1. git diff --check clean.

Lint: full make lint completed but failed on broad existing pylint findings, mainly unrelated cyclic-import reports; targeted pylint also hits existing large adapter style thresholds.

gltanaka:
## Codex xhigh adversarial re-review round 5 — PR #1993 @ 6607caf4fcc1c2442ca4c0bb41321310043eec95

**Verdict:** DON'T MERGE
**Findings:** 1 total (1 P1, 0 P2)
**Approval:** Not approved; unresolved stack P1 remains.

### Live State
- Base: `codex/issue-1932-jest-adapter` @ `2a38d3eb8ca7f2c6d634a40146b29ba875b26c08`
- Head: `codex/issue-1932-vitest-adapter` @ `6607caf4fcc1c2442ca4c0bb41321310043eec95`
- Merge state: `CLEAN`
- Current checks: `auto-heal` SUCCESS, `heal` SUCCESS

### P1 (blocking)
- [P1] Pathless explicit JS validator command arguments can still execute candidate-controlled code — `pdd/commands/sync_core.py:143-167`, `pdd/sync_core/runner.py:740-758`, `pdd/sync_core/runner.py:932-942`, `pdd/sync_core/runner.py:1092-1121`
  The Vitest adapter inherits the same trust gap as the Jest parent and extends it to `vitest_command`. The command parser only requires argv[0] to be absolute, while `_command_uses_candidate_checkout()` skips later tokens with no `/`. A protected setting such as `/abs/python fake_vitest.py` is accepted, then `_run_vitest()` executes it from `cwd=root`; `fake_vitest.py` can be committed/modified by the candidate and is not part of the Vitest support closure, changed-path quarantine, or command identity digest. My probe changed only that pathless candidate runner between base and head and still got trusted PASS evidence.
  **Suggested fix:** Treat pathless non-flag operands in explicit validator commands as unsafe unless resolved and digest-bound outside the candidate checkout. For interpreter-style command prefixes, require absolute external script/module paths or a protected wrapper executable; add Vitest coverage for `/abs/python fake_vitest.py` and keep the Jest parent regression.

### P2 (non-blocking)
- None.

### Prior xhigh verification
- Verified resolved for the Vitest-specific prior findings: protected-test imports, side-effect imports, parent-directory imports, alias rejection, default candidate `node_modules` rejection, obvious explicit candidate-local `vitest_command` paths, command identity binding for path-like argv, and production `--vitest-command` CLI wiring.
- Still unresolved: the explicit-command trust boundary remains bypassable through pathless interpreter/script operands. The parent Jest instance of the same blocker also remains unresolved.

### Verification
- `git diff --check 2a38d3eb8ca7f2c6d634a40146b29ba875b26c08..6607caf4fcc1c2442ca4c0bb41321310043eec95` -> clean.
- Focused read-only archive suite from current top stack: `pytest -q tests/test_sync_core_runner_jest.py tests/test_sync_core_runner_vitest.py tests/test_sync_core_runner_playwright.py tests/test_sync_core_human_attestation.py tests/test_sync_core_reporting.py` -> `143 passed, 1 warning`.
- Read-only adversarial probe: in a temp repo, base committed `fake_vitest.py`, head changed only `fake_vitest.py`, and `run_obligation(..., RunnerConfig(vitest_command=(sys.executable, "fake_vitest.py")))` returned `PASS :: 1 protected Vitest tests passed` instead of quarantining/rejecting the candidate-mutated validator.

gltanaka:
## Stack update: rebased on #1988 pathless operand fix

Rebased onto updated `codex/issue-1932-jest-adapter` and added Vitest-specific coverage for the same round-5 class.

Commit: `2b04d5b65` (`fix(sync): reject pathless Vitest validator operands`)

What changed:
- `vitest_command` now uses the shared protected explicit-validator argv guard.
- `vitest_command` identity now uses the same resolved byte digest helper as Jest.
- Added regression for `(sys.executable, "fake_vitest.py")` so pathless candidate scripts cannot self-certify.

Evidence:
- Focused Vitest pathless/candidate-local trust tests -> 2 passed.
- Full Vitest suite: `pytest -q tests/test_sync_core_runner_vitest.py` -> 31 passed, 1 warning.
- Targeted pylint with existing large-adapter disables -> 10.00/10.
- `git diff --check` clean.

Pushed head: `2b04d5b6565d0d71d27b360c11a657c81ac959b6`.

gltanaka:
## Codex xhigh adversarial re-review round 6 — PR #1993 @ f59a7179f3fd0a75f8dd6794ef9c53a416d41fce

**Verdict:** MERGE

### P1 (blocking)
- None.

### P2 (non-blocking)
- None.

### Prior xhigh verification
- Verified resolved at this head: protected-test import closure, side-effect imports, parent-directory/index imports, alias rejection, default/explicit candidate-local Vitest command rejection, pathless script operand rejection, command identity binding, and production `pdd validate --vitest-command` wiring.
- Inherited Jest layer from #1988 is also clean at the live base.
- Re-ran the ignored bare-package dependency probe against the current stack: it does **not** produce a trusted Vitest PASS because protected-base collection fails before checked-head execution can use candidate-local ignored `node_modules` content.

### Verification
- Live base/head: `codex/issue-1932-jest-adapter` @ `4ec56a581e6dec75322cb7846af9ba41c178b56a`; head `codex/issue-1932-vitest-adapter` @ `f59a7179f3fd0a75f8dd6794ef9c53a416d41fce`; merge state `CLEAN`.
- Current checks: `auto-heal` SUCCESS, `heal` SUCCESS. No `unit-tests` check is present in the live rollup.
- `git diff --check 4ec56a581e6dec75322cb7846af9ba41c178b56a..f59a7179f3fd0a75f8dd6794ef9c53a416d41fce` -> clean.
- Read-only archive run from exact head: `python -m pytest -q tests/test_sync_core_runner_jest.py tests/test_sync_core_runner_vitest.py tests/test_sync_core_reporting.py` -> `85 passed, 1 warning`.

gltanaka:
## Stack rebuild evidence — foundation 7fcd7edcaf1e80af1cd6a14ee743e4313c0dcc46

Rebuilt from the PR pre-rewrite merge-base by replaying the exact own-commit range in order. Push used an explicit force-with-lease. Post-push checks: clean worktree, HEAD equals origin head, merge-base equals the expected parent HEAD, scoped name diff inspected, git diff --check passed, and python -m compileall -q pdd passed.

Head: 80e2cf345c247ca040fd6d4e7887ecaae88b8822
Parent: a9c58e58ee7e77749607b17b9c46b350b985af50
Own commits: 12; changed files: 4.
No-exclusion sync-core suite: 418 passed, 1 failed. The sole failure is the inherited candidate replay-ledger parent safety check in test_sync_core_cli.py on the local macOS filesystem.

gltanaka:
## Codex adversarial review

**PR:** #1993 @ `470e62af9e19d6eb7b367eb8d2979a6f8e3307c1` against `codex/issue-1932-jest-adapter` @ `b94d8ba6fb20e809d0cad14b2461f102f5d2be6e`
**Verdict:** DON'T MERGE
**Findings:** 10 total (9 P1, 1 P2)

### P1 (blocking)

- [P1] Protected command flags can silently select only part of the obligation — `pdd/sync_core/runner.py:1142`, `pdd/sync_core/runner.py:1359`, `pdd/sync_core/runner.py:1797`
  `_protected_command_error()` skips every flag token, so prefixes containing `--testNamePattern`, `--project`, `--shard`, `--related`, or equivalent controls are accepted (a focused probe with `--testNamePattern=smoke --retry=3 --shard=1/2` returned no error). The same filtered command is then used to derive the protected-base identity set and execute the head, so omitted protected tests are absent from both sides and the reduced subset can produce trusted PASS evidence.
  **Suggested fix:** Restrict `vitest_command` to a launcher/entrypoint-only schema, reject all selection/execution flags, force protected values for every relevant control, and derive collection identities independently of the execution command.

- [P1] Retry and repeat failures collapse to trusted PASS — `pdd/sync_core/runner.py:669`, `pdd/sync_core/runner.py:1286`
  Vitest allows retries in config, CLI, and per-test/suite options, but the adapter neither rejects them nor records attempts. Vitest 4's JSON reporter exposes the final `passed` status while retaining prior errors in `failureMessages`; `_vitest_result()` drops those errors and returns PASS. A native-shaped retry result with `status=passed` and a prior failure was normalized to PASS, contrary to the repository contract that diagnostic retries preserve the first outcome and cannot create trusted pass.
  **Suggested fix:** Reject retry/repeat controls at config, argv, and parsed-source levels, or use a trusted reporter that records every attempt and aggregate the worst outcome as FLAKY/non-pass.

- [P1] Declared product modules are quarantined as validator support — `pdd/sync_core/runner.py:718`, `pdd/sync_core/runner.py:793`, `pdd/sync_core/runner.py:1894`
  The Vitest closure recursively follows all relative imports from protected tests but has no `code_under_test_paths` parameter or exclusion. A normal test importing `../src/product` therefore hashes the product as support; an intended candidate product edit changes the validator digest and protected-path set, yielding ERROR/QUARANTINED instead of allowing unchanged tests to certify changed code. A focused probe confirmed that changing only the imported product changes `vitest_validator_config_digest`.
  **Suggested fix:** Thread the protected product/support role map through digesting and every phase, exclude only explicitly declared product paths from support quarantine, and bind those roles into the signed profile/runner identity.

- [P1] Config, loader, transform, and runtime-resource closure remains incomplete — `pdd/sync_core/runner.py:539`, `pdd/sync_core/runner.py:669`, `pdd/sync_core/runner.py:718`
  The regex recognizes only quoted relative import spellings, while valid template-literal/computed dynamic imports, package `imports` mappings, runtime filesystem loads, snapshots, and similar loaders are unbound. Config analysis binds only `setupFiles` and a `transform` map and does not bind or reject executable controls such as `snapshotEnvironment`, snapshot serializers/resolvers, custom runner/pool/environment/reporters, coverage providers, or relevant package metadata. Probes showed both a changed template-literal import target and a changed `snapshotEnvironment` implementation leave the digest unchanged, allowing candidate-controlled head behavior outside quarantine.
  **Suggested fix:** Use a real JS/TS AST with a positive loader/runtime capability model, bind every supported local config/resource dependency plus governing package metadata transitively, and fail closed on unsupported dynamic/config semantics.

- [P1] Git symlinks and other nonregular closure members are trusted as ordinary source — `pdd/sync_core/runner.py:718`, `pdd/sync_core/git_io.py:7`
  Closure construction hashes blob bytes but never checks Git mode or the materialized file type. A committed `120000` setup/support member is accepted; its blob binds only the target pathname while the external target bytes are what Vitest executes. A focused probe changed an absolute symlink target without changing the validator digest.
  **Suggested fix:** Require Git modes `100644`/`100755` for config, tests, support, resources, and snapshots, reject symlinks/gitlinks, and no-follow verify materialized mode and bytes before and after each phase.

- [P1] The external Vitest toolchain is only partially identity-bound and can change across phases — `pdd/sync_core/runner.py:919`, `pdd/sync_core/runner.py:942`, `pdd/sync_core/runner.py:1817`, `pdd/sync_core/runner.py:2085`
  The command digest hashes only path-like argv members (typically Node and `vitest.mjs`), not Vitest's transitive `node_modules`, loader/plugin packages, lockfile, Node runtime dependency closure, or any toolchain manifest. The base run, head run, and post-execution runner digest reread live external paths independently, so different bytes can execute across phases and the final signature can describe only the last observed partial state.
  **Suggested fix:** Require a protected external toolchain manifest/lock closure, capture one complete identity before execution, enforce it before/after every base/head phase, and sign that immutable identity.

- [P1] Candidate tests inherit protected host capabilities and bypass process/resource supervision — `pdd/sync_core/runner.py:1270`, `pdd/sync_core/runner.py:1369`
  `_vitest_environment()` copies almost the entire parent environment unless a key contains a small secret-marker list; it therefore exposes capabilities such as `PDD_ATTESTATION_SIGNER_COMMAND`/issuer data and other host integration variables that the shared `untrusted_child_environment()` intentionally removes. Vitest also uses plain `subprocess.run`, with no network/filesystem sandbox, process-group reaping, surviving-descendant check, or finite output/resource enforcement, so candidate code can access host resources and leave work running after a reported PASS or timeout.
  **Suggested fix:** Use the shared allowlisted child environment and supervised sandbox, enforce bounded output/memory/process counts, and treat timeout or any surviving descendant as non-pass.

- [P1] Collection/execution trees and the reporter channel are mutable by candidate code — `pdd/sync_core/runner.py:1356`, `pdd/sync_core/runner.py:1915`
  The exact-commit clones are writable, the trusted output pathname is exposed to every test process, and no closure/tree identity is checked after Vitest exits. Candidate product code can rewrite a later protected test/support/snapshot while preserving its identity, use accepted `--update` behavior, or race/replace `results.json`; the adapter then reads the modified result and never detects that phase inputs changed. Checking the original checkout before cloning and rebuilding its snapshot afterward does not protect these ephemeral phase trees.
  **Suggested fix:** Materialize a fresh read-only tree per collection/execution phase, capture and compare full closure modes/bytes and commit identity around each run, force snapshot updates off, and use a result channel candidate workers cannot overwrite.

- [P1] Explicit-command Vitest attestations cannot be consumed by canonical reporting or idempotent finalization — `pdd/commands/sync_core.py:428`, `pdd/sync_core/runner.py:2085`, `pdd/sync_core/finalize.py:187`, `pdd/sync_core/reporting.py:167`
  Vitest validation requires an explicit external command and signs a runner digest containing that command identity. The persisted envelope stores only the resulting digest; `_reusable_result()` and reporting later call `runner_identity_digest()` with default `RunnerConfig()`, producing a different digest and rejecting the freshly issued evidence. A focused probe confirmed explicit/default runner digests differ, so the production CLI path cannot remain canonically VERIFIED or finalize idempotently.
  **Suggested fix:** Persist a protected, typed validator/toolchain descriptor with the evidence and make every receiver reconstruct the same runner identity, or define the signed identity entirely in terms of a protected toolchain ID receivers can resolve.

### P2 (non-blocking)

- [P2] Launch and malformed-output failures can escape normalization and crash validation — `pdd/sync_core/runner.py:1286`, `pdd/sync_core/runner.py:1368`
  Reporter parsing catches only selected exceptions, so valid JSON with a non-object root or malformed nested values raises uncaught `AttributeError`/`TypeError`; process launch catches only timeout, so a missing/non-executable launcher or text-decoding failure also escapes `RunnerExecution`. These should be deterministic ERROR/COLLECTION_ERROR outcomes rather than CLI crashes.
  **Suggested fix:** Validate the complete JSON schema before dereferencing it and normalize `OSError`, decoding errors, and all malformed payload shapes.

### Prior Finding Verification

- Earlier PR findings for ordinary/side-effect/parent-directory imports, alias rejection, candidate-local and pathless command operands, and CLI option reachability remain fixed at this head.
- The new findings are equivalent gaps beyond those prior checks: product-role separation, unsupported loader/runtime paths, immutable modes/toolchain/phases, retries, receiver replay, and process capabilities.

### Verification

- Read all 23 prior PR issue comments; no submitted review bodies or inline review comments exist.
- `PYTHONDONTWRITEBYTECODE=1 conda run --no-capture-output -n pdd python -m pytest -q -p no:cacheprovider tests/test_sync_core_runner_vitest.py tests/test_sync_core_reporting.py` -> `54 passed, 1 warning`.
- Focused read-only probes reproduced retry-to-PASS, unchanged digests for `snapshotEnvironment` and template-literal imports, acceptance of a `120000` closure member with unbound target bytes, product/support conflation, accepted filtering/retry/shard flags, and explicit/default receiver digest mismatch.
- `git diff --check` and worktree cleanliness passed.

gltanaka:
## Implementer response to Codex review

@codex Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4948879639

### P1 #1: Protected command flags can silently select only part of the obligation
Status: **FIXING**
Plan: Replace permissive argv filtering with a launcher-only schema, reject selection/retry/repeat/update/execution controls, and derive protected collection independently.

### P1 #2: Retry and repeat failures collapse to trusted PASS
Status: **FIXING**
Plan: Reject retry/repeat controls in argv, config, and source, validate every reported attempt/error, and preserve every non-pass so prior failures cannot normalize to PASS.

### P1 #3: Declared product modules are quarantined as validator support
Status: **FIXING**
Plan: Thread the declared product/support role map through closure construction, phase verification, and signed identity while excluding only declared product paths from support quarantine.

### P1 #4: Config, loader, transform, and runtime-resource closure remains incomplete
Status: **FIXING**
Plan: Adopt the established tree-sitter AST positive capability model, bind supported static imports/loaders/resources/config/package metadata transitively, and reject unsupported executable or dynamic semantics.

### P1 #5: Git symlinks and other nonregular closure members are trusted as ordinary source
Status: **FIXING**
Plan: Require Git modes 100644/100755 and perform no-follow materialized mode/byte verification before and after each phase.

### P1 #6: External Vitest toolchain is partially bound and mutable across phases
Status: **FIXING**
Plan: Require and validate one complete external Node/Vitest dependency lock closure, capture it once, enforce it around every phase, and sign the immutable identity.

### P1 #7: Candidate tests inherit protected host capabilities and bypass supervision
Status: **FIXING**
Plan: Route execution through shared `untrusted_child_environment()` and `run_supervised()` with finite output/resource/process constraints and fail on timeout or survivors.

### P1 #8: Collection/execution trees and reporter channel are mutable
Status: **FIXING**
Plan: Use fresh read-only phase trees, checker-owned result transport, updates disabled, and post-phase commit/closure/tree validation.

### P1 #9: Explicit-command attestations cannot be reported or finalized canonically
Status: **FIXING**
Plan: Persist a typed protected Vitest toolchain descriptor and reconstruct the exact runner identity in evidence storage, reporting, trust verification, and finalization.

### P2 #1: Launch and malformed-output failures can crash validation
Status: **FIXING**
Plan: Strictly validate the complete result schema and normalize launch, decoding, timeout, malformed-root, and malformed-nested failures into deterministic runner outcomes.

### Sibling and prior-commit investigation
- Sibling occurrences: the current Vitest path duplicates earlier permissive Playwright/Jest-era behavior in command parsing, regex source analysis, live toolchain rereads, writable clones, environment inheritance, and receiver-side default `RunnerConfig()` reconstruction. All Vitest occurrences are in scope; unrelated adapters will not be changed.
- Shared primitives found: `pdd.sync_core.isolation.untrusted_child_environment`, `pdd.sync_core.supervisor.run_supervised`, Git regular-mode/no-follow helpers, protected evidence descriptors, and runner identity reconstruction.
- Prior fixes examined: `e49685384` (AST and immutable Playwright toolchain closure), `782eade41` (typed protected descriptor and receiver replay), `20adbd58e` (phase/tree integrity), `df8f8b39f` (isolated trust descriptors), `454414367` (dependency environment binding), and `fb34318c3` (retain trust checks around isolation).
- Convention: use tree-sitter positive grammars, checker-owned immutable descriptors, shared environment/supervisor boundaries, one pre-captured toolchain closure, and fail-closed phase checks. The Vitest implementation will follow that shape rather than introduce parallel weaker mechanisms.
- TDD: genuine regressions for all ten findings will be committed and pushed as a separate red `test(...)` commit before production changes; red output will be posted here.

gltanaka:
## Red TDD evidence — complete Vitest trust boundary

Pushed separate test commit `30f6021e3` before production changes.

Command:
`PYTHONDONTWRITEBYTECODE=1 conda run --no-capture-output -n pdd python -m pytest -q -p no:cacheprovider tests/test_sync_core_runner_vitest.py`

Red result: `11 failed` in the newly added trust regressions (with existing adapter tests continuing through the run). A focused `-x` rerun captured the first concrete regression:

`test_vitest_prior_retry_failure_cannot_normalize_to_pass` expected a prior failed retry attempt to remain non-pass, but `_vitest_result()` returned `EvidenceOutcome.PASS`.

The failing set covers launcher-only command controls, retry history, product/support roles, AST loader/config closure, nonregular Git members, protected child environment, shared supervision, typed toolchain identity, immutable phase trees, malformed payload shapes, and launch-error normalization.

gltanaka:
## Green implementer evidence — Vitest trust boundary complete

Pushed production commit `91ac82f0b362a1c8861eacf8ab8a5a35fea202ce` after the separate red test commit `30f6021e3`.

### Finding resolution
1. **Launcher-only schema:** Vitest accepts exactly an external launcher and entrypoint; selection, shard, related, retry/repeat, update, and interpreter controls are rejected.
2. **Attempts/non-pass:** config retry/repeat controls are rejected and prior `failureMessages` force non-pass even when the final status says passed.
3. **Product/support roles:** declared `code_under_test_paths` are threaded through digesting, phase closure, and signed runner identity; only declared product edges are excluded from support quarantine.
4. **AST/config/resources:** tree-sitter positively resolves supported static import/export/import()/require() forms, binds package metadata and transitive resources, and rejects dynamic loaders plus unsupported executable config controls.
5. **Git/materialized modes:** closure reads require Git `100644`/`100755`; symlinks/gitlinks are rejected and phase trees are mode/byte checked without following symlinks.
6. **Toolchain closure:** a typed manifest binds launcher, entrypoint, complete dependencies, and lockfile once; the identity is rechecked around every phase and signed.
7. **Environment/supervision:** execution uses `untrusted_child_environment()` and `run_supervised()`, bounded sandbox output/process handling, timeout normalization, and surviving-descendant rejection.
8. **Immutable phases/channel:** base collection and head execution use fresh exact-commit read-only trees; only checker cache/result roots are writable, updates are rejected, and post-phase tree closure is validated.
9. **Canonical replay:** command, manifest, and immutable toolchain identity are signed, persisted, strictly decoded, and reconstructed by reporting/idempotent finalization.
10. **Failure normalization:** missing/non-executable launchers, unsupported sandbox launches, decode failures, non-object roots, malformed nested results, and missing reporter output normalize to deterministic non-pass outcomes.

### Verification
- Trust/cross-adapter suite: `152 passed`
  - `tests/test_sync_core_runner_vitest.py`
  - `tests/test_sync_core_runner_jest.py`
  - `tests/test_sync_core_runner.py`
  - `tests/test_sync_core_reporting.py`
  - `tests/test_sync_core_evidence_store.py`
  - `tests/test_sync_core_supervisor.py`
- Final Vitest/reporting/evidence rerun: `75 passed`
- Final receiver-focused rerun: `25 passed`
- Production pylint: `10.00/10`
- `python -m compileall -q pdd tests/test_sync_core_runner_vitest.py`: passed
- `git diff --check`: passed
- Real Node/Vitest E2E using a complete external manifest and a product-importing TypeScript test: `PASS 1 protected Vitest tests passed`
- Branch is clean and pushed; local and remote head both equal `91ac82f0b362a1c8861eacf8ab8a5a35fea202ce`.

gltanaka:
## Codex adversarial review — PR #1993 @ `86b361a864270d722ad225ee372526d35328235d`

**Verdict:** DON'T MERGE
**Findings:** 8 total (6 P1, 2 P2)

### P1 (blocking)

- [P1] Executed Vitest bytes are not bound to the declared toolchain — `pdd/commands/sync_core.py:161-186`, `pdd/sync_core/runner.py:1209-1300`, `pdd/sync_core/runner.py:1601-1611`
  **Impact:** A protected run can execute launcher/entrypoint bytes outside the signed manifest closure, and the merged duplicate helper silently weakens mode/symlink/cache integrity. Evidence can therefore remain identity-valid after the executed entrypoint changes.
  **Evidence:** The CLI accepts `--vitest-command` without a manifest (and `tests/test_sync_core_reporting.py:577-611` asserts that behavior). With a manifest, no code checks that `vitest_command` equals its `launcher`/`entrypoint`; once `vitest_toolchain_identity` is populated, `_validator_command_identity_digest` records only literal argv. A focused probe at this exact head used a manifest for toolchain A and command B: `_protected_command_error` accepted it, and mutating B's entrypoint left the command identity unchanged. In addition, the mode/symlink-aware `_directory_identity` at line 1209 is overridden at runtime by the inherited Jest definition at line 1601; the effective function ignored a file mode change and a symlink-directory target-byte change, and included `.vite` cache writes that the earlier definition intended to exclude.
  **Smallest correction:** Require command and complete manifest together, parse one typed descriptor, require exact command-to-manifest launcher/entrypoint equality, capture immutable bytes once, and use one uniquely named/consolidated no-follow mode-sensitive directory identity implementation.

- [P1] The Linux sandbox cannot see the declared external Vitest entrypoint — `pdd/sync_core/runner.py:1939-1998`
  **Impact:** The production path can fail to launch real Node/Vitest under the only supported protected sandbox, so the claimed real E2E is not represented by the final code path.
  **Evidence:** `_prepare_vitest_dependencies` copies dependencies into the phase tree, but `_run_vitest` still executes the original absolute external `command_prefix[1]`. Its `run_supervised` call supplies no `readable_roots`; `supervisor._runtime_roots` exposes `command[0]` and the phase cwd, not the external entrypoint (or Node's complete native closure). The Jest path explicitly supplies its entrypoint/toolchain roots, while the new Vitest path does not.
  **Smallest correction:** Execute the identity-bound copied entrypoint from the immutable phase tree, or explicitly expose the exact manifest-bound launcher/entrypoint/native closure as read-only sandbox roots; add a real Linux `run_supervised` Vitest test.

- [P1] Candidate code can write the authoritative Vitest result channel — `pdd/sync_core/runner.py:1972-2017`
  **Impact:** Candidate-authored tests/application code can forge the JSON identities and statuses used to issue PASS evidence.
  **Evidence:** `results.json` is inside `temporary`, all of `temporary` is passed as a writable sandbox root, and its path is disclosed both in argv and `PDD_TRUSTED_VITEST_OUTPUT`. `_vitest_result` then trusts that file. The test fixture itself demonstrates the capability by writing arbitrary result JSON through this environment variable; there is no privilege or IPC boundary separating candidate workers from the result sink.
  **Smallest correction:** Move result transport behind a checker-owned channel inaccessible to candidate workers (for example, a protected reporter/controller IPC boundary), then add a regression where test code attempts to overwrite/forge the sink.

- [P1] Supported Node/Vite loader forms bypass the immutable support closure — `pdd/sync_core/runner.py:874-923`
  **Impact:** A protected test can execute mutable local helper code that is absent from the closure digest and quarantine set.
  **Evidence:** The AST logic only recognizes call functions whose bytes are exactly `import` or `require`. At this head, direct `require(variable)` was rejected, but `module.require(variable)` and `import.meta.glob('./helpers/*.ts')` were accepted and returned an empty import set. Both are executable local-loading forms in the supported JS/Vite ecosystem.
  **Smallest correction:** Positively bind supported alternate loaders/globs or reject them fail-closed, including `module.require`, `createRequire` aliases, and `import.meta.glob`; add direct and transitive mutation regressions.

- [P1] Vitest hardening regresses the merged Jest command contract — `pdd/sync_core/runner.py:1642-1678`
  **Impact:** Existing trusted Jest configurations that need launcher flags (for example Node's `--experimental-vm-modules`) now fail before Jest starts.
  **Evidence:** Relative to exact merged Jest base `e37b0bf4`, this PR changes the shared `_protected_command_error` branch for every `-` operand from `continue` to an error. Both Jest and Vitest CLI parsing use that shared helper, but only Vitest requires a two-operand launcher schema. No Jest regression test covers launcher flags.
  **Smallest correction:** Preserve the merged Jest parser/contract and apply the exact two-operand/no-control schema in a Vitest-specific validator.

- [P1] Exact-head required CI is red and unit CI is absent — `.github/workflows/auto-heal.yml:38`, `.github/workflows/unit-tests.yml:23`
  **Impact:** This SHA has no passing merge-gate evidence and must not merge.
  **Evidence:** Exact-head `heal` and `auto-heal` both failed in run `29223611103` / Cloud Build `71531c5e-2e26-4c47-8b28-8346deda9600`; the commit has no `unit-tests` check run at all. `gh pr checks 1993` reports both available checks failed.
  **Smallest correction:** Fix the code findings, push a new exact head to trigger `pull_request.synchronize`, inspect the Cloud Build failure, and require both unit and auto-heal checks to pass on that same SHA.

### P2 (non-blocking robustness/test defects)

- [P2] Toolchain setup and post-phase races can escape outcome normalization — `pdd/sync_core/runner.py:2012-2015`, `pdd/sync_core/runner.py:2480-2542`
  **Impact:** Manifest/dependency disappearance, permission errors, malformed setup data, or copy failures can raise out of `run_profile` instead of producing deterministic `RunnerExecution(ERROR, ...)`.
  **Evidence:** `_prepare_vitest_dependencies` is called outside a normalization guard in both phase constructors, and the post-phase `_vitest_toolchain_identity` call is outside the earlier `try`. Both perform filesystem operations that raise `OSError`/`ValueError`.
  **Smallest correction:** Wrap phase preparation and every identity recheck, normalize filesystem/JSON/copy errors to a Vitest ERROR execution, and add deletion/permission/copy-failure tests.

- [P2] The new Vitest suite is not portable to the repository's fail-closed macOS policy — `tests/test_sync_core_runner_vitest.py:303-376`
  **Impact:** The advertised focused suite fails for macOS contributors even though unsupported sandbox execution is intentionally fail-closed.
  **Evidence:** On this exact head, the requested combined target produced `171 passed, 46 skipped, 13 failed`; nine failures were Vitest assertions expecting PASS/SKIP/TIMEOUT/etc. but receiving `unsupported protected sandbox: macOS cannot prove process lifetime isolation`. Jest has an autouse local subprocess fixture and the shared runner tests explicitly skip protected execution on Darwin; Vitest has neither.
  **Smallest correction:** Give Vitest unit tests a controlled supervisor fixture like Jest, and keep a separate Linux-only real-sandbox E2E; otherwise explicitly skip execution-outcome cases on unsupported platforms.

### Notes

- Fresh `origin/main` is exactly merged Jest SHA `e37b0bf4e21c8aa176ec3990554b7dc888a873d7`; it is both merge-base and ancestor of this exact head.
- Historical-to-current `git range-diff` maps all 14 pre-merge Vitest commits in order. The only additional commit is `86b361a` (restoring dependency-copy setup after the rebase), but final semantic inspection found the duplicate helper override above.
- Protected-base profile ownership, candidate-removal/change rejection, reporting, and denominator/inventory checks passed in the focused integration slice (`16 passed, 36 deselected`). Direct npx/npm-style launcher indirection is rejected, and the Vitest child environment uses the shared credential-reducing supervisor path.
- Worktree remained clean and local/remote head stayed exactly `86b361a864270d722ad225ee372526d35328235d`.

gltanaka:
## Implementer response to Sol review

@sol Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4954477898

### P1 #1: Executed Vitest bytes are not bound to the declared toolchain
Status: **FIXING**
Plan: Introduce one typed complete descriptor, require exact command/manifest launcher-entrypoint equality, bind immutable launcher/entrypoint/native/dependency/lockfile bytes, and consolidate directory identity into one uniquely named no-follow cache-excluding implementation.

### P1 #2: Linux sandbox cannot see the external Vitest entrypoint
Status: **FIXING**
Plan: Execute a descriptor-bound copied entrypoint with the exact read-only native closure available to the supervised sandbox, covered by a Linux-only real supervisor test.

### P1 #3: Candidate code can forge authoritative Vitest results
Status: **FIXING**
Plan: Replace candidate-writable result transport with a checker-owned controller channel and add a candidate forging regression.

### P1 #4: Alternate Node/Vite loaders bypass support closure
Status: **FIXING**
Plan: Fail closed on unbound module.require, createRequire aliases, and import.meta.glob loader forms, including direct and transitive regressions.

### P1 #5: Vitest hardening regresses Jest command flags
Status: **FIXING**
Plan: Restore the merged shared Jest launcher-flag contract and move exact two-operand/no-control validation into a Vitest-specific validator, with Jest regression coverage.

### P1 #6: Exact-head required CI is red and unit CI absent
Status: **FIXING**
Plan: Push a corrected exact head, diagnose the Cloud Build failure, trigger unit and auto-heal, and require both checks to pass on that SHA.

### P2 #1: Toolchain setup and post-phase races escape normalization
Status: **FIXING**
Plan: Normalize descriptor, copy, permission, deletion, and identity-recheck failures to deterministic Vitest ERROR executions and cover each boundary.

### P2 #2: Vitest suite is not portable on macOS
Status: **FIXING**
Plan: Add a controlled supervisor unit fixture for Darwin-portable tests and retain a separate Linux-only real sandbox E2E.

gltanaka:
## Sol correction cycle — failing regression evidence

Exact failing-test commit: `ab4596360`

Command:
`conda run -n pdd pytest -q tests/test_sync_core_runner_vitest.py tests/test_sync_core_runner_jest.py tests/test_sync_core_reporting.py tests/test_sync_core_runner_vitest_linux.py`

Expected red result: collection fails because `_load_vitest_toolchain_descriptor` does not yet exist. This demonstrates the new tests require the typed descriptor and related trust-boundary implementation before they can pass.

gltanaka:
## Exact-head auto-heal diagnosis — `348ebc53dc1f5becd78f8d73f30deebc7365b8de`

- Workflow: https://github.com/promptdriven/pdd/actions/runs/29228510776
- Cloud Build: `439e9a59-f43a-45ec-9a79-9ce2dc85fb23`
- Result: `UNBASELINED` only for `checkup_prompt_main` and `sync_main`; 0 conflicts, 0 failures, 0 stale metadata.

This is an external trusted-detector false positive, not branch drift. The pdd_cloud detector's reverse-dependency predicate accepts basename/suffix matches:

- changed `pdd/sync_core/evidence_store.py` (`evidence_store.py`) falsely matches unchanged include `pdd/evidence_store.py`, selecting `checkup_prompt_main`;
- changed `pdd/sync_core/runner.py` (`runner.py`) falsely matches unchanged include `pdd/agentic_sync_runner.py`, selecting `sync_main`.

Reproduction against the exact PR diff:

```text
python pdd_cloud/scripts/ci/detect_pdd_cli_auto_heal_modules.py --diff-base origin/main...HEAD
# includes checkup_prompt_main and sync_main

python -m pdd.ci_drift_heal --dry-run --json --budget-cap 5.00 \
  --diff-base origin/main...HEAD --modules <detector output>
# ok=false; unbaselined=2; checkup_prompt_main, sync_main
```

No branch correction is appropriate: feature-local baselines or prompt edits for these protected, unchanged units would violate protected-base ownership. The detector must compare normalized repo paths rather than leaf basenames in pdd_cloud.

gltanaka:
## Sol exact-head correction review — PR #1993 @ `ec032368ace6b4f5d7bf3d730fef567242d47bfe`

**Verdict: DON'T MERGE**

Two P1 code findings remain. Five of the original eight findings are resolved; the exact-head CI finding is still pending/separately blocked by the acknowledged auto-heal basename collision.

### Remaining findings

- **[P1] Executed dependency bytes can differ from the evidence-bound toolchain identity** — `pdd/sync_core/runner.py:1386-1397`, `pdd/sync_core/runner.py:1401-1445`
  - **Impact:** PASS evidence can identify one Node/Vitest dependency closure while the phase executes different copied bytes.
  - **Evidence:** The descriptor computes and stores `dependencies_identity`, but no production path consumes it. `_prepare_vitest_toolchain` copies the source tree, checks only the copied launcher, then re-hashes the original external source. An adversarial probe changed `dep.js` only during `_copy_vitest_dependencies`, restored the source before the descriptor reload, and `_prepare_vitest_toolchain` accepted a phase whose copied `dep.js` contained `attacker`. Thus the source recheck passed while the executed copy was not the measured closure. The same design leaves the native runtime as mutable original paths rather than a captured immutable phase closure.
  - **Smallest correction:** Capture/stage the complete executable closure once, verify a symlink/mode/byte-sensitive identity of the exact staged representation against the descriptor, expose/execute only that immutable staged launcher/dependencies/native closure, and bind evidence to that identity. Add a mutation-during-copy regression.

- **[P1] A supported aliased CommonJS loader still bypasses the immutable support closure** — `pdd/sync_core/runner.py:902-972`
  - **Impact:** A candidate can change executable local test-support code without changing `vitest_validator_config_digest`, weakening protected-base tests while retaining the approved profile digest.
  - **Evidence:** The AST rejects direct/member loader spellings, but accepts `const load = require; load('./helper.cjs')`. Against this exact head, changing only `helper.cjs` across commits produced the same digest (`371d4fbe...` both times). A separate Vitest **4.1.10** probe with `test.globals=true` confirmed this exact alias form executes successfully and the test passes.
  - **Smallest correction:** Track/bind loader aliases or reject assignment/destructuring/indirection from `require` fail-closed, including a transitive digest-mutation regression using executable `.cjs`.

### Original-finding verification

- Typed descriptor schema, exact command/manifest launcher-entrypoint equality, direct executable resolution, mode/byte/symlink-sensitive source identity, and no duplicate top-level helper: **resolved**, except for the staged-byte gap above.
- Copied entrypoint visibility in the Linux sandbox: **resolved**. Exact-head CI's provisioned real-Vitest focused runner step passed.
- Private FIFO result transport: **resolved**. The FIFO is opened/unlinked before candidate execution; the pathname cannot replace the open channel. A Vitest 4.1.10 worker probe could not access inherited FD 198.
- Alternate loaders: **not fully resolved**; aliased `require` remains open as above.
- Jest launcher flags: **resolved**; shared parsing preserves flags and the focused Jest regression passes.
- Setup and post-phase descriptor/recheck failures: **resolved** to deterministic `ERROR` outcomes for the reported boundaries.
- macOS fixture/Linux real-E2E separation: **resolved** for Vitest. The combined local slice was `220 passed, 16 skipped`; four reporting finalizer tests failed only because the repository intentionally rejects protected pytest execution on Darwin.
- Liveness/Alias integration and protected ownership/denominator behavior: **retained**. Alias, manifest, profile, reporting, Jest/Vitest, and supervisor coverage passed apart from those Darwin-only finalizer executions. Range-diff retains the unique Vitest series atop current main; no inherited duplicate Jest helper remains.

### Exact-head CI at review time

- Exact head/base confirmed: `ec032368ace6b4f5d7bf3d730fef567242d47bfe` / `66b75a839186427ef4e1f18c5559f99e2caba183`; base is the merge-base and ancestor; worktree remained clean.
- Linux `Run focused protected-runner tests`, protected pytest/signer smokes, Public CLI Regression, Story Regression, Package Preprocess Smoke, Repo Bloat Docker E2E, and CodeQL: **passed**.
- Broad `Run Unit Tests`: **still in progress** at posting time.
- `heal` / `auto-heal`: **failed** with the separately tracked basename-collision detector issue; not counted as a Vitest code finding here.

gltanaka:
## Implementer response to Sol exact-head correction review

Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4955521945

### P1 #1: Executed dependency bytes can differ from evidence-bound identity
Status: **FIXING**
Plan: derive an immutable expected member manifest from the captured typed descriptor and verify the exact staged launcher, entrypoint, dependency, lock, and native closure before every collection/execution phase, rejecting byte/mode/type/link/membership drift and extras.

### P1 #2: Aliased CommonJS loader bypass
Status: **FIXING**
Plan: positively track statically provable `require` and `createRequire` aliases through supported declarations/assignments and transitive local modules, while rejecting dynamic, reassigned, ambiguous, or arbitrary member-call loaders fail-closed.

The approved Vitest/Liveness/Alias/Jest behavior and protected-base ownership remain non-goals for modification.

gltanaka:
## Red TDD evidence — remaining Vitest P1s

Test-only head: `e18f68ba2`

```text
19 failed, 2 passed, 67 deselected
```

The failures reproduce:
- dependency mutation during copy;
- staged dependency byte/mode/symlink/missing/extra drift;
- staged launcher, lockfile, and native-runtime mutation;
- direct `require` aliases, named/renamed/destructured `createRequire`, assignment aliases, and transitive CommonJS aliases not binding helper bytes;
- dynamic reassignment, ambiguous provenance, escaped aliases, and arbitrary member-call loaders not failing closed.

The two passing controls are source mutation already detected by the external descriptor recheck and the already-rejected dynamic `createRequire` operand. Implementation follows in a separate fix commit.

gltanaka:
## Codex critical correction review — PR #1993 @ d2468f46b

**Verdict: DON'T MERGE**

**Counts:** P1: 2 · P2: 1

1. **[P1] Fail closed when the initial toolchain capture fails** — `pdd/sync_core/runner.py:3533-3543`. The initial descriptor exception is swallowed. If the external source/manifest is transiently invalid or unreadable at that point but becomes readable before `run_obligation`, that obligation can execute successfully against its own locally captured identity while the outer `config` remains unbound; `runner_identity_digest` and the signed binding can then be derived from a later, different source state (and `vitest_toolchain_identity` remains `None`). Do not start Vitest obligations after a failed initial capture; retain one captured descriptor identity in the profile config and use it for every execution, final digest, and binding.

2. **[P1] Reject indirect and destructured CommonJS loader provenance** — `pdd/sync_core/runner.py:1083-1114`. The second pass only treats a member/unknown call as ambiguous when an argument is itself a static local string. Valid CommonJS such as `const p='./helper.cjs'; require.apply(null,[p])`, `Reflect.apply(require,null,[p])`, or `const [load=require]=[]; load(p)` is accepted with an empty closure, so a changed transitive helper can influence/pass the test without being protected. Track loader provenance through destructuring/all assignment forms and indirect calls, and reject any unresolved dynamic invocation that can carry `require`/`createRequire` provenance.

3. **[P2] Make JavaScript parsers deterministic and offline** — `pdd/sync_core/runner.py:931`. A fresh parser lookup can attempt a network fetch; in a network-restricted fresh environment this raised an unnormalized parser download error. Normalize parser initialization failures to fail-closed Vitest evidence; bundled grammar distribution can be handled separately if the dependency cannot supply installed artifacts.

gltanaka:
## Implementer response to critical correction review

@codex Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4957652058

### P1 #1: failed initial descriptor capture
Status: **FIXING**
Plan: make profile-level Vitest capture fail closed and retain its exact identity for every obligation, digest, and signed binding.

### P1 #2: indirect/destructured CommonJS provenance
Status: **FIXING**
Plan: reject unproven loader provenance in patterns and indirect member/apply calls, with regressions for every demonstrated bypass.

### P2 #1: offline parser initialization
Status: **FIXING normalization; ACCEPTING grammar packaging as follow-up risk**
Plan: normalize parser initialization failures to deterministic fail-closed validation; dependency-level offline grammar bundling is outside the two requested P1s and does not change accepted loader semantics.

gltanaka:
## Codex correction verification — PR #1993 @ 5e93cdfe6

**DON'T MERGE**

- **[P1] Fail closed when loader provenance passes through aliases or helpers.** The new check in `pdd/sync_core/runner.py:1106-1142` only rejects `.apply`/`.call`/`.bind` when known provenance is syntactically present in that member call. Valid support code such as `const p='./helper.cjs'; function invoke(loader,arg){ return loader(arg) } invoke(require,p);`, `const {apply}=Reflect; apply(require,null,[p]);`, or `const load=module.constructor._load; load(p,module);` is still accepted; the same bypass works in a transitive imported helper, and changing `helper.cjs` leaves `vitest_validator_config_digest` unchanged even though Vitest executes it. Destructured `Reflect.apply` and `_load` variants fail similarly, so the provenance check must fail closed beyond direct member calls.

gltanaka:
## Sol exact-head review — PR #1993 @ `5e93cdfe6e0298777bc05b4a26346e7dde528f56`

**Verdict: DON'T MERGE** — 1 P1, 1 P2.

### P1

- **[P1] Loader provenance can still escape through expression statements and internal-loader aliases** — `pdd/sync_core/runner.py:996-1155`
  - **Impact:** Executable local test-support bytes can change without changing `vitest_validator_config_digest`, allowing protected tests to be weakened while retaining the approved profile digest.
  - **Evidence:** The current checks find provenance mainly in assignments and syntactically direct calls. At this exact head, each of these was accepted and changing `helper.cjs` left the digest unchanged:
    - `const p='./helper.cjs'; function invoke(loader,arg){ return loader(arg) } invoke(require,p);`
    - `const p='./helper.cjs'; const {apply}=Reflect; apply(require,null,[p]);`
    - `const p='./helper.cjs'; (0,require)(p);`
    - the helper-parameter form inside a transitively imported `loader.cjs`;
    - `import Module from 'node:module'; const p=process.cwd()+'/helper.cjs'; const load=Module._load; load(p);`
  - Vitest **4.1.10** executed and passed the helper-parameter, destructured-`Reflect.apply`, comma-`require`, and `Module._load` forms. Direct static `require`, simple direct aliases, and named/renamed/destructured `createRequire(import.meta.url)` aliases were correctly bound; direct dynamic/reassigned forms were correctly rejected. Existing tests cover direct aliases and direct `module.constructor._load`, but not these escaping expression/transitive forms.
  - **Smallest coherent correction:** Implement a lexical def-use rule for loader provenance: allow a proven `require`/`createRequire` alias only when every use is another simple alias definition or a direct call with one static local literal/template; reject passing, returning, storing, destructuring, sequence/computed/member use, reassignment, reflective invocation, and all `Module._load`/equivalent internal-loader references regardless of how the path is constructed. Add direct and transitive mutation regressions for each escape class.

### P2

- **[P2] JavaScript grammars are fail-closed but not provisioned offline** — `pyproject.toml:80`, `pdd/sync_core/runner.py:931-938`
  - Parser download/initialization failures now normalize deterministically to `ValueError`, so no unsupported source is accepted. However, `tree-sitter-language-pack` stores JavaScript/TypeScript grammars in a per-user cache and its installed wheel exposes runtime download APIs; the wheel itself did not contain those grammar libraries. A fresh offline install therefore disables the advertised Vitest adapter instead of parsing locally.
  - **Smallest coherent correction:** Ship deterministic JavaScript and TypeScript grammar artifacts as install dependencies/assets (for example, packaged `tree-sitter-javascript` and `tree-sitter-typescript` languages), and add an empty-cache/network-denied installation test.

### Resolved copied-closure P1

The descriptor-bound copied closure is now correctly checked. The initial descriptor identity is retained in `RunnerConfig` and the signed attestation binding; each obligation reload must match that identity. `_prepare_vitest_toolchain` compares the staged dependency membership and every copied launcher, entrypoint, lockfile, and native-runtime member to the descriptor's captured records before supervision. Probes/tests reject source or copy-time mutation, bytes, modes, kinds/symlinks, missing paths, extra paths, launcher, lockfile, and native-runtime drift. Internal symlinks preserve link spelling and escaping links fail closed.

### Validation / CI

- Exact base/head: `76403bd758d0c712b6c809bab5e5af5b98ac9a1d` / `5e93cdfe6e0298777bc05b4a26346e7dde528f56`; base is the merge-base and ancestor; remote head matches; worktree remained clean.
- Local focused Jest/Vitest/supervisor: **153 passed, 14 skipped**.
- Local Alias/Liveness/profile/manifest/fingerprint/global-dry-run slice: **300 passed**.
- Exact-head Linux CI: protected sandbox provisioning, real-Vitest focused protected-runner tests, pytest/signer smokes, architecture checks, Public CLI Regression, Story Regression, Package Preprocess Smoke, Repo Bloat Docker E2E, and CodeQL **passed**.
- Broad `Run Unit Tests` was still in progress at posting time. Exact-head `heal` and `auto-heal` were **failed**, so merge-gate evidence is not green independently of the P1.

gltanaka:
## Implementer response to Sol exact-head review

Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4957895049

### P1: Loader provenance can escape through expression statements and internal-loader aliases
Status: **FIXING**
Plan: replace spelling-oriented checks with a lexical positive capability model that permits only simple proven aliases and direct static-local calls, rejecting all capability forwarding, reflective/sequence/member invocation, reassignment/shadowing, and internal Module loaders; add direct/transitive binding and rejection regressions.

### P2: JavaScript grammars are fail-closed but not provisioned offline
Status: **FIXING**
Plan: pin and package the JavaScript/TypeScript grammar distributions required by the protected parser, use those installed grammars directly, and add a clean built-wheel no-network/empty-cache smoke while retaining deterministic fail-closed errors.

gltanaka:
## Critical correction review — PR #1993 @ `d5de4d768939d690857fe4b574e89ef56206db8d`

**Verdict: DON'T MERGE** — 1 P1 found in the bounded P1/P2 review.

### P1

- **[P1] Ambient loader recovery still escapes the positive capability model** — `pdd/sync_core/runner.py:996-1221`
  Escaped identifiers (`requ\\u0069re`, escaped `createRequire`), `eval`/`Function` recovery, reflective access through aliases of `module`, and computed/reflective reconstruction of `_load` are accepted without binding the executed local module. Direct static and simple alias forms remain correctly bound. The analyzer must reject identifier escapes and all ambient/reflective/internal loader-recovery capabilities unless they are one of the explicitly supported direct static forms.

### Verification

- Focused loader/grammar tests: `38 passed`.
- Independent probes reproduced the unbound cases above.
- P2 wheel/offline implementation had no finding before the reviewer session was interrupted; it will be reverified after correction.

gltanaka:
## Implementer response to critical correction review

Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4958087803

### P1: Ambient loader recovery escapes
Status: **FIXING**
Plan: extend the positive model to reject escaped identifiers, dynamic code constructors, reflection/call indirection, internal module namespace aliases, and computed callable extraction; retain only the explicitly supported direct static import/require/createRequire capability path and add exact regression probes.

gltanaka:
## Implementer correction evidence — PR #1993 @ `22c90eb35ac487d89cd8f3c82bf6eceece7dbdb1`

**DON'T MERGE automatically; maintainer review remains required.**

### Corrected

- Positive loader-capability analysis rejects helper forwarding, reflection/call indirection, sequence calls, internal Module loaders, shadowing/reassignment, escaped identifiers, dynamic code constructors, ambient module recovery, and computed callable extraction.
- Supported direct static imports/requires and proven `createRequire(import.meta.url)` aliases remain bound, including transitive helper mutation.
- Protected parsing now imports exact pinned wheel-provisioned JavaScript/TypeScript grammars and fails deterministically on missing/version-mismatched packages.
- Package smoke builds a wheel, materializes dependencies, installs with `PIP_NO_INDEX=1`, denies sockets before importing `pdd`, and parses JavaScript/TypeScript offline.
- Offline probe stays inline in the protected workflow; no feature-local ownership path remains.

### Validation

- Vitest/Jest/supervisor: **183 passed, 14 skipped**.
- Trust/reporting/evidence/Liveness/Alias/rollout/CLI: **268 passed, 57 skipped, 4 known macOS sandbox tests deselected**.
- Fresh built-wheel no-index grammar smoke: **PASS**; exact versions `tree-sitter==0.25.2`, `tree-sitter-javascript==0.25.0`, `tree-sitter-typescript==0.23.2`.
- `pylint pdd/sync_core/runner.py`: **10.00/10**.
- `git diff --check`: **clean**; worktree clean; remote head exact.

### Auto-heal diagnosis

The prior exact-head workflow reached Cloud Build successfully; detector/authorization/submission passed, then Cloud Build `67438978-8957-4207-8560-0e9840324879` returned `FAILURE`. Detailed Cloud logs are unavailable locally because `gcloud builds describe` requires interactive reauthentication. This exact head triggered new Unit Tests run `29251673505` and auto-heal run `29251671405`; per instruction, implementation did not wait on them.

### Review note

The critical reviewer found the ambient-recovery P1 and its committed probes now pass. Same-session verification was attempted but Codex reported a hard usage limit until July 19, so no independent final MERGE verdict is claimed.

gltanaka:
## Implementer response to exact-head correction findings

Re: requested correction cycle at `22c90eb35ac487d89cd8f3c82bf6eceece7dbdb1`

### P1 #1: package exports/imports local mapping closure
Status: **FIXING**
Plan: resolve only deterministic static local export/import mappings to exact local files, traverse their supported closures, and bind their content identity into support quarantine.

### P1 #2: persisted adapter identity
Status: **FIXING**
Plan: retain path-independent content identities for every configured adapter and have issuance, reporting, trust, and finalization verify signed identities without reopening temporary manifests.

### P1 #3: FIFO result transport
Status: **FIXING**
Plan: drain the sandbox result FIFO concurrently under a strict byte cap with deterministic writer/EOF/overflow outcomes before any candidate result access.

### P1 #4: protected Ubuntu Vitest invocation
Status: **FIXING**
Plan: invoke the copied Vitest entrypoint through protected Node with `--disable-wasm-trap-handler`, bind that command in evidence identity, and retain the 2 GiB address-space and existing limits.

gltanaka:
## Sol high review — exact head `23618b6a88f62429d7336333becac6e5193a4ba1`

**Verdict: DO NOT MERGE**

A second substantial correction cycle is required, so the bounded review loop stops here. Remaining findings:

- **P1:** package-map resolution uses the repository-root `package.json` instead of the nearest committed package scope. Nested `imports` and self-exports can bind the wrong helper.
- **P1:** adapter identity capture failures are caught, then `runner_identity_digest` recaptures and raises. Missing, non-executable, pathless, candidate-local, and indirect launcher cases must return normalized ERROR evidence without a second live capture.
- **P2:** persisted Vitest identity still hashes raw manifest bytes and absolute role paths, so identical relocated toolchains produce different identities.
- **P2:** FIFO empty-result handling runs before return-code handling and masks launcher 126/127/nonzero failures as collection errors.

Confirmed fixed: concurrent capped FIFO draining, `--disable-wasm-trap-handler` in executed and digested argv, and existing 2 GiB/CPU/process/output limits.

Local evidence: focused regressions 11 passed/1 skipped; full Jest/Vitest 167 passed/1 skipped/10 failed; worktree clean and exact SHA verified.

gltanaka:
## Implementer response to Sol review

Re: https://github.com/promptdriven/pdd/pull/1993#issuecomment-4959894422

- P1 package scopes: **FIXING** — resolve each static import and package self-export from its nearest committed package manifest, failing closed for unsupported mappings.
- P1 identity capture: **FIXING** — normalize a single capture result and pass it through digesting without a second live capture.
- P2 persisted Vitest identity: **FIXING** — canonicalize identity to typed roles, member-relative names, bytes, modes, and launch policy.
- P2 FIFO outcome precedence: **FIXING** — evaluate timeout/survivor/exit status before a successful no-writer collection error.

I will add focused regression coverage before implementing each behavior, preserve the approved branch work, and push the resulting commits without merging.

gltanaka:
## Sol high narrow verification — PR #1993 @ `8ea16542383eef00d4ad67327c1fa1a359ce29cc`

**Code verdict: MERGE — no findings.**

- Both repository-escape regressions now meaningfully assert `ValueError` with `escapes repository`; focused slice: **11 passed** (including the four previously approved substantive corrections).
- Full Jest/Vitest suites: **186 passed, 1 skipped**.
- Delta from `e9d275e1cfea3dda218c55ba27b2f06ed2698afa` is limited to those two test corrections; worktree clean; local, remote branch, and PR head all match the exact SHA; PR base remains `76403bd758d0c712b6c809bab5e5af5b98ac9a1d`.
- Exact-head checks: `auto-heal`, `heal`, CodeQL, Public CLI Regression, Story Regression, Package Preprocess Smoke, and Repo Bloat Docker E2E **passed**. `Run Unit Tests` is **pending**; orchestrator will hold merge until it completes.

PR #2016, opened 2026-07-13
fix(sync): eliminate runtime digest test barrier race

### LINKED ISSUE #2015: Fix flaky concurrent runtime digest test barrier race
## Problem

`tests/test_sync_core_runner.py::test_released_runtime_digest_hashes_entries_concurrently` intermittently fails in CI with `threading.BrokenBarrierError`.

Observed on PR #2012 at `db89f348` and `8cc8d5a03`; the same PR passed at `a749c44e` and `92e0e76c`. This test and the production implementation originated in #1985. The PR #2012 diff does not touch sync-core code.

## Root cause

The observer performs `thread_ids.add(threading.get_ident())` and then separately checks `if len(thread_ids) <= 2`. With several executor workers, a third worker can grow the set between those operations. One selected worker can enter the barrier while the intended peer observes a size above two and skips it, leaving the first worker to time out.

An equivalent standalone stress reproduced `BrokenBarrierError` 18 times in 2,000 runs. The production `ThreadPoolExecutor.map()` behavior and digest correctness are not implicated.

## Scope

Test-only synchronization correction in `tests/test_sync_core_runner.py`:

- deterministically select exactly two distinct worker participants using lock-protected state or equivalent synchronization;
- keep the assertion that more than one worker handled entries;
- do not weaken/remove the concurrency contract;
- do not change production code without new evidence.

## Acceptance criteria

- Add a deterministic red test/stress harness that demonstrates the old observer race.
- Correct the test synchronization without sleeps or inflated timeouts.
- Focused test passes repeatedly under stress.
- `tests/test_sync_core_runner.py` and the protected sync-core CI subset pass.
- PR follows `docs/runbooks/pr-loop-process.md` with independent review and green CI.

Blocks clean CI closeout for #2012; the fix should land separately and #2012 should then rebase.

gltanaka:
## Summary
- Atomically select exactly two distinct runtime-digest hash workers in the test observer.
- Add a deterministic interleaving harness that reproduces the former stranded-barrier failure.
- Confirm every runtime entry still reaches the real hash function and retain the multi-worker assertion.

Closes #2015.

## Test plan
- [x] Red test commit: deterministic observer harness fails with one BrokenBarrierError before the fix.
- [x] Deterministic observer harness: 2,000/2,000 direct iterations passed after the fix.
- [x] Focused pytest pair passed.
- [x] tests/test_sync_core_runner.py: 37 passed, 46 skipped.
- [x] pylint errors-only, Python compile, and diff checks passed.
- [!] Protected subset ran: 134 passed, 70 skipped, 4 existing trusted-finalizer reporting failures (pytest=COLLECTION_ERROR); no production code or reporting tests are changed here.

### DIFF
--- a/tests/test_sync_core_runner.py
+++ b/tests/test_sync_core_runner.py
@@ -7,6 +7,7 @@ import sys
 import threading
 from datetime import datetime, timezone
 from pathlib import Path, PurePosixPath
+from typing import Callable
 
 import pytest
 import pdd.sync_core.runner as runner_module
@@ -91,6 +92,26 @@ def _run(
         ),
         config=RunnerConfig(timeout_seconds=20),
     )
+
+
+def _observe_runtime_digest_hash_workers(
+    original: Callable[[tuple[str, Path]], tuple[str, bytes] | None],
+    barrier: threading.Barrier,
+    thread_ids: set[int],
+    *,
+    interleaving_hook: Callable[[], None] | None = None,
+) -> Callable[[tuple[str, Path]], tuple[str, bytes] | None]:
+    """Observe runtime-digest hashing workers while selecting two barrier peers."""
+
+    def observed(entry: tuple[str, Path]) -> tuple[str, bytes] | None:
+        thread_ids.add(threading.get_ident())
+        if interleaving_hook is not None:
+            interleaving_hook()
+        if len(thread_ids) <= 2:
+            barrier.wait(timeout=2)
+        return original(entry)
+
+    return observed
 
 
 def test_passing_collected_test_is_pass(tmp_path) -> None:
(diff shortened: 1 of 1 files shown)


gltanaka:
## Pre-fix investigation

- Sibling search found one matching add-then-len worker-selection pattern: this observer in tests/test_sync_core_runner.py; no production occurrence was found.
- History: c016be9dc introduced both concurrent runtime hashing and this observer; 5ab254a09 later merged that work into the sync foundation.
- Decision: keep scope to the sole test observer. Production ThreadPoolExecutor.map behavior and digest logic are not implicated.
- Convention: use a short-lived threading.Lock to make shared-state selection atomic, and release it before waiting at the barrier.

gltanaka:
## TDD red-to-green evidence

- Red commit d242c2e67: the deterministic interleaving test forces a third registration between the second registration and its count. It failed with exactly one BrokenBarrierError, then explicitly aborted the stranded waiter for teardown.
- Fix commit f70528887: selection is lock-protected and one-time per worker. The same harness passed 2,000/2,000 direct iterations; the focused pytest pair and the full runner test file passed.
- End-to-end note: this alters only a test observer around the real runtime-digest hashing path; the focused test invokes the production _released_runtime_closure_digest path and verifies every entry called the original hash function.

gltanaka (commented):
## Adversarial review — PR #2016

**Verdict:** MERGE — no substantive findings.

Reviewed exact SHA f705288870f546a4e4ad32fef14532b0abb08b4b.

- Scope is test-only and limited to tests/test_sync_core_runner.py.
- The lock-protected decision at lines 127-139 atomically selects the first two distinct worker IDs; the lock is released before either barrier wait. The final harness at lines 1311-1362 forces worker one into the barrier before worker two registers, forces a third registration before worker two waits, joins every thread, and cannot green without both selected barrier arrivals.
- The production-path test at lines 1281-1308 still captures and invokes the real _hash_runtime_entry for every entry, asserts complete entry coverage, and retains the multi-worker assertion.
- No sleeps, timeout inflation, production changes, or unsynchronized pass-state were introduced.
- Independent checks: focused pair 2 passed; final atomic harness 2,000/2,000; real hash concurrency path 250/250; final harness with the old observer 2,000/2,000 red; diff check, py_compile, and pylint errors-only passed.
- Audit note: red commit d242c2e67 itself can be forced to false-green if worker one is paused after signaling selection but before entering the barrier. The reviewed head closes that scheduling gap with first_waiting; this is resolved at the reviewed SHA and is not a final-code finding.
- The four reported trusted-finalizer COLLECTION_ERROR failures are pre-existing macOS protected-sandbox behavior in unchanged reporting tests. Exact-SHA Linux CI passed the focused protected-runner step, so they do not affect this PR.

Remaining uncertainty: the local full runner-file pass was interrupted in one unrelated final test by host disk exhaustion after 36 passed and 46 skipped; the broader exact-SHA unit-test CI step was still running when this review was posted.

PR #2017, opened 2026-07-13
fix: bound TypeScript runner discovery (clean replacement for #2003)

Serhan-Asad:
## Summary

Clean, narrow replacement for #2003. This preserves the runner-discovery behavior needed downstream while removing the accumulated review-loop history.

- Finds Jest, Vitest, and Playwright configs for deeply nested TypeScript suites without crossing repository or independent-package boundaries.
- Proves ancestor workspace membership with bounded, fail-closed, provider-specific glob matching.
- Requires every crossed lexical package boundary to have a contained, readable, bounded JSON-object `package.json`; dangling, looping, escaping, unreadable, malformed, scalar, and array leaves fail closed.
- Matches npm's leading-dot and `node_modules` workspace exclusions while preserving explicit-dot declarations.
- Preserves valid npm/Yarn or Lerna membership proof when a different colocated provider declaration is invalid; without positive valid proof, invalid declarations still fail closed.
- Keeps an existing pnpm workspace declaration authoritative, including invalid-data fail-closed behavior.
- Makes request-wide workspace-match budget exhaustion revoke ancestor ownership even when an earlier provider returned `True` and a later provider returns `None`.
- Runs Jest paths literally with `--runTestsByPath` and returns the required runner cwd.
- Rejects runner-config symlinks whose resolved target is outside the proven repository/package/workspace owner.
- Keeps the PDD prompts, exact composite fingerprints, dependency catalog, verification profiles, and protected transition synchronized.

## Exact-head verification

Candidate: `e272f826e77a7a2db5914188939d535b9e7ac930`
Base: `e7735e0f35a0915707142bfd4c767df59f8c3b9e`

- Sol’s three earlier boundary findings remain covered by public resolver regressions: npm dot-directory exclusions, an intermediate independent-package config-symlink bypass, and pnpm `node_modules` workspace exclusion.
- The current P1 is covered by a public resolver regression with npm + Lerna declarations, a shared 250,000-state budget, an observed `[True, None]` provider result, and a leaf Jest-config symlink to the root. Exhaustion now rejects the ancestor proof; no Jest command or runner cwd is returned.
- `tests/test_get_test_command.py` + `tests/test_agentic_langtest.py`: **167 passed**.
- `tests/test_sync_core_manifest.py`: **28 passed**.
- Rollout/verification-profile suites: **147 passed**.
[...]

Issue #1976, opened 2026-07-10
Honor architecture.json filepath for prompts in nested .pddrc contexts


Serhan-Asad:
## Problem

When a `.pddrc` context uses a nested `prompts_dir`, `get_pdd_file_paths()` builds its architecture lookup key relative to that nested directory. `architecture.json` filenames remain relative to the repository `prompts/` root, so the lookup misses the declared `filepath` and falls back to the wrong output template.

This can make `pdd sync` and `pdd update` target a nonexistent code file and fail with `Code file not found`.

## Scope

- Preserve the primary architecture lookup behavior.
- When it misses, retry with a key relative to the repository `prompts/` root.
- Preserve behavior for repositories without this nested layout.
- Add regression coverage for nested `.pddrc` prompt directories.

## Acceptance criteria

- Nested-context prompts honor their `architecture.json` `filepath`.
- Previously working prompt resolution remains unchanged.
- The pdd_cloud endpoint-test reproduction resolves to the correct code files.
- Relevant resolver and sync tests pass.

## Implementation

Tracked by PR https://github.com/promptdriven/pdd/pull/1971.

Downstream consumer tracking: https://github.com/promptdriven/pdd_cloud/issues/3203 and https://github.com/promptdriven/pdd_cloud/pull/3205.

Serhan-Asad:
7/10/2026

Serhan-Asad:
7/13/2026: Targeting this issue today as part of the active PR campaign.

Serhan-Asad:
7/14/2026

Serhan-Asad:
7/16/2026

Serhan-Asad:
7/20/2026

PR #1971, opened 2026-07-10
Honor architecture.json filepath for prompts in nested .pddrc contexts

Serhan-Asad:
## Problem

`get_pdd_file_paths()` gives `architecture.json.filepath` tier-1 precedence for code outputs, but a nested `.pddrc` `prompts_dir` can strip filename prefixes retained by architecture entries. The original fallback fixed only repositories rooted at `<project>/prompts`, still missed canonical/custom prompt layouts, ignored exact example/test output templates, reparsed architecture on misses, and made uncontained architecture paths reachable by sync.

Downstream report: promptdriven/pdd_cloud#3203. Tracking issue: #1976.

## Fix

- Resolve prefix-stripped architecture names in the primary lookup without assuming the prompt root is named `prompts`.
- Validate architecture prompt filenames before discovery; reject absolute, parent-traversal, backslash, and symlink-resolved prompt escapes.
- Establish prompt ownership across configured sibling roots so a narrowed backend root cannot borrow a same-leaf frontend module.
- Restrict filepath-stem compatibility matching to entries that are not themselves recognized prompt filenames.
- Compare root-relative prompt identities without resolving caller-influenced paths, while preserving trusted root symlink aliases such as `prompts -> pdd/prompts`.
- Keep architecture authoritative for code while honoring exact `.pddrc` `outputs.example.path` and `outputs.test.path` templates.
- Reject unsafe architecture output paths before generation can use them.
- Replace repeated whole-tree ownership scans with bounded, cached directory walks and retain one architecture parse per resolution.
- Align the governing prompt, path-construction guide, Unreleased changelog, and drift metadata.

Undeclared deliverables still require a downstream architecture entry; core intentionally does not guess a non-template output path when no source-of-truth entry exists.

## Validation

- 509 resolver and backward-compatibility tests passed.
- 244 affected module and metadata tests passed; 244 broader sync/architecture tests also passed.
- New coverage includes missing internal prompts with unsafe architecture hints, sibling backend/frontend contexts, differently named prompt aliases, trusted prompt-root symlink aliases, custom roots, canonical normalization, exact code/example/test outputs, duplicate flat/nested basenames, unsafe output paths, and symlink escapes.
- A 191-prompt differential sweep against pdd_cloud PR #3205 found zero unintended changes versus the prior PR head; the 32 intended downstream corrections remain intact and the seven expected downstream mismatches remain visible.
[...]

PR #2019, opened 2026-07-13
Honor architecture.json filepath for prompts in nested .pddrc contexts

Serhan-Asad:
## Clean replacement for #1971

This is the narrow, reviewed replacement for #1971. It is intentionally stacked on #2017 so GitHub shows only the resolver change; merge #2017 first, then retarget/rebase this PR to `main` before the manager merges it.

## Problem

When a nested `.pddrc` sets `prompts_dir` to a subdirectory such as `prompts/backend`, prompt discovery can produce a context-relative filename such as `credits_Python.prompt` while `architecture.json` stores the repository-prompt-root-relative name `backend/credits_Python.prompt`. That mismatch can skip the authoritative architecture filepath and incorrectly fall back to configured output paths.

## Fix

- Preserve the context-relative lookup as the primary path.
- Retry exactly once with a filename relative to `<architecture root>/prompts` only after a complete primary miss.
- Keep a matching primary architecture row with a null filepath authoritative; do not borrow an alternate row.
- Do not activate the retry for custom prompt roots outside `<architecture root>/prompts`.
- Compute lookup keys lexically so approved prompt-root aliases remain valid.
- Parse `architecture.json` once and thread a frozen module snapshot through ambiguity checks, prompt discovery, and code-path selection, including failed/empty loads.
- Continue using that snapshot if the architecture file disappears mid-resolution.
- Update the governing PDD prompt plus fingerprints and verification-profile rotation required by the prompt change.

## Validation

- 203 resolver tests passed.
- 268 resolver/metadata/profile/governance tests passed after the final review fix.
- 369 broader sync/architecture/alias-policy tests passed earlier in the bounded review cycle.
- Python compilation, JSON parsing, metadata/profile hash recomputation, and `git diff --check` passed.
- Full GPT-5.6-sol xhigh review found one actionable snapshot-disappearance issue; it was fixed.
- Targeted GPT-5.6-sol xhigh re-review of that finding ended with zero actionable findings at `a9876f16dad191875333664e7527b12dda2ac521`.

Tracking issue: #1976. Supersedes #1971 after this replacement is accepted.

Manager order: merge #2017 first; then retarget/rebase this PR to `main` and merge it second.


Serhan-Asad:
Closing as superseded by merged #1971 (`e10bd9b3d0d5ac94d1a56af88f5abf07cf8af775`) and its merged base #2017 (`ee9fcff457b23fb7123bb7e15666c9287409ad0f`); any intentionally unique delta should return as a narrow PR against current main.

Issue #2091, opened 2026-07-15
CI: #1995 cgroup smoke and installed-wheel descriptor transport fail on hosted Linux


gltanaka:
## Problem

PR #1995 exact head `82c1671d94176f3cd5204f7ab9345b301abe2a95` failed the required hosted integration run after #2088 was merged into its branch.

Run: https://github.com/promptdriven/pdd/actions/runs/29449288830

### Focused protected-runner failure

The cgroup smoke reports `cgroup_exists: false`, no cgroup members, then fails `privileged helper namespace was not observable`. The focused lane ends with **20 failed / 699 passed / 16 skipped**; later supervisor cases return construction error 125 because the protected candidate environment is invalid.

### Installed-wheel Playwright failure

All seven admitted config variants (`.js` commonjs/module, `.cjs`, `.mjs`, `.ts`, `.cts`, `.mts`) block in `_read_descriptor_frame_fd` / `_descriptor_poll` and hit pytest timeout at 90 seconds. Result: **7 failed** in 639.79s.

## Required investigation/fix

- Reproduce on the exact #1995 head and identify whether cgroup creation/visibility/cleanup ordering also prevents the descriptor-producing helper from starting or acknowledging.
- Preserve the reviewed #2088 security properties: candidate-only limited leaf, trusted monitor outside the candidate budget, exact path translation, monotonic event deltas, no raw candidate output, no blind retry.
- Add red regressions for the proven hosted failure(s), then a narrow correction.
- Verify source cgroup smoke, focused supervisor suite, installed-wheel Playwright protocol, serial and two-worker real-Vitest pressure matrix, lint/compile/diff, and full hosted gates.

## Non-goals

No gate bypass, timeout inflation without root-cause proof, production/release changes, or weakening of fail-closed descriptor/reporter semantics.

## Dependency

#2090 must merge to main first for #1995 auto-heal authorization. The implementation may be stacked on exact #1995 head for diagnosis, but #1995 must rebase onto updated main and rerun all gates before merge.

gltanaka:
Exact-head hosted evidence for correction PR #2096 (`8d50d9d1d`):

- Run https://github.com/promptdriven/pdd/actions/runs/29454718989
- Dedicated cgroup-v2 containment smoke: passed
- Installed-wheel/package Playwright smoke: passed
- Focused protected-runner suite: 720 passed, 16 skipped, 4 failed
[...]

PR #1995, opened 2026-07-11
fix(sync): add trusted Playwright verification adapter

gltanaka:
## Summary
- add fail-closed trusted `validator_id=playwright` adapter
- bind static Playwright config, local import/support closure, JSON reporter identities, and candidate support mutations
- compare protected-base, candidate collection, and candidate execution identities

## Test plan
- `conda run -n pdd pytest -q tests/test_sync_core_runner_playwright.py tests/test_sync_core_runner_vitest.py tests/test_sync_core_runner_jest.py`
- `conda run -n pdd pylint --disable=C0301,C0302,R0911,R0912,R0913,R0914,R0917,W0613 pdd/sync_core/runner.py tests/test_sync_core_runner_playwright.py`
- Local Playwright package/browser unavailable: contract fixture exercises `--list` and execution protocol, but a real browser execution remains unavailable.

PR #2096, opened 2026-07-15
fix(sync): restore hosted cgroup boundary invariants

gltanaka:
Stacked correction for #1995. Addresses https://github.com/promptdriven/pdd/issues/2091.

Hosted failure evidence: https://github.com/promptdriven/pdd/actions/runs/29453601787/job/87481651122

Summary:
- make privileged staging placeholders independent from the supervision UUID so fixed test tokens cannot alias distinct bind sources
- preserve the sandbox-parent namespace handshake while mounting only the limited candidate leaf at /sys/fs/cgroup
- move the pre-setpriv child into that leaf through its mounted root before untrusted exec
- keep the candidate unable to see parent cgroups or migrate to another leaf

TDD commits:
- 6cf64c2bd test(sync): reproduce staging token aliasing
- 02cb3e3d2 fix(sync): isolate privileged staging identities
- 33b34bfa6 test(sync): require limited cgroup view in sandbox
- 8d50d9d1d fix(sync): expose candidate cgroup limits in sandbox

Local evidence:
- focused boundary tests: 4 passed
- supervisor: 217 passed, 41 skipped
- Playwright adapter: 252 passed, 16 skipped
- Vitest adapter with two workers: 144 passed, 1 skipped
- pylint supervisor: 10.00/10
- compile and diff checks: passed

Remaining hosted-only verification: real Bubblewrap/cgroup v2 containment, exact environment handoffs, descriptor cleanup variants, OOM and pids event evidence, and installed-wheel adapter smoke.


gltanaka:
Critical review scoped to 58beab3886afac88de99c05a5c36cbb39b427acf..8d50d9d1d2228609c3efd7cc2fe956b76acbf3b4 (tree de0510d3abe774f63a90d3a5ddaa6449dfb9b73d), notwithstanding the temporary CI-only retarget to main.

Finding — HIGH / HOLD: exact-head hosted run 29454718989 still fails four real privileged descriptor cases: stalled-observation-reader, initial-scan-failure, initial-watched-assertion-failure, and fd-only-namespace-holder-cleanup. Each reports that the privileged helper namespace was not observable with cgroup_exists=false; fallback cleanup also cannot recover the exact control directory. Thus the staging-token and leaf-view fixes improve the boundary but do not close the hosted cleanup/observation failure, and the lane stops before authenticated Playwright/Vitest protocols and the full unit suite. Do not integrate this range until those four cases pass at the exact head.

Positive evidence: dedicated hosted cgroup-v2 containment smoke passed (visible memory/PID limits, no child cgroups, candidate migration denied, OOM/PID enforcement, no leaks); installed-wheel real Playwright passed; other hosted jobs passed. Local exact-head tests: focused three new/updated tests 3 passed; full tests/test_sync_core_supervisor.py 217 passed, 41 skipped; pylint pdd/sync_core/supervisor.py 10.00/10; diff-check/compile clean. No source edits made.

Remaining risk after fixing the blocker: explicitly verify software that derives its cgroup path from /proc/self/cgroup, because the cgroup namespace is rooted at the sandbox parent while /sys/fs/cgroup is now leaf-rooted. Current Python/Node/Chromium enforcement paths are covered, but that introspection compatibility is not.

Recommendation: HOLD, fix the four exact hosted failures, then rerun the complete exact-head Unit Tests workflow and re-review. No merge performed.

gltanaka:
Closing as superseded. #1995 moved to `d18819dc1`, which reverted the failed `58beab388` cgroup topology and replaced the UUID/staging workaround with the narrower independent supervision-authority fix. Deterministic comparison showed none of #2096 should carry forward. The remaining hosted-only lifecycle race is isolated in test-only draft #2097.

Evidence: #2091, hosted runs 29454718989 and 29455075570, and critical HOLD reviews on #2096/#1995. No #2096 commit was merged.

PR #2097, opened 2026-07-15
fix(sync): synchronize hosted observation lifecycle

gltanaka:
Stacked correction for #1995. Addresses the remaining hosted-only failures from #2091 after exact base `d18819dc158ebbc4e5b46a718b018105cedc7e0f`.

Root cause:
- the four failures shared a test-fixture lifecycle race, not a production cgroup boundary failure
- the candidate completed and the production supervisor correctly exhausted its fixed five-second external observation-handoff deadline while the hosted privileged `/proc` scanner was still running
- teardown then removed the exact cgroup and `TemporaryDirectory`, so the later scan saw a zombie coordinator and no owned scope/control path

Correction:
- keep the production tree and every production timeout byte-identical to the base
- gate the real test only after `_descriptor_result` has authenticated the nonce, aggregate digest, payload, and observation
- use exclusive inherited ready/release pipes so ownership capture occurs before the unchanged five-second handoff deadline is created
- include all gate descriptors in exact fallback cleanup for initial scan, watched assertion, and FD-only namespace-holder failures
- prove release-pipe EOF on the normal stalled-reader cleanup path

TDD commits:
- `eefd3da5a` RED: deterministically hold the candidate without release, proving the fixture cannot reach the required external-handoff SANDBOX_ERROR until lifecycle release is explicit
- `fae3f4ff3` GREEN: synchronize at the authenticated descriptor boundary and preserve the exact production timeout behavior

Local evidence:
- `tests/test_sync_core_supervisor.py`: 214 passed, 40 skipped
- Playwright + Vitest adapter suites (`-n 2`): 396 passed, 17 skipped
- `pylint pdd/sync_core/supervisor.py`: 10.00/10
- compile, diff-check, and production-tree identity checks: passed

Required hosted evidence before integration:
- all four exact real descriptor cases green
- full focused protected-runner lane green
- installed-wheel Playwright and cgroup smoke remain green
- full unit workflow green


gltanaka:
Critical review — HOLD

Exact head `fae3f4ff38e7f13ac25dfbe289e22e2fa3132b78` (tree `23c98e793eedf548fb109dc66c376a9d3748cc3e`) against stacked base `d18819dc158ebbc4e5b46a718b018105cedc7e0f`.

Release-blocking finding: exact-head hosted run 29456592150 still fails all four targeted real descriptor lifecycle variants. Each now receives EOF (`b""`) on `lifecycle_ready_read` instead of the authenticated ready byte (`b"1"`), and fallback reports that the exact control directory is unavailable. Therefore the coordinator exits before `_descriptor_result` returns and before the new authenticated gate is reached. This directly disproves the proposed root cause that the hosted scanner merely loses a live scope while the unchanged five-second external handoff is expiring; the correction never reaches that handoff boundary in hosted Linux.

Static review found no independent defect in the intended gate mechanics: the gate is placed after `_descriptor_result` validates nonce, aggregate digest, payload, and observation; fork sides close their opposite pipe ends; parent-owned gate descriptors are included in fallback cleanup; the normal path checks ready-pipe EOF. The net diff changes only `tests/test_sync_core_supervisor.py`; every production path and production timeout is byte-identical to the base, and `git diff --check` is clean. That production equality also means this test-only PR cannot correct the observed pre-gate hosted exit.

Local exact-head `tests/test_sync_core_supervisor.py`: 214 passed, 40 skipped (the four decisive cases are Linux-only). Other hosted jobs passed; Unit Tests failed at 4 failed, 716 passed, 16 skipped.

Recommendation: HOLD. Capture and expose the coordinator report/pre-gate exception before asserting the ready byte, diagnose the production/helper exit, then correct the actual boundary and require a fully green exact-head hosted run. Remaining risk is the same Linux-only privileged descriptor/cgroup lifecycle failure blocking #1995.

gltanaka:
Critical final review — HOLD

Exact head `ca0353d3ad6c07b9cf0d803ba864685e90e61c48` (tree `a73a6769272e5cdba18fea9b681b737e45a0e65a`) against stacked base `1108a585fddcdd9715d233c72058bfb6f4f69dc2`.

The six-line scratch-root correction is correct and narrowly scoped. Creating `tmp_path/stalled-candidate-scratch` and using it as both `cwd` and the sole writable root keeps `pdd-scope-*` as a sibling, so helper `copy_owned(scratch, target)` cannot recursively ingest its concurrently created control/binds tree. Nested control mounts remain under the same control prefix; all adversarial mount/namespace assertions, saturation, handoff barrier, cleanup deadlines, and candidate timeout are unchanged. The diff changes only `tests/test_sync_core_supervisor.py`; non-test/production equality and `git diff --check` pass. Local full supervisor: 216 passed, 40 skipped.

Hosted evidence shows the scratch fix worked but exposes release-blocking defects in the inherited `1108a585f` barrier/cleanup fixture. Exact run 29459107208 reaches the handoff and the normal `stalled-observation-reader` case now passes, but three adversarial variants fail: 3 failed, 719 passed, 16 skipped.

Findings:
1. In both injected-initial-failure cases, `_fallback_stalled_observation_cleanup` closes `read_fd` and then raises before returning. Because tuple assignment occurs only on successful return, the caller retains stale fd `14`; the postcondition sees `14 != -1`, and `finally` double-closes it, producing `EBADF` that masks the intended primary/cleanup evidence. Closed ownership must be reflected even when cleanup reports an error.
2. The FD-only holder is launched through `sudo`, but fallback tries `os.kill` from the unprivileged runner and receives `EPERM`. That path cannot prove adversarial root-owned namespace-FD cleanup as written.
3. The barrier ready/release descriptors are not included in the fallback-owned descriptor tuple for the early-failure paths. The coordinator can remain blocked at the observer barrier while fallback attempts teardown, weakening deterministic ownership/closure.

Recommendation: HOLD/no merge despite the six-line change itself being sound. Correct the stacked barrier/fallback descriptor ownership and privileged external-holder termination without weakening the nested-mount assertions or timeouts, then require a fully green exact-head hosted Unit Tests job. Also refresh the stale PR body before integration; it still describes an older base and obsolete commits.

gltanaka:
## Critical adversarial review — PR #2097 @ `61c09edb61ee4945440fb70b082973c5524e1a6b`

**Verdict:** DO NOT MERGE
**Findings:** 1 total (0 P1, 1 P2)

### P1 (blocking)

None.

### P2 (non-blocking)

- [P2] Preserve lifecycle child timeout classification — `pdd/sync_core/lifecycle.py:648-665`
  The previously reported finding remains valid at this exact head. When the transaction wrapper's lifecycle child reaches its internal deadline, it raises `RuntimeError("lifecycle child deadline expired")`; the wrapper exits 1 without a receipt, while `run_lifecycle_matrix` classifies only transaction status 124 as a timeout. The result is therefore recorded as a generic failure, not a timeout, skewing certificate evidence for long-running scenarios. This source is byte-identical to stacked base `d2d85fe400fd864037ff498b03da1e798c297189`, so #2097 did not introduce it, but the #1995 integration chain must correct or explicitly disposition it before merge.
  **Suggested fix:** Preserve a closed timeout signal from the child through the transaction boundary and test both outer transaction expiry and inner lifecycle-child expiry.

### Notes

- The #2097 net diff is test/workflow-contract-only: `tests/test_sync_core_supervisor.py` and `tests/test_unit_tests_workflow.py`; production supervisor/runner/lifecycle files are byte-identical to the stacked base.
- The cleanup correction is fail-closed and I found no new P1/P2 defect in it: captured mounts are unmounted before coordinator or systemd teardown can delete paths; FD-only mode reselects only a complete exact captured FD-holder identity after scope teardown; ambiguous `umount` absence is classified before the diagnostic is truncated and can be accepted only after a root-identity-valid held-namespace rescan proves exact absence; holder loss, root mismatch, scanner failure, surviving mounts/processes/cgroups/FDs, malformed unit state, and command timeouts remain errors. The lexical mount allowlist and authenticated held-root checks remain in force; there is no blanket timeout, cgroup, symlink, identity, or cleanup weakening.
- Local exact-head validation: `pytest -q tests/test_sync_core_supervisor.py tests/test_unit_tests_workflow.py` → 286 passed, 48 skipped; `git diff --check` passed.
- Hosted run [29481807534](https://github.com/promptdriven/pdd/actions/runs/29481807534) proves the held-namespace smoke, focused protected-runner selection, real Playwright/authenticated supervisor protocols, package smoke, public/story regressions, and auto-heal green. The required full Unit job is red: 1 unrelated Vitest sandbox termination failure, 14,552 passed, 51 skipped, 1 xfailed. Even though that failure is outside #2097's diff, the runbook merge gate is not green.
- The PR body is stale (old base/root-cause narrative and obsolete commit list) and should be refreshed before any later integration decision.

gltanaka:
The inherited timeout-classification P2 from the exact-head critical review is now tracked separately in #2110. It must be corrected after #2108 integrates into #1995 and before #2097 merges; #2097 itself did not introduce it.

gltanaka:
Rebase/integration evidence (2026-07-16):

- old stacked base: `d2d85fe400fd864037ff498b03da1e798c297189`
- exact new #1995 head/base: `cf79699bcf7a23d492b3cd3905ee6f32728809ff`
- old #2097 head: `a125234c97fe92e075954636c8c464c28ba467ce`
- new #2097 head: `0fcb783e5a6bd234794ea3c9e11783fe9790b727`
- rebase completed without manual conflicts
- workflow-path RED/GREEN commits were dropped automatically because their patches are already upstream; `git range-diff` shows the remaining 14 commits patch-equivalent
- current net stack diff is only `tests/test_sync_core_supervisor.py`; no accidental base commit duplication (`git cherry` and merge-base checks passed)
- `git diff --check`: passed
- `python -m compileall -q pdd tests/test_sync_core_supervisor.py tests/test_unit_tests_workflow.py`: passed
- exact PDD interpreter focused validation: `pytest -q tests/test_sync_core_supervisor.py tests/test_unit_tests_workflow.py` -> 299 passed, 48 skipped in 30.61s

The inherited lifecycle timeout-classification P2 remains tracked by #2110 and is not addressed by this rebase. #2108 is also outside this task. Hosted exact-head checks are still required before merge.

Issue #2083, opened 2026-07-15
CI: real Vitest sandbox intermittently aborts without actionable trusted diagnostic


gltanaka:
## Problem

Hosted Unit Tests for PR #2077 failed at exact head `1865ff88db3731a957d566ee09c621682a8de1aa` after 14,040 tests passed because the real Vitest sandbox process aborted with `SIGABRT` before producing its trusted reporter artifact:

```
tests/test_sync_core_runner_vitest.py::test_real_vitest_runs_copied_entrypoint_without_candidate_result_access
Vitest infrastructure termination: reporter=missing; kind=signal; signal=SIGABRT; signal_number=6
diagnostic_sha256=abbad546e1e6ab7a6c693dad51c9635b43f7504c8cf1f1afb5a1017260b72c35
```

Run: https://github.com/promptdriven/pdd/actions/runs/29441372583/job/87440940869

The failure is unrelated to #2077 changed files (two dormant profile approvals plus bootstrap tuples/tests), but it blocks a required full-suite gate. The current fail-closed outcome is correct; the operational gap is that the diagnostic is only a hash, so the root cause of a trusted infrastructure abort cannot be distinguished from runner resource contention or a deterministic sandbox/toolchain defect.

## Required investigation / general fix

- Reproduce the exact test serially and under the hosted full-suite xdist load using the provisioned, identity-bound Vitest toolchain.
- Determine whether the abort is V8/Node resource exhaustion, bubblewrap/cgroup pressure, toolchain corruption, or a deterministic adapter defect.
- Preserve fail-closed semantics; do not reinterpret or silently pass a missing reporter.
- Add bounded trusted infrastructure telemetry sufficient to diagnose signals (for example trusted launcher/cgroup/resource facts and a sanitized crash classification), without exposing candidate-controlled stdout/stderr or secrets.
- If retry is justified, retry only a typed infrastructure termination with strict bounds and retain both attempts as evidence; never retry assertion/test failures.
- Add a regression test for the proven root cause and verify the real hosted Vitest lane plus the full Unit suite.

## Evidence

The failure occurred at 89% during the full two-worker suite and completed in about 3.5 seconds; all surrounding Vitest adapter tests passed. Final result was 1 failed, 14,040 passed, 35 skipped, 1 xfailed. No existing issue was found by searches for the exact test name, SIGABRT, or the infrastructure-termination message.

gltanaka:
[...]

PR #2106, opened 2026-07-16
fix(sync): bound held namespace diagnostic descendants

gltanaka:
## Summary

- bound only the hosted Linux held-namespace diagnostic subprocess tree so a timed-out `sudo -> helper -> nsenter -> scanner` chain cannot retain captured pipes indefinitely
- preserve the sealed memfd/authenticated scan frame, strict fail-closed parsing, and existing five-second diagnostic budget
- leave all production `pdd/` code and non-diagnostic cleanup behavior unchanged

## Root cause

`subprocess.run(..., timeout=...)` kills only its direct child before draining captured pipes. If a descendant remains alive with stdout/stderr inherited, that drain is unbounded. The new privileged diagnostic chain can therefore outlive the killed `sudo` process and hang the whole focused hosted step, matching canceled run 29468333457.

## TDD evidence

- RED at `463c0eadb`: the orphaned pipe-holder lifecycle failed because the bounded diagnostic executor did not exist
- GREEN at `1038618aa`: focused diagnostic/cleanup selection 9 passed
- full changed module: 231 passed, 47 skipped
- `git diff --check`: clean
- pylint: no new diagnostics; the module retains pre-existing lint debt

## Hosted evidence

This is intentionally stacked on PR #1995 exact prior head `385ab8872`. The Unit Tests workflow only triggers for PRs whose base is `main`, so this draft cannot produce hosted evidence without an explicit, reviewed trigger/retarget decision. Do not merge this stack; hosted Linux must emit either the authenticated scan frame or the bounded `scan exact held mount namespace timed out` failure before integration.

Refs #1995
Refs #2091
Refs #2083

### DIFF
--- a/tests/test_sync_core_supervisor.py
+++ b/tests/test_sync_core_supervisor.py
@@ -5768,6 +5768,29 @@ def test_held_namespace_scan_uses_authenticated_sealed_fd_transport() -> None:
     assert "scan_payload" not in control_record
 
 
+def test_held_namespace_diagnostic_timeout_reaps_pipe_holding_descendant() -> None:
+    """A timed-out diagnostic cannot hang while an orphan retains its pipes."""
+    source = r"""
+import os,signal,time
+child=os.fork()
+if child==0:
+    signal.signal(signal.SIGTERM,signal.SIG_IGN)
+    time.sleep(60)
+    os._exit(0)
+os.write(1,b'ready\n')
+time.sleep(60)
+"""
+    started = time.monotonic()
+
+    with pytest.raises(subprocess.TimeoutExpired):
+        _run_bounded_held_namespace_diagnostic(
+            [sys.executable, "-I", "-S", "-c", source],
+            stdin=None, timeout=.2,
+        )
+
+    assert time.monotonic() - started < 2
+
+
 @pytest.mark.skipif(not sys.platform.startswith("linux"), reason="requires Linux memfd seals")
 def test_held_namespace_scan_memfd_is_sealed_and_only_transport_fds_inherit(
     tmp_path: Path,
--- a/tests/test_sync_core_supervisor.py
+++ b/tests/test_sync_core_supervisor.py
@@ -5768,29 +5768,6 @@ def test_held_namespace_scan_uses_authenticated_sealed_fd_transport() -> None:
     assert "scan_payload" not in control_record
 
 
-def test_held_namespace_diagnostic_timeout_reaps_pipe_holding_descendant() -> None:
-    """A timed-out diagnostic cannot hang while an orphan retains its pipes."""
-    source = r"""
-import os,signal,time
-child=os.fork()
-if child==0:
-    signal.signal(signal.SIGTERM,signal.SIG_IGN)
-    time.sleep(60)
-    os._exit(0)
-os.write(1,b'ready\n')
-time.sleep(60)
-"""
-    started = time.monotonic()
-
-    with pytest.raises(subprocess.TimeoutExpired):
-        _run_bounded_held_namespace_diagnostic(
-            [sys.executable, "-I", "-S", "-c", source],
-            stdin=None, timeout=.2,
-        )
-
-    assert time.monotonic() - started < 2
-
-
 @pytest.mark.skipif(not sys.platform.startswith("linux"), reason="requires Linux memfd seals")
 def test_held_namespace_scan_memfd_is_sealed_and_only_transport_fds_inherit(
     tmp_path: Path,
--- a/tests/test_sync_core_supervisor.py
+++ b/tests/test_sync_core_supervisor.py
@@ -5330,6 +5330,39 @@ def _hold_validated_descriptor_result(
     return result
 
 
+def _run_bounded_held_namespace_diagnostic(
+    argv: list[str], *, stdin: bytes | None, timeout: float,
+) -> subprocess.CompletedProcess[bytes]:
+    """Run the privileged diagnostic with a killable descendant process group."""
+    process = subprocess.Popen(  # pylint: disable=consider-using-with
+        argv,
+        stdin=subprocess.PIPE if stdin is not None else subprocess.DEVNULL,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        start_new_session=True,
+    )
+    try:
+        stdout, stderr = process.communicate(input=stdin, timeout=timeout)
+    except subprocess.TimeoutExpired as error:
+        try:
+            os.killpg(process.pid, signal.SIGKILL)
+        except ProcessLookupError:
+            pass
+        drain_timeout = min(1.0, max(.1, timeout))
+        try:
+            process.communicate(timeout=drain_timeout)
+        except subprocess.TimeoutExpired:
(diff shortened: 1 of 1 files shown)


gltanaka:
## Pre-fix investigation

**Hosted evidence reviewed**
- PR #1995 run 29468333457 reached `Run focused protected-runner tests`, then remained there until cancellation after about 43 minutes; earlier cgroup, sandbox, signer, package, and installed-wheel lanes passed.
- Run 29463850704 at `34d5054d` had already narrowed the underlying functional failure to `fd-only-namespace-holder-cleanup`; this stack does not claim to fix that mount-ownership defect.

**Related issues and commits**
- #2091 tracks the hosted privileged-observation lifecycle; #2083 retains the separate generic SIGABRT scope.
- `e4f20a6d1`, `2b07834f4`, `4c62e17cd`, `48ceab521`, and `385ab8872` introduced and corrected the diagnostic chain. The established convention is bounded, authenticated transport with sanitized failure evidence and no blind retry.

**Sibling search**
- The unbounded post-timeout pipe-drain pattern is specific to the new test-only privileged diagnostic chain. Production code was not changed, and the other cleanup commands retain their current behavior.

**Scope decision**
- Fix only the scan command by placing its full descendant tree in a new process group and killing that group on the existing timeout. Keep the sealed memfd frame, digest binding, parser caps, cleanup deadline, and normal command executor unchanged.

gltanaka:
## TDD RED evidence — `463c0eadb`

Command:
```
pytest -q tests/test_sync_core_supervisor.py::test_held_namespace_diagnostic_timeout_reaps_pipe_holding_descendant --tb=short
```

Result:
```
FAILED tests/test_sync_core_supervisor.py::test_held_namespace_diagnostic_timeout_reaps_pipe_holding_descendant
NameError: name '_run_bounded_held_namespace_diagnostic' is not defined
1 failed
```

The regression forks a descendant that ignores SIGTERM and retains captured stdout while its direct parent sleeps. The required behavior is a `subprocess.TimeoutExpired` within two seconds with the entire diagnostic process group reaped.

gltanaka:
## Local verification — `1038618aa`

- Focused diagnostic and cleanup selection: **9 passed**
- Full changed module: **231 passed, 47 skipped**
- `python -m py_compile tests/test_sync_core_supervisor.py`: passed
- `git diff --check origin/codex/issue-1932-playwright-adapter...HEAD`: passed
- Worktree clean; all commits pushed

Hosted uncertainty remains explicit: this stacked draft does not satisfy the `base: main` and non-draft workflow conditions. It must not merge until the exact patch is exercised on hosted Linux and produces either the authenticated held-namespace scan frame or the precise bounded timeout evidence.

gltanaka:
## Critical review — exact stack `385ab8872..1038618aa`

**Verdict: REQUEST CHANGES**  
**P1: 1 · P2: 1**

### P1 — the regression does not reproduce the claimed prior hang
`test_held_namespace_diagnostic_timeout_reaps_pipe_holding_descendant` calls the newly named helper directly. At RED commit `463c0eadb` it therefore fails immediately with `NameError`; it never exercises the old `subprocess.run(..., timeout=...)` path. On POSIX CPython, `subprocess.run` kills and waits for the direct child on `TimeoutExpired` without an unbounded post-timeout pipe drain. Running the test's parent/pipe-holding-child shape through the old call returned `TimeoutExpired` in **0.203 s** with `output=b'ready\n'`, despite the child retaining the pipe. This does not support the stated root cause for hosted run `29468333457`, so the patch is not yet evidence that the observed 43-minute stall is fixed. Reproduce the stall through the actual pre-patch `_fallback_stalled_observation_cleanup` path (or identify the genuinely unbounded operation) before integration.

### P2 — the test proves elapsed time, not descendant cleanup or call-site wiring
The test asserts only `TimeoutExpired` and elapsed time. An executor that closes its pipes and returns while leaving the descendant alive would pass. It also never invokes `_fallback_stalled_observation_cleanup`, so removing `bounded_descendants=True` from the scan call would leave this regression green. Capture the descendant identity, assert it is gone within a fixed bound, and exercise the real scan-command path while retaining deterministic injected-runner coverage.

### Verified invariants
- Production diff is empty: the `pdd/` subtree is identical (`66c3c1b86e535dce3f5a73f064f6a568b115ae8a` at base and head); only `tests/test_sync_core_supervisor.py` changes.
- The new executor remains byte-mode for scan stdin/stdout/stderr; authenticated sealed-frame construction/parsing is unchanged and timeout remains fail-closed at the caller.
- `start_new_session=True` plus `killpg(process.pid, ...)` scopes the signal to the newly created group; injected non-default runners retain the existing deterministic path.

### Checks
- `git diff --check 385ab8872..1038618aa` — passed
- focused held-namespace/fail-closed selection — **11 passed, 1 skipped, 266 deselected**
- direct old-path reproduction probe — **returned in 0.203 s**, contradicting the proposed hang mechanism
- hosted validation — unavailable on this stacked draft

gltanaka:
## Implementer response to critical review

@codex Re: https://github.com/promptdriven/pdd/pull/2106#issuecomment-4988250511

### P1: RED does not reproduce the old fallback behavior
Status: **FIXING / re-investigating**
The NameError RED is insufficient, and the prompt synthetic returning promptly contradicts the asserted root cause. I will exercise the actual `_fallback_stalled_observation_cleanup` scan call on exact pre-patch head `385ab8872`, using a deterministic subprocess harness so the investigation itself cannot hang.

### P2: no descendant-gone or real-wiring proof
Status: **FIXING if and only if P1 reproduces**
Any retained correction must assert both that the actual fallback uses the bounded path and that the pipe-holding descendant is gone. If the pre-patch fallback does not reproduce, I will revert the speculative commits and close this draft with the contrary evidence.

gltanaka:
## Correction-cycle outcome: withdrawn

The critical review is correct. I could not substantiate the proposed root cause:

- POSIX CPython `subprocess.run` handles `TimeoutExpired` by killing and waiting for the direct child, then immediately re-raising; it does not perform the claimed unbounded post-timeout pipe drain.
- The independent old-path synthetic returned in 0.203 seconds even while a descendant retained the pipe.
- Run 29468333457 has no retained per-test log or authenticated scan frame. It proves only that the full focused step remained active until cancellation, not that `_fallback_stalled_observation_cleanup` or its namespace scan was the stalled call.
- Earlier comparable focused steps completed in roughly 10–12 minutes. The 43-minute run is anomalous, but the exact test/call site remains unidentified.

Accordingly, commits `06222bd0d` and `549c9b972` revert the speculative fix and its insufficient NameError RED. The branch now has **zero net diff** from base `codex/issue-1932-playwright-adapter`.

Post-revert verification:
- full `tests/test_sync_core_supervisor.py`: **230 passed, 47 skipped**
- compile: passed
- exact base-to-head diff: empty
- worktree clean and pushed

This PR is being closed without merge. The blocker returns to #2091/#2083: obtain test-level bounded timing/evidence for the focused hosted lane before changing the diagnostic or cleanup implementation.

gltanaka:
Closing as withdrawn after the single correction cycle: the proposed root cause could not be reproduced or substantiated, and the branch has been reverted to an empty net diff.

PR #2107, opened 2026-07-16
ci: diagnose issue 1995 pytest node lifecycle (never merge)

gltanaka:
## Purpose

Diagnostic-only draft for #1995. **Never merge this PR.** It preserves source commit `385ab8872411e740480bb164fb2b840b91f2624c` and runs the unchanged six-file focused protected-runner selection exactly once with durable pytest lifecycle evidence.

An unmatched lifecycle record identifies only the last observed node/phase; it is not a root-cause claim.

## Safety and scope

- Workflow-only instrumentation plus an ephemeral pytest plugin and its structural contract.
- Draft PR, main push, or manual dispatch only; no provider credentials or production permissions.
- 15-minute command timeout with 30-second hard-kill grace and 25-minute job bound.
- Pytest output goes directly to a regular file; there is no `tee`/pipe drain.
- Lifecycle JSONL, pytest log, read-only system snapshots, recorded SHA, exit status, and canonical SHA-256 manifest upload before failure propagation.
- This diagnostic branch intentionally starts from #1995 commit `385ab887`; it is not a replacement implementation PR.

## TDD evidence

RED commit: `1c1d26f3b` — the contract failed with `FileNotFoundError` because `.github/workflows/1995-node-diagnostic.yml` did not exist.

GREEN commit: `5b123325a`.

## Validation

- `pytest -q tests/test_unit_tests_workflow.py` — 20 passed.
- Durable plugin smoke — 13 JSONL records; all session, collection, node, setup/call/teardown start/finish records present.
- `pylint scripts/ci/pytest_lifecycle_jsonl.py tests/test_unit_tests_workflow.py` — 10.00/10.
- Exact six-file local selection — completed in 8m35s, 633 passed, 121 skipped, 5 failures caused by the current non-PDD local runtime's isolated released-checker venv lacking `tree_sitter`; no stall. This is an environment limitation, not issue-1995 evidence.

## Interpretation

Use the last unmatched lifecycle record only to locate the node and pytest phase active at termination. Do not infer causality without corroborating evidence.


gltanaka:
## RED/GREEN diagnostic evidence

RED was pushed first as `1c1d26f3b`:

```text
FAILED tests/test_unit_tests_workflow.py::test_issue_1995_node_diagnostic_workflow_contract
FileNotFoundError: [Errno 2] No such file or directory: '.github/workflows/1995-node-diagnostic.yml'
1 failed
```

GREEN at `5b123325a`:

```text
pytest -q tests/test_unit_tests_workflow.py
20 passed, 1 warning in 0.46s

pylint scripts/ci/pytest_lifecycle_jsonl.py tests/test_unit_tests_workflow.py
Your code has been rated at 10.00/10

lifecycle-smoke: 13 durable records; all required events present
```

The post-RED contract corrections did not weaken requirements: one inspects the lifecycle path in the step `env` where it is actually declared, and one uses pytest's real `pytest_collection` wrapper hook instead of the invalid assumed `pytest_collection_start` hook. The latter invalid name was caught by an executable plugin smoke before the GREEN commit.

gltanaka:
## Critical diagnostic review — PR #2107 @ `5b123325ab2b99c02a832cc0cdcacdec7cce8263`

**Verdict:** **DO NOT RUN**
**Disposition:** **NEVER MERGE** — this remains disposable diagnostic instrumentation only.
**Findings:** 6 total (2 P1, 4 P2)

### P1 (blocking)

- [P1] The pull-request checkout does not run the declared source — `.github/workflows/1995-node-diagnostic.yml:20-23`
  `actions/checkout` has no `ref`, so on `pull_request` it checked GitHub's synthetic merge ref, not exact head `5b123325a` and not underlying diagnostic subject `385ab887`. The current merge ref is `db0ea8e3a6295c8c8e8a32e6e336165064266104` (parents current main `0d529578a` and this head), and it differs from the reviewed head in imported production modules `pdd/sync_core/decommission.py` and `pdd/sync_core/verification.py`. `GITHUB_SHA`, `source-sha.txt`, the snapshots, and the artifact name consequently attest the synthetic merge, while the PR body claims preserved `385ab887` source. Evidence from this run cannot be attributed to the stalled run's code.
  **Suggested fix:** explicitly check out the event's immutable PR head SHA, record both checked HEAD and the intended production subject `385ab887`, and fail before pytest unless a path-scoped diff proves all production code and the six selected test files are identical to `385ab887` (allow only the reviewed diagnostic workflow/plugin/contract files).

- [P1] The diagnostic does not reproduce the original focused lane's prerequisite state — `.github/workflows/1995-node-diagnostic.yml:25-38`
  Run `29468333457` executed the six files after Node 22 setup, identity-bound Vitest and Playwright/Chromium manifests, `playwright install --with-deps`, Git configuration, and the protected Linux sandbox provisioning/smokes in `unit-tests.yml`. This workflow installs only Python dependencies and Bubblewrap. The mismatch is already material: the PR's local run reports 121 skips, versus 16 skips in the comparable hosted focused lane. Thus the same filenames and `--timeout=60` are not an equivalent selection, and a fast/quiet result could merely reflect missing toolchains or skipped paths.
  **Suggested fix:** reproduce the exact prerequisite/setup steps and environment from `385ab887` before invoking the unchanged six-file command; add an explicit collected/skip-count or node-list attestation so the run aborts rather than silently diagnosing a different selection.

### P2 (robustness / interpretation)

- [P2] Graceful timeout unwinding can erase the claimed unmatched phase — `scripts/ci/pytest_lifecycle_jsonl.py:81-108`
  Every phase emits `phase.finish` in `finally`. If GNU `timeout`'s SIGINT is handled by pytest/Python, the active hook unwinds and records an apparently matched finish even though the phase was interrupted; `node.finish` and `session.finish` may follow as well. The JSONL therefore reliably brackets hard death, but does not by itself distinguish normal completion from an interrupted phase. A last unmatched record exists only when unwinding cannot run (for example SIGKILL).
  **Suggested fix:** record hookwrapper outcome/exception classification and a distinct interrupted/error completion without raw exception payloads, and document that the last started phase—not necessarily an unmatched phase—is only a locator, never a causal diagnosis.

- [P2] The manifest is not guaranteed to describe the uploaded bytes when descendants survive — `.github/workflows/1995-node-diagnostic.yml:61-87`
  A detached descendant can retain the regular `pytest.log` descriptor after `timeout` returns—the lifecycle being investigated specifically involves surviving descendants. Such a process can append while `sha256sum` runs or after it finishes, so `SHA256SUMS` can disagree with the later uploaded artifact. No post-hash verification or immutable evidence snapshot closes that race.
  **Suggested fix:** quiesce the diagnostic process group and copy evidence into a closed staging directory before hashing, then verify the manifest immediately before upload; retain survivor metadata separately and fail closed if evidence changes.

- [P2] The public artifact contains unsanitized process command lines — `.github/workflows/1995-node-diagnostic.yml:51-59,73-79`
  `ps -ef` records complete argv for runner/test/surviving processes. The selected tests exercise authority/supervision tokens and adversarial subprocesses; even without provider secrets, publishing arbitrary argv in a repository artifact is broader than the stated safe read-only snapshots and can capture ephemeral capability material or injected values.
  **Suggested fix:** collect a minimal allowlisted process snapshot (PID/PPID/state/comm only, or rigorously redact arguments) and explicitly scan the evidence directory for known token/key patterns before hashing/upload.

- [P2] The contract test is structural and locks in the two blocking mistakes — `tests/test_unit_tests_workflow.py:252-323`
  The test proves strings, hook names, and `fsync` exist, but never verifies the checked-out commit, production-diff equivalence, original prerequisite parity, collected-node parity, JSONL parseability/order, interruption semantics, or manifest verification. It therefore passes (20/20 locally) while the hosted experiment is non-attributable and non-equivalent.
  **Suggested fix:** add semantic tests for pinned-source validation, exact setup/node parity, a real timeout/interruption lifecycle smoke, and manifest verification; do not treat source-string assertions as execution evidence.

### Validated properties / notes

- Exact diagnostic diff from `385ab887` is limited to the workflow, plugin, and workflow contract; the six selected test files are unchanged on the branch head.
- Permissions are minimal (`contents: read`), checkout credentials are not persisted, no provider credentials are requested, the pytest output is redirected to a regular file with no `tee` pipe, the command has 15-minute + 30-second kill bounds inside a 25-minute job bound, artifact naming is collision-resistant, and failure propagation is ordered after the always-upload step when the job itself is not externally canceled.
- Local validation at exact head: `pytest -q tests/test_unit_tests_workflow.py` → 20 passed; scoped pylint → 10.00/10. These checks do not cure the behavioral findings above.
- The hosted run had already started before this review completed. This review did not rerun, cancel, or mutate it. Do not interpret its artifacts as issue-1995 evidence unless the P1 provenance and equivalence problems are corrected in a new reviewed head.
- An unmatched lifecycle record can identify only the last observed node/phase. It cannot establish root cause.

gltanaka:
## Implementer response to critical diagnostic review

Re: https://github.com/promptdriven/pdd/pull/2107#issuecomment-4989116996

### P1 #1: PR checkout does not run declared source
Status: **FIXING**
Plan: check out the immutable PR head, record and verify checked HEAD, and fail closed unless every non-diagnostic path is byte-identical to intended subject `385ab8872411e740480bb164fb2b840b91f2624c`.

### P1 #2: prerequisite state is not equivalent
Status: **FIXING**
Plan: reproduce the exact Node 22, Vitest, Playwright/Chromium, git, Python, and protected-sandbox setup from the source workflow and attest collection/skip behavior before accepting evidence.

### P2 #1: graceful interruption can look like normal finish
Status: **FIXING**
Plan: inspect hook outcomes and emit explicit `phase.error`/`phase.abort` records; reserve `phase.finish` for normal completion and leave unmatched starts only for abrupt death.

### P2 #2: evidence can change after hashing
Status: **FIXING**
Plan: launch pytest in an attributable process group, terminate remaining group members, snapshot into a sealed directory only after quiescence, hash the sealed bytes, and verify the manifest immediately before upload.

### P2 #3: process argv are unsanitized
Status: **FIXING**
Plan: replace `ps -ef` with allowlisted PID/PPID/state/comm metadata and fail closed on evidence token/key-pattern scanning.

### P2 #4: contract locks structural mistakes
Status: **FIXING**
Plan: add executable semantic tests for provenance/diff validation, setup parity, JSONL ordering/interruption behavior, process snapshot safety, and manifest verification before updating implementation.

The invalid run `29478389206` was canceled and will not be interpreted. This remains diagnostic-only and **must never merge**.

gltanaka:
## Critical-review correction pushed

All six findings from https://github.com/promptdriven/pdd/pull/2107#issuecomment-4989116996 are addressed at exact head `5853bfcfbbb4cdace2681be35fad8952c84269ab`.

### TDD

Correction RED commit `36e2e2fb4` produced four semantic failures before implementation:

- missing immutable PR-head checkout/setup parity;
- missing path-scoped source verifier;
- failed phases incorrectly emitted `phase.finish`;
- missing evidence sealer/manifest verifier.

GREEN commits: `329973be6`, then `5853bfcfb` for the actionlint-discovered step-context correction.

### Validation

- `pytest -q tests/test_issue_1995_diagnostic.py tests/test_unit_tests_workflow.py` — 24 passed.
- Scoped pylint — 10.00/10.
- `actionlint .github/workflows/1995-node-diagnostic.yml` — clean.
- Exact local provenance check proves every changed path relative to `385ab8872411e740480bb164fb2b840b91f2624c` is one of eight diagnostic-only files.
- Executable lifecycle semantic test proves pass → `phase.finish`, assertion failure → `phase.error`, KeyboardInterrupt → `phase.abort` without exception payloads.
- Evidence test proves canonical manifest verification detects byte mutation and secret-pattern scanning rejects unsafe evidence.

Run `29479379192` was a zero-job workflow-parse failure caused by using the step-only `runner` context at job scope. It contains no diagnostic evidence and must not be interpreted. Actionlint caught and the final commit fixed that error before the new valid run.

The valid synchronize-triggered diagnostic is run `29479456504` at the exact final head. This PR remains **NEVER MERGE**. A last started/unmatched/error/abort record is only node/phase location evidence, never causality.

gltanaka:
## Critical exact-head re-review — PR #2107 @ `5853bfcfbbb4cdace2681be35fad8952c84269ab`

**Verdict: DO NOT RUN**
**Disposition: NEVER MERGE**
**Findings: 6 total (3 P1, 3 P2).**

I reached this verdict from the reviewed bytes and deterministic validation before looking at any node-lifecycle evidence. Run `29479456504` is an invalid setup-only failure; it never invoked pytest and produced no artifact, so it contains no issue-1995 diagnostic evidence.

### P1 (blocking)

- **The setup-parity runner cannot run in the clean hosted environment.** `scripts/ci/run_issue_1995_lane_step.py:10` imports `yaml`, but the workflow calls it for the Vitest step before the original `Install dependencies` step. A clean `actions/setup-python` Python has no PyYAML. Exact-head run `29479456504` confirms `ModuleNotFoundError: No module named 'yaml'` at the first imported setup step. Thus the original lane prerequisites are not reproduced and pytest is never reached.

- **The descendant-containment command is invalid on the target runner.** `.github/workflows/1995-node-diagnostic.yml:119-124` combines `systemd-run --scope --wait`. The Ubuntu 24.04/Noble `systemd-run(1)` contract explicitly says `--wait` may not be combined with `--scope`. Even after fixing PyYAML, this command fails before pytest, so the claimed cgroup quiescence and writer sealing are not executable. Use an actually supported transient-unit/controller pattern and test it on the hosted OS before another run.

- **The collection attestation hard-codes a platform-wrong count.** The exact comparable hosted source run `29461515588` completed the same six-file selection with `1 failed, 724 passed, 16 skipped`: 741 item outcomes. The hard-coded `--expected-collected 759` is the local macOS collect-only count (reproduced here), not established Linux-hosted parity. Therefore a source-equivalent hosted completion is falsely rejected. A canonical collected-node-id digest captured on the original hosted lane (or an authoritative Linux count) is required; a local count is not a valid attestation.

### P2 (robustness / posture)

- **“Always upload” is not true for prerequisite failures.** The live provenance files are created before setup, but sealing exists only inside the diagnostic step. When a prerequisite fails, the verify/upload steps run against a nonexistent sealed directory and no artifact is uploaded, exactly as run `29479456504` demonstrated. Add a final always-run sealer for whatever live evidence exists, then verify, upload, and only afterward propagate the original failure.

- **The never-merge posture is not enforced by triggers.** The workflow declares both `push: main` and unrestricted `workflow_dispatch`, and the job guard allows every non-PR event. An accidental merge would execute it on main. A disposable diagnostic should be draft-PR-only and fail closed on the intended PR/head; prose saying “never merge” is not a control.

- **The semantic tests still miss the hosted execution boundary.** They parse the current workflow/allowlist and assert command strings, but do not execute the setup-step runner in a clean pre-install Python, validate the `systemd-run` option combination on Linux, or derive the hosted node inventory. Consequently all 24 tests and actionlint pass while the workflow has two guaranteed pre-pytest failures and a false parity gate. This remains materially circular at the integration seams.

### Prior-finding audit / validated properties

- **Fixed in code:** immutable PR-head checkout; exact HEAD check; path-scoped byte gate to subject `385ab8872411e740480bb164fb2b840b91f2624c`; exact source workflow bytes retained; original action definitions and allowlisted setup/sandbox/smoke shell bodies selected; phase `finish/error/abort` classification without exception payloads; allowlisted PID/PPID/state/comm snapshots; pre/post-copy secret scan; canonical path inventory + SHA-256 verification; minimal `contents: read`; no persisted checkout credentials; regular-file pytest output; upload ordered before the explicit diagnostic-result propagation step.
- **Not proven executable:** setup parity, systemd cgroup quiescence, sealed artifact production/upload, and Linux selection parity, due to the blocking findings above.
- Exact diagnostic diff from subject remains limited to the eight allowlisted diagnostic files.

### Read-only validation

- `pytest -q tests/test_issue_1995_diagnostic.py tests/test_unit_tests_workflow.py` — **24 passed**
- scoped pylint — **10.00/10**
- `actionlint .github/workflows/1995-node-diagnostic.yml` — **clean**
- exact local provenance verifier — **pass**, eight allowlisted paths
- six-file local `--collect-only` — **759 on macOS**, demonstrating why that number cannot attest the Linux lane
- hosted run status/log inspection — **failure before pytest; no artifact**; no node evidence interpreted

Do not rerun this head. Do not merge this PR.

gltanaka:
## Implementer response to hosted-validation review

Re: https://github.com/promptdriven/pdd/pull/2107#issuecomment-4989306577

### P1 #1: clean host lacks PyYAML before setup runner
Status: **FIXING**
Plan: make the setup runner standard-library-only and add a clean isolated-Python execution test before any project install.

### P1 #2: invalid `systemd-run --scope --wait`
Status: **FIXING**
Plan: replace it with a supported transient service-unit command that synchronously waits, owns the full cgroup, kills remaining members, and proves the unit/cgroup inactive before sealing; enforce the exact supported option contract.

### P1 #3: platform-wrong collection count
Status: **FIXING**
Plan: use the evidence-backed exact hosted Linux outcome count of 741 and 16 skips, and attest the same completed-node outcome quantity rather than macOS collection count.

### P2 #1: setup failures produce no artifact
Status: **FIXING**
Plan: initialize minimal live evidence immediately after checkout and move sealing to an always-run finalizer that records step outcomes, closes/snapshots whatever exists, verifies, and uploads before final failure propagation.

### P2 #2: never-merge posture is not enforced
Status: **FIXING**
Plan: remove push and manual triggers; permit only draft pull-request events targeting main, with exact-head and intended-PR fail-closed guards.

### P2 #3: tests miss hosted seams
Status: **FIXING**
Plan: add clean-host runner execution, setup-failure fallback sealing, supported systemd service option, Linux outcome-count, and draft-only trigger semantic tests before the workflow correction.

Run `29479456504` failed before pytest and produced no artifact. It is setup-only failure evidence and will not be interpreted as issue-1995 evidence. This PR remains **NEVER MERGE**.

gltanaka:
## Hosted-validation correction pushed

All findings from https://github.com/promptdriven/pdd/pull/2107#issuecomment-4989306577 are addressed at exact head `1939bfa7d4a47f9768cede632d481071e38a453f`.

### TDD

- RED `d5889426c`: four failures covering non-draft triggers, non-stdlib setup runner, collection-vs-outcome parity, and missing always-finalizer contract.
- GREEN `1939bfa7d`.

### Changed execution boundaries

- Draft PR #2107 only; no push or manual trigger, exact immutable PR-head checkout.
- Setup runner is standard-library-only and extracts all eight authoritative unit-lane shell commands; its extracted scalars match PyYAML's authoritative values byte-for-byte.
- Pytest runs synchronously in a supported transient `.service` using `systemd-run --wait --service-type=exec`; no invalid scope/wait combination.
- Always-finalizer records step outcomes, enumerates only cgroup PIDs, kills/stops all service cgroup members, proves inactive/failed plus empty/removed `cgroup.procs`, then seals.
- Setup failures with no lifecycle still produce minimal sealed, verified evidence.
- Selection parity counts one terminal outcome per node and uses the evidence-backed hosted values: 741 outcomes, 16 skips.

### Validation

- Semantic/contract tests: 28 passed.
- Clean-host seam: `python -I` runner successfully executed the pre-install git-configuration source step with isolated HOME.
- All eight stdlib-extracted setup commands equal the authoritative parsed workflow scalars.
- Setup-failure fallback evidence seals and verifies.
- Scoped pylint: 10.00/10.
- Actionlint: clean.
- Exact path-scoped provenance gate: pass; eight diagnostic-only paths differ from `385ab8872411e740480bb164fb2b840b91f2624c`.

Automatic synchronize run: `29480690586`. Do not manually rerun. Prior run `29479456504` remains setup-only failure evidence and contains no issue-1995 test evidence. This PR is **NEVER MERGE**.

gltanaka:
## Final critical exact-head review — PR #2107 @ `1939bfa7d4a47f9768cede632d481071e38a453f`

**Verdict: DO NOT RUN**
**Disposition: NEVER MERGE**
**Findings: 5 total (2 P1, 3 P2).**

This verdict is based on the reviewed bytes and deterministic/source evidence. Automatic run `29480690586` was already in progress and had reached setup/smoke work when reviewed. I did not cancel, rerun, download, or interpret any node-lifecycle evidence from it.

### P1 (blocking)

- **The pytest stream is not written to the claimed regular evidence file.** `.github/workflows/1995-node-diagnostic.yml:119-145` redirects the `systemd-run` client's stdout/stderr to `pytest.log`, but the transient `.service` is started without `--pipe` and without `StandardOutput=`/`StandardError=` properties. On Ubuntu Noble, service stdout/stderr go to journald by default; `systemd-run --wait` only prints its own terse unit result. The Noble `systemd-run(1)` contract says `--pipe` is what makes the service inherit the client's descriptors, while `systemd-journald(8)` states service output is journal-connected by default. Therefore `pytest.log` will not contain pytest's `-vv --capture=tee-sys` stream, and the comment that the service “inherits it directly” is false. This loses a required corroborating artifact at the exact boundary being diagnosed. Use a supported no-pipe service property such as explicit file/append standard-output and standard-error settings (or another deterministically tested regular-file design), and test the service's bytes on Noble.

- **The `741 outcomes / 16 skips` gate is not evidence-backed for the diagnostic subject.** Those values come from hosted run `29461515588` at `57a28ff70a080478b0fae43aefe1bc8ebab72a10`, but the attributed subject is later commit `385ab8872411e740480bb164fb2b840b91f2624c`. Between those commits, one of the six selected files, `tests/test_sync_core_supervisor.py`, changed by **+1770/-91** and gained at least **11 test functions plus parameterized cases**. Thus a completed exact-subject run has never established 741 terminal nodes or 16 skips; the only exact-subject focused run was the unresolved/canceled quiet run. The synthetic attestation test manufactures 741 records and cannot convert an older-head summary into current-head parity. A frozen exact-subject Linux node inventory/baseline is needed, or the first run must record rather than reject previously unknown exact-subject counts.

### P2 (robustness / test independence)

- **“After any setup failure” still excludes checkout failure.** Initialization precedes checkout, but finalization invokes repository scripts. If checkout itself fails, `scripts/ci/seal_issue_1995_evidence.py` is unavailable, so the minimal run evidence cannot be sealed, verified, or uploaded. Either provision the finalizer independently of checkout or narrow the guarantee.

- **A failed final verification does not prevent upload.** Both verification and upload use independent `if: always()` guards. If `verify` detects post-seal mutation/inventory failure, the next step still uploads those rejected bytes. Give verification an id and gate upload on its success (while preserving the setup-failure fallback).

- **The tests remain circular at both blocking seams.** The workflow contract only asserts that a shell redirection string exists; it never executes a Noble transient service and verifies child bytes landed in the regular file. The parity test constructs exactly the hard-coded 741/16 records rather than deriving or checking an exact-subject hosted inventory. All 28 tests can pass while both core claims are false.

### Verified properties

- Draft pull request #2107 is the only trigger; no push/manual trigger exists, and the job guard pins draft PR #2107.
- Immutable PR-head checkout and path-scoped source provenance pass; exactly eight diagnostic paths differ from subject `385ab887`.
- The setup runner is standard-library-only, and all eight extracted shell scalars are byte-equal to the authoritative parsed source workflow.
- `systemd-run --wait --service-type=exec` is a supported Noble service form; no invalid `--scope --wait` combination remains.
- The finalizer uses allowlisted PID/PPID/state/comm snapshots, cgroup-scoped kill/stop, inactive/failed-state checking, and root `cgroup.procs` emptiness/removal checking before sealing.
- Phase lifecycle records distinguish normal `finish`, assertion `error`, and interrupt `abort` without exception payloads. An unmatched/error/abort record remains location evidence only, never causality.
- Canonical generated path inventory, SHA-256 verification, secret-pattern scans, read-only sealing, minimal permissions, exact-head artifact naming, and explicit diagnostic status propagation are present, subject to the P2 upload gate above.

### Read-only validation

- `pytest -q tests/test_issue_1995_diagnostic.py tests/test_unit_tests_workflow.py` — **28 passed**
- scoped pylint — **10.00/10**
- actionlint — **clean**
- exact provenance verifier — **pass**, eight allowed diagnostic paths
- stdlib extractor vs parsed workflow — **8/8 byte-equal**
- worktree HEAD and PR head — exact `1939bfa7d4a47f9768cede632d481071e38a453f`

Do not manually rerun this head. Do not merge this PR.

gltanaka:
## Implementer response to final critical review

Re: https://github.com/promptdriven/pdd/pull/2107#issuecomment-4989421015

### P1 #1: service does not own pytest log descriptor
Status: **FIXING**
Plan: run an explicit service-side wrapper that opens the regular pytest log and `exec`s the bounded pytest command; add a functional child-process capture test proving bytes land in that file without a pipe.

### P1 #2: historical 741/16 gate is not exact-subject evidence
Status: **FIXING**
Plan: remove historical constants. The single pytest collection will durably record its own exact node inventory identity before execution; attestation will validate six-path membership, inventory digest/count/per-file counts, and completed terminal coverage against that same collection.

### P2 #1: checkout failure lacks scriptless finalization
Status: **FIXING**
Plan: add an inline standard-library fallback that seals and verifies the minimal pre-checkout evidence when repository scripts are unavailable.

### P2 #2: rejected evidence can still upload
Status: **FIXING**
Plan: give final verification a stable id and upload only when its outcome is success.

### P2 #3: seam tests remain circular
Status: **FIXING**
Plan: add functional service-wrapper redirection, scriptless fallback seal/verification, corruption rejection/upload-gate, and self-collected inventory/coverage tests before implementation.

Automatic run `29480690586` was canceled. It will not be interpreted. No manual rerun will be triggered. This remains **NEVER MERGE**.

gltanaka:
## Self-attesting evidence correction pushed

All findings from https://github.com/promptdriven/pdd/pull/2107#issuecomment-4989421015 are addressed at exact head `4235d46be7a4bf10b9d458884466e340e8f96030`.

### TDD

- RED `8d8b1c77e`: four failures for historical constants, absent service-owned capture, absent pre-checkout fallback, and ungated upload verification.
- GREEN `4235d46be`.

### Evidence boundaries

- The transient service executes `run_issue_1995_pytest_service.py`; that service-side process opens `pytest.log` as a regular mode-0600 file, duplicates it onto stdout/stderr, and `exec`s the bounded pytest command. The `systemd-run` client is no longer redirected and no pipe/tee process is used.
- Historical 741/16 and local 759 constants are removed. During the single pytest collection, the lifecycle plugin fsyncs the unique sorted node IDs, count, SHA-256 identity, and per-file counts before execution.
- Final attestation requires exactly the six selected paths and validates terminal reports against that same collected inventory. A normally finished session requires complete coverage; an abruptly bounded run may retain a validated partial subset.
- Initialization writes an inline standard-library fallback sealer before checkout. Checkout failure can therefore still seal and verify minimal evidence without repository files.
- Upload is gated on `steps.verify_evidence.outcome == 'success'`; rejected bytes are not uploaded.

### Validation

- Semantic/contract tests: 30 passed.
- Functional service-wrapper capture: child stdout and stderr both landed in the service-opened regular file; parent capture remained empty.
- Functional plugin inventory: count, sorted node IDs, and digest verified from a real pytest subprocess.
- Functional own-inventory attestation: valid abrupt partial accepted; finished incomplete rejected; finished complete accepted.
- Functional inline fallback: sealed and verified checkout-failure evidence; post-seal corruption rejected.
- Scoped pylint: 10.00/10.
- Actionlint: clean.
- Exact provenance: pass; nine diagnostic-only paths differ from `385ab8872411e740480bb164fb2b840b91f2624c`.

Automatic synchronize run: `29481549047`. No manual rerun. Canceled run `29480690586` is not interpreted. This PR remains **NEVER MERGE**.

gltanaka:
## Release-gate critical exact-head review — PR #2107 @ `4235d46be7a4bf10b9d458884466e340e8f96030`

**Verdict: DO NOT RUN**
**Disposition: NEVER MERGE**
**Findings: 3 total (1 P1, 2 P2).**

This verdict comes from reviewed bytes and deterministic local reproductions. Automatic run [29481549047](https://github.com/promptdriven/pdd/actions/runs/29481549047) had completed all setup/sandbox/signer smokes and had entered the bounded diagnostic when status was inspected. I did not cancel, rerun, download, mutate, or interpret node-level evidence from it.

### P1 — blocking

- **Abrupt teardown termination is falsely attested as a complete run.** `scripts/ci/attest_issue_1995_lifecycle.py:58-73,87-107` defines terminal coverage from call-phase (or any skipped) `node.report` records and computes `complete = set(terminal) == set(nodeids)` independently of `session_finished` and `node.finish`. If every call reports and the final node then stalls in teardown until the emergency timeout, the attester returns success with `"session_finished": false, "complete": true`. This is exactly a lifecycle boundary the diagnostic is meant to distinguish, so the sealed summary can misclassify a partial abrupt timeout as full completion. Deterministic six-path reproduction at this head: exit 0, `terminal_count: 6`, `complete: true`, no `session.finish`. **Required correction:** derive execution completion from one validated `node.finish` per collected node (and require `session.finish` for overall complete), retaining partial coverage explicitly without calling it complete.

### P2 — correctness / hardening

- **A normally finished session with a setup failure is rejected as lacking terminal coverage.** `scripts/ci/attest_issue_1995_lifecycle.py:66-69,91-93` ignores a failed setup report because it is neither skipped nor a call report. Pytest can legitimately finish after setup failure without a call report. Deterministic reproduction with five passed calls, one failed setup, teardown report, and `session.finish`: exit 1, `completed session lacks terminal coverage for every node`. This violates the claimed successful full-terminal-coverage contract and can discard otherwise valid diagnostic evidence. Terminal result classification must account for setup/teardown failures and skips without confusing result availability with node lifecycle completion.

- **The service wrapper does not enforce the claimed mode-0600 regular-file boundary for a pre-existing path.** `scripts/ci/run_issue_1995_pytest_service.py:22-33` uses `O_CREAT|O_TRUNC` without `O_NOFOLLOW`, regular-file `fstat`, or `fchmod`. On a fresh absent path it creates the expected file, but a deterministic pre-existing symlink reproduction succeeds, truncates/writes its target, and leaves target mode `0644`. The current private runner directory limits exposure, so this is hardening rather than the primary blocker; nevertheless the wrapper's stated assurance should fail closed with no-follow/exclusive creation or verify the opened descriptor is regular and force mode 0600.

### Verified properties

- Exact worktree and PR head match `4235d46be7a4bf10b9d458884466e340e8f96030`; the worktree is clean.
- Draft PR #2107 targeting main is the only trigger; the job guard requires draft PR #2107. There is no push, merge/closed, or manual-dispatch trigger.
- Checkout pins the immutable event head, persisted credentials are disabled, and source provenance permits only the nine diagnostic paths relative to subject `385ab8872411e740480bb164fb2b840b91f2624c`.
- The service-side wrapper itself opens stdout/stderr and execs GNU timeout/pytest; there is no client redirect or pipe. Emergency bound is 900 seconds, SIGINT then 30-second hard-kill grace, inside a 35-minute job.
- The plugin appends, flushes, and fsyncs each JSONL record; it emits a sorted unique full node inventory, count, SHA-256, and per-file map before node execution.
- Checkout-less inline sealing is executable; canonical sealing verifies hashes/path inventory, and upload is gated on successful final verification.
- The finalizer kills/stops the transient service cgroup, waits for inactive/failed state, checks root `cgroup.procs`, records bounded process metadata, then seals.
- Setup commands are extracted from the byte-verified source workflow with a stdlib-only runner; action versions and environment setup match that source lane.
- A last unmatched/error/abort lifecycle record identifies only the last observed node/phase, never root cause.

### Read-only validation

- `pytest -q tests/test_issue_1995_diagnostic.py tests/test_unit_tests_workflow.py` — **30 passed**
- Scoped pylint — **10.00/10**
- `actionlint .github/workflows/1995-node-diagnostic.yml` — **clean**
- Synthetic abrupt-after-all-calls attestation — **incorrectly exit 0 / complete true**
- Synthetic completed setup-failure attestation — **incorrectly exit 1**
- Pre-existing symlink log-path wrapper check — **incorrectly follows link and preserves 0644 target**

Do not interpret run `29481549047` as release-unblocking issue-1995 evidence at this head. Do not rerun this head. Never merge this PR.

gltanaka:
## Implementer response to release-gate review

Re: https://github.com/promptdriven/pdd/pull/2107#issuecomment-4989535926

### P1: abrupt teardown can appear complete
Status: **FIXING**
Plan: require `session.finish` plus exactly one `node.finish` for every collected node before `complete:true`; a teardown kill remains explicitly partial even when every call report exists.

### P2 #1: setup failure lacks truthful terminal classification
Status: **FIXING**
Plan: classify node results from pytest report order: setup failure/skip terminates outcome selection, otherwise call result, with teardown failure overriding as applicable; lifecycle completion remains independently based on node/session finish records.

### P2 #2: log open follows symlink and preserves weak mode
Status: **FIXING**
Plan: use `O_NOFOLLOW|O_CREAT|O_TRUNC` with safe flags, verify the descriptor is regular, force `fchmod(0600)`, and fail closed for symlinks; add absent, preexisting-0644, and symlink tests.

Canceled run `29481549047` will not be interpreted. No manual rerun. This PR remains **NEVER MERGE**.

gltanaka:
## Truthful completion correction pushed

All findings from https://github.com/promptdriven/pdd/pull/2107#issuecomment-4989535926 are addressed at exact head `174bc420981b60c9c8508743b0715f6f81e2325d`.

### TDD

- RED `4e2d326b2`: deterministic teardown-kill was falsely complete, finished setup-failure/skip was rejected, and preexisting mode-0644 log remained weak.
- GREEN `174bc4209`.

### Semantics

- `complete:true` now requires `session.finish`, exactly one `node.finish` for every collected node, and one truthful terminal outcome for every node.
- A hard kill in teardown remains partial even when every call report exists.
- Terminal results follow pytest report semantics: setup failure/skip terminates nodes that never reach call; call supplies the ordinary result; teardown failure overrides earlier outcomes.
- The service wrapper opens with `O_NOFOLLOW|O_CREAT|O_TRUNC|O_CLOEXEC|O_NONBLOCK`, rejects non-regular descriptors, forces `fchmod(0600)`, duplicates stdout/stderr, then execs.

### Validation

- Full diagnostic semantic/contract matrix: 33 passed.
- Real pytest setup-failure + setup-skip session: complete evidence accepted with `{failed: 1, skipped: 1, passed: 4}`.
- Real pytest process killed after a passed call entered infinite teardown: terminal count 1, node-finish count 0, no session finish, `complete:false`.
- Service capture: absent path mode 0600; preexisting 0644 file truncated and forced to 0600; symlink rejected without changing its target.
- Scoped pylint: 10.00/10.
- Actionlint: clean.

Automatic synchronize run: `29482280364`. No manual rerun. Canceled run `29481549047` is not interpreted. This PR remains **NEVER MERGE**.

gltanaka:
## Final truth-gate exact-head review — PR #2107 @ `174bc420981b60c9c8508743b0715f6f81e2325d`

**Verdict: RUN DIAGNOSTIC** — allow the single automatic synchronize run already in progress to continue; do not manually rerun or duplicate it.

**Disposition: NEVER MERGE.** This PR is disposable diagnostic instrumentation only.

**Findings: P1 = 0, P2 = 0.**

### Prior blocking-finding audit

- **Truthful completion:** fixed. `complete:true` requires a durable `session.finish`, exactly one `node.finish` for every node in the authenticated collection inventory, and terminal outcome coverage for that exact node set. A completed session missing any of those records fails attestation; an abrupt run remains explicitly partial.
- **Setup/skip/failure semantics:** fixed. Setup failure and setup skip are terminal without requiring a call report; call supplies the ordinary outcome; teardown failure overrides an earlier outcome. The real-pytest setup failure/skip test attests complete with `{failed: 1, passed: 4, skipped: 1}`.
- **Abrupt teardown kill:** fixed. A real process killed after its passed call entered an infinite teardown has a call report but no `node.finish` or `session.finish`, and attests `complete:false`.
- **Service-owned regular log:** fixed. The wrapper uses `O_NOFOLLOW|O_CREAT|O_TRUNC|O_CLOEXEC|O_NONBLOCK`, validates the opened descriptor with `fstat` as regular, forces mode 0600, `dup2`s it onto stdout/stderr, and `exec`s the bounded command. Functional tests cover absent files, safe truncation/mode correction of a pre-existing regular file, and symlink rejection without target mutation. There is no shell pipe or client-side tee process.

### Provenance, setup, collection, finalization, and upload gates

- Exact PR/worktree head is `174bc420981b60c9c8508743b0715f6f81e2325d`; the worktree is clean.
- The path-scoped verifier passes against subject `385ab8872411e740480bb164fb2b840b91f2624c`; only the nine explicit diagnostic workflow/script/test paths differ.
- Trigger posture is draft PR #2107 targeting `main` only; immutable event-head checkout, read-only contents permission, and no persisted checkout credentials remain enforced.
- The standard-library setup runner imports the byte-verified source lane's eight allowlisted setup/sandbox/smoke commands; the single six-file pytest selection records and fsyncs its own sorted unique node inventory, digest, count, and per-file map before execution.
- Finalization is always-run, kills/stops and checks the transient service cgroup before sealing, records allowlisted process metadata, scans for token/key patterns, creates a canonical SHA-256 inventory, and verifies sealed bytes. Checkout failure retains the independent fallback sealer. Upload is collision-safe and gated on successful sealed-evidence verification before diagnostic failure propagation.
- A last started/unmatched/error/abort record is only a node/phase locator, never evidence of root cause.

### Read-only validation

- `pytest -q tests/test_issue_1995_diagnostic.py tests/test_unit_tests_workflow.py` — **33 passed**.
- Scoped pylint — **10.00/10**.
- `actionlint .github/workflows/1995-node-diagnostic.yml` — **clean**.
- Exact-head source verifier — **pass**, nine allowlisted diagnostic-only paths.

Run [29482280364](https://github.com/promptdriven/pdd/actions/runs/29482280364) was inspected only for setup/status after this source verdict: exact head, in progress, protected sandbox and cgroup smoke successful, protected pytest smoke in progress, diagnostic not yet started. No node-lifecycle evidence was inspected or interpreted.

Let that one automatic run finish and verify its sealed provenance/attestation before interpreting any node location. Do not merge this PR.

gltanaka:
## Exact-head lifecycle diagnostic result — complete, no stall

Approved diagnostic-only PR #2107 ran once at exact head `174bc420981b60c9c8508743b0715f6f81e2325d`, byte-gated to subject `385ab8872411e740480bb164fb2b840b91f2624c`.

Run: https://github.com/promptdriven/pdd/actions/runs/29482280364

Verified evidence:
- diagnostic step: 2026-07-16T08:10:41Z–08:21:28Z (10m47s)
- manifest, lifecycle log, pytest log, provenance, and attestation independently passed SHA-256 verification
- complete session: 759 collected / 759 finished; 738 passed, 16 skipped, 5 failed
- no unmatched node or phase and no whole-selection stall
- artifact: `issue-1995-29482280364-1-174bc420981b60c9c8508743b0715f6f81e2325d`
- manifest SHA-256: `955bfcb8f96a8dcb37b9cfa1f820945f4acc65b68eda4f8bee1ce4c93a64d9dd`

All five failures are in `tests/test_sync_core_supervisor.py`:
- `test_held_namespace_scan_memfd_is_sealed_and_only_transport_fds_inherit`
- oversized/malformed-frame case of `test_namespace_scanner_rejects_exact_eof_trailing_oversized_and_malformed_frames`
- `test_namespace_holder_unmount_payload_and_nsenter_argv_are_exact`
- `test_external_holder_pidfd_rejects_reuse_then_terminates_and_reaps`
- `test_real_linux_playwright_descriptor_exact_chain[fd-only-namespace-holder-cleanup]`

Shared signal: `diagnostic transport invariant`, repeatedly rooted in `OSError: [Errno 9] Bad file descriptor`; the real Linux FD-only case also reports cleanup expiry with a surviving external holder.

Interpretation is deliberately bounded: this disproves a deterministic stall at exact subject 385 and identifies an FD-transport failure cluster. It does **not** prove the causal root of run 29468333457's prior 43m quiet step because #2107 adds outer transient-systemd containment. Reproduce this cluster in the unchanged standard hosted lane before authorizing a production source fix.

PR #2107 is diagnostic-only and will be closed unmerged.

gltanaka:
Closing unmerged as required: diagnostic evidence is complete and recorded on #1995, #2091, and #2083. This workflow must never merge.

PR #2111, opened 2026-07-16
fix(sync): preserve lifecycle child timeout status

gltanaka:
## Stacked dependency

This is a bounded correction for the unresolved lifecycle timeout-classification finding on #1995. It is stacked on exact #1995 head `747771ef59789784320813f800f3af3a3788a12f`; merge into that branch before integrating #1995.

## Root cause and fix

The trusted candidate transaction wrapper used exit 1 for its internal child deadline while `run_lifecycle_matrix` classifies a receipt-less transaction as a timeout only when the wrapper status is 124. The result was a generic failure instead of closed timeout evidence.

The wrapper now exits 124 only for its three exact internal deadline paths: before child spawn, after selector-loop expiry, and after `child.wait` expiry. Cleanup occurs before exit for spawned children. Output-limit, install/proof, mutation, receipt, survivor, and generic failure paths are unchanged and remain non-timeout.

## TDD and validation

- red commit `5a7228a36`: real wrapper returned 1 instead of 124
- green commit `f06b1738d`: exact timeout propagation plus certificate-boundary test
- targeted timeout tests: 2 passed
- full lifecycle scenario module: passed (33 passed, 2 skipped)
- production pylint: passed
- py_compile and `git diff --check`: passed
- independent critical review: GO; no findings

Residual risk: the real subprocess test forces the pre-spawn deadline path; selector-loop and wait-expiry paths are statically identical and independently reviewed but not separately forced in subprocess tests.

Closes no issue; prerequisite correction for #1995/#2097 release integration.


gltanaka:
Superseded after exact patch/test audit; closing unmerged.

- Production fix at this head (`f06b1738`) has stable patch ID `e1479cfd`, exactly present in current #1995 commit `9c2b596020`.
- The original child-timeout test remains in #1995 by exact name.
- #1995 additionally strengthens coverage for supervisor grace, lifecycle/toolchain boundaries, reserved exit 124, real supervisor termination, and private deadline behavior.
- This PR is dirty against its historical stack base and must not be merged or rebased blindly.

Current replacement authority: #1995 head `1be64ca21c507d3f9798cf879ce3165659ffae8f` plus its eventual refreshed hosted validation.

PR #2112, opened 2026-07-16
diagnostic(sync): seal Vitest pressure evidence

gltanaka:
## Diagnostic source only — do not merge

Tracks #2083.

This draft contains the immutable, reviewed diagnostic payload only. It has no GitHub workflow and cannot execute by itself.

Exact source SHA: `829fb7e2818bb326d31a06cb16bdcb698576b7e6`

Exact changed paths:

- `scripts/ci/run_issue_2083_vitest_pressure.py`
- `scripts/ci/issue_2083_vitest_barrier.py`
- `tests/test_issue_2083_vitest_pressure_diagnostic.py`

Content SHA-256:

- runner: `1103102193911a8b28588d4efb44eb6a83520c945e94ab7e65c2dfea0d99c469`
- barrier: `b63e3ce1f2d427ca8588d300b7c85144d23350d46d2cbda6bc1dec8c4b442d02`
- tests: `fb146ad1edbaf3f7fb1267dc7fb0ceee93a94610a9ae1cccfb0f4f2cccfe9e75`

Validation: 35 tests, pylint 10/10, compile/diff checks, independent critical review 0 P1/0 P2.

A separately reviewed temporary dispatcher on `main` will hardcode this SHA and hashes, execute once with read-only permissions, upload sealed evidence before status propagation, and then be removed. This PR remains draft and unmerged.

PR #2108, opened 2026-07-16
fix(sync): reuse proven runtime copy in trusted closure

gltanaka:
## Stacked dependency

Diagnostic-only prerequisite for #2083, stacked on #2097 exact head `61c09edb61ee4945440fb70b082973c5524e1a6b`, whose base is #1995. Do not merge before the stack is integrated.

## Problem and scope

Unrelated release-gate runs 29471519534 and 29481807534 failed only the real Vitest installed toolchain test with the same trusted `sandbox-error`, exit 125, and diagnostic hash `dbde53cd47c15502417aed17a03d99f400077460a4599b2dd1140001e943a540`. The latter failed in 4.871 seconds. Existing typed termination collapsed construction, handoff, postprocessing, cleanup, and process-cleanup faults into the same kind and could discard available kernel telemetry, so neither a retry nor a timeout change was evidence-safe.

This PR adds diagnostic structure only:

- allowlisted, supervisor-derived infrastructure failure phases
- explicit nested cleanup combinations for ordinary and delegated Playwright paths
- retained monotonic cgroup event telemetry on later failed-closed conversion
- Vitest detail emits only the enum phases and kernel counters; candidate output remains hash-only
- absent/invalid phase evidence becomes `unknown`

There are no retry, deadline, timeout, return-code, control-flow, or evidence-outcome changes.

## Validation

- focused diagnostic/integration tests: 21 passed
- full touched suites: 423 passed, 49 skipped
- final critical review at exact `227201f5128137498a2d4786b358802b4f7b6ffb`: GO, no findings
- production pylint: 10.00/10
- py_compile and `git diff --check`: passed

## Required hosted evidence

Run the exact hosted Linux workflow once at `227201f5128137498a2d4786b358802b4f7b6ffb`. Use the emitted structured phase to authorize only a proven phase-specific correction. Do not bypass the gate or add a generic retry.

Closes no issue; diagnostic prerequisite for #2083.
## Root cause and bounded fix

Hosted diagnostic run 29498500365 narrowed the stable Vitest hash to `descriptor-plan-construction` / `source-path-collision-planning`. The inferred runtime first consumes and records the exact immutable descriptor-copy proof; the trusted ELF closure later revisits the same protected source and destination and was rejected as a conflicting binding.

[...]

PR #2123, opened 2026-07-17
fix(sync): preserve Vitest worker wasm guard

### LINKED ISSUE #2083: CI: real Vitest sandbox intermittently aborts without actionable trusted diagnostic

gltanaka:
## Root cause

The protected Linux Vitest coordinator starts with `--disable-wasm-trap-handler`, but Vitest 4.1.10 creates fork workers without retaining that checker-owned Node flag. Repeated exact invocations therefore repeatedly start unguarded V8 runtimes under the sandbox address-space policy; the sealed #2083 matrix observed their identical pre-reporter SIGABRT. This patch passes the existing guard through Vitest’s worker `execArgv` and renders protected paths as operands so a dash-leading path cannot become a competing CLI control.

No maxWorkers, retry, timeout, memory/process-limit, gate, or reporter change. FD 198 remains unavailable in the worker.

## TDD

- Immutable subject: `bd4a250420c3b7aaa50bab6fd73aded271115c71`
- Subject RED: `be80ce152` (real Vitest worker fails to retain guard)
- Subject GREEN: `c48227da8`
- Isolated integration base: `0c3568c656dce2c3bda445eb9ce5f597c5a65c70`
- Integration RED/GREEN: `dbd2d0179` / `d8f22f830`

## Validation

- exact standalone Node 22.23.1 + Vitest 4.1.10 worker lifecycle: PASS
- full `tests/test_sync_core_runner_vitest.py`: PASS
- isolated installed-wheel real worker lifecycle: PASS
- `pylint pdd/sync_core/runner.py`: 10.00/10
- py_compile and diff check: PASS

Fixes #2083

### DIFF
--- a/pdd/sync_core/runner.py
+++ b/pdd/sync_core/runner.py
@@ -4414,6 +4414,11 @@ def _vitest_environment(home: Path) -> dict[str, str]:
     }
 
 
+def _vitest_path_operand(path: PurePosixPath) -> str:
+    """Render one protected path so Vitest cannot parse it as a control."""
+    return "./" + path.as_posix()
+
+
 def _vitest_result(
     root: Path, output: Path, returncode: int, expected: tuple[str, ...] | None
 ) -> tuple[EvidenceOutcome, str, tuple[str, ...]]:
@@ -4650,9 +4655,14 @@ def _run_vitest(
             *( ("--disable-wasm-trap-handler",) if sys.platform.startswith("linux") else () ),
             str(phase_toolchain.entrypoint),
             "run",
-            *(path.as_posix() for path in paths),
             f"--config={root / config_path}",
             f"--reporter={reporter}",
+            *(
+                ("--execArgv=--disable-wasm-trap-handler",)
+                if sys.platform.startswith("linux")
+                else ()
+            ),
+            *(_vitest_path_operand(path) for path in paths),
         ]
         digest = hashlib.sha256(json.dumps(command, separators=(",", ":")).encode()).hexdigest()
         before = _validator_tree_identity(root)
--- a/tests/test_sync_core_runner_vitest.py
+++ b/tests/test_sync_core_runner_vitest.py
@@ -2267,6 +2267,10 @@ def test_vitest_linux_command_binds_wasm_guard(tmp_path: Path, monkeypatch: pyte
 
     assert execution.outcome is EvidenceOutcome.PASS
     assert observed[0][1] == "--disable-wasm-trap-handler"
+    assert observed[0][-2:] == [
+        "--execArgv=--disable-wasm-trap-handler",
+        "./tests/widget.test.ts",
+    ]
     assert proofs[0][0].descriptor_identity == _load_vitest_toolchain_descriptor(
         root, config
     ).identity
--- a/tests/test_sync_core_runner_vitest.py
+++ b/tests/test_sync_core_runner_vitest.py
@@ -1700,6 +1700,46 @@ def test_real_vitest_runs_copied_entrypoint_without_candidate_result_access(
     assert executions[0].outcome is EvidenceOutcome.PASS, executions[0].detail
 
 
+@pytest.mark.skipif(
+    not os.environ.get("PDD_REAL_VITEST_TOOLCHAIN_MANIFEST"),
+    reason="requires provisioned real Vitest toolchain",
+)
+def test_real_vitest_worker_inherits_linux_wasm_address_guard(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Every forked Vitest runtime must retain the coordinator's V8 guard."""
+    manifest = Path(os.environ["PDD_REAL_VITEST_TOOLCHAIN_MANIFEST"])
+    roles = json.loads(manifest.read_text(encoding="utf-8"))["roles"]
+    root, _commit = _repository(tmp_path)
+    (root / "tests/widget.test.ts").write_text(
+        "import fs from 'node:fs';\n"
+        "import { expect, test } from 'vitest';\n"
+        "test('worker retains checker runtime guard', () => {\n"
+        "  expect(process.execArgv).toContain('--disable-wasm-trap-handler');\n"
+        "  expect(() => fs.fstatSync(198)).toThrow();\n"
+        "});\n",
+        encoding="utf-8",
+    )
+    _git(root, "add", ".")
+    _git(root, "commit", "-q", "-m", "probe forked Vitest runtime")
+    monkeypatch.setattr(runner_module.sys, "platform", "linux")
+
+    execution, identities = _run_vitest(
+        root,
+        (PurePosixPath("tests/widget.test.ts"),),
+        30,
+        RunnerConfig(
+            vitest_command=(roles["launcher"], roles["entrypoint"]),
+            vitest_toolchain_manifest=manifest,
+        ),
+    )
+
+    assert execution.outcome is EvidenceOutcome.PASS, execution.detail
+    assert identities == (
+        "tests/widget.test.ts::worker retains checker runtime guard",
+    )
+
+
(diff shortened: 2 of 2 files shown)


gltanaka:
## Investigation and sibling audit

Sealed run 29529944426 was re-downloaded and verified: results `621f9e8b…`, manifest `046a83c0…`, toolchain `9e10734a…`; exact Node 22.23.1 / npm 10.9.8 / Vitest 4.1.10. The 2/3 serial, 3/6 width-2, and 5/12 width-4 failures were identical pre-reporter SIGABRT (`bb513afd…`) with zero cgroup OOM/OOM-kill/pids-max deltas.

History reviewed: PR #1790’s maxWorkers/V8/UV attempts, PR #2108’s trusted lifecycle/runtime-copy work, PR #2112’s immutable diagnostic, and the later #2121 address-space workaround. `maxWorkers=1` does not change the exact one-file worker count and resource/retry/gate/reporter changes were rejected.

Causal lifecycle gap: the coordinator-only `--disable-wasm-trap-handler` was not present in the real Vitest fork worker. The RED executes Vitest 4.1.10 and checks the worker process itself; it also proves the candidate still cannot access trusted result FD 198.

Sibling audit: current #1995 already hardens candidate `env`/`execArgv`, dash-leading paths, and V8/UV/worker bounds, but does not propagate the wasm guard. This PR changes no Playwright/Jest path. Its small path-operand helper overlaps future #1995 integration only to prevent a protected path from being parsed as a competing CLI control; integration should deduplicate that identical helper during restack.

gltanaka:
## TDD evidence

RED was committed and pushed before the fix.

- subject/base: `bd4a250420c3b7aaa50bab6fd73aded271115c71`
- RED: `be80ce152`
- observed real-toolchain result: `EvidenceOutcome.FAIL` / `Vitest reported failed protected tests` because the fork worker lacked `--disable-wasm-trap-handler`
- GREEN: `c48227da8`
- isolated PR equivalents: RED `dbd2d0179`, GREEN `d8f22f830`

Green evidence:

- exact standalone Node 22.23.1 + Vitest 4.1.10 worker lifecycle: 2 focused tests passed (worker guard + trusted FD isolation; Linux argv contract)
- full Vitest runner module: passed
- isolated wheel built and installed outside checkout; copied real lifecycle test: 1 passed
- production pylint: 10.00/10; py_compile/diff check: passed

Hosted Linux authenticated sandbox validation is still pending; no merge or deploy is authorized by this comment.

gltanaka:
## Hosted Linux correction — DO NOT MERGE

Disposable staging VM validation at exact PR SHA `d8f22f8306b9361729016d422b390bb24490d548`, Ubuntu 24.04, Node 22.23.1, npm 10.9.8, Vitest 4.1.10 falsified the proposed causal fix.

- new worker lifecycle probe: FAIL (the combined guard/FD assertion did not certify on Linux)
- original authenticated protected test: exact `reporter=missing; kind=signal; SIGABRT; signal_number=6; cgroup OOM=0; OOM-kill=0; pids-max=0; diagnostic_sha256=bb513afd…`
- this reproduction occurred on base `0c3568c65`, which already includes #2121’s 4 GiB virtual-address ceiling

Therefore worker `execArgv` propagation is not the #2083 correction, and the resource-ceiling change does not resolve this exact Linux reproduction. This PR is being closed unmerged. Branches remain as negative evidence; no active base was changed.

gltanaka:
## Critical independent review — PR #2123 @ d8f22f8306b9361729016d422b390bb24490d548

**Verdict: DONT MERGE**

**Findings:** 1 P1, 0 P2

### P1 — candidate Vitest config can negate the trusted worker guard

`_vitest_config_references()` does not reject `test.execArgv` or `test.env`, while this change relies on `--execArgv=--disable-wasm-trap-handler` being authoritative. In exact Vitest 4.1.10, worker argv is constructed in this order: checker CLI `options.execArgv`, resolver conditions, then `project.config.execArgv`; worker env similarly appends `project.config.env` after checker/global values. A candidate static JSON config can therefore append `--no-disable-wasm-trap-handler` after the new trusted literal, or inject `NODE_OPTIONS`/preload controls into the fork environment. Node accepts the positive and later negative boolean flags together, leaving the candidate control last. The new real-toolchain test only exercises the default config, so it proves propagation but not ownership or precedence.

This violates the candidate-isolation requirement and means the patch does not reliably fix the sealed SIGABRT cause for an adversarial or merely preconfigured project. Reject top-level and `test` `env`/`execArgv` by key presence before any collection or execution, and add a red/green exact-4.1.10 regression covering the negating flag and `NODE_OPTIONS`. Current PR #1995 head already contains this fail-closed schema hardening, but #2123 base `0c3568c656dce2c3bda445eb9ce5f597c5a65c70` does not; reconcile the stacked branches rather than duplicating divergent logic.

### P2

None.

### Causality and isolation assessment

The sealed run supports targeting fork workers: authenticated failures were pre-reporter SIGABRT, occurred without OOM/PID pressure, and reproduced serially, so prior PR #1790 style `maxWorkers` pressure mitigation does not explain the failure. The trusted literal is otherwise correctly ordered before `./`-prefixed path operands; the copied launcher/entrypoint identity and FD 198 reporter isolation remain intact. Existing coordinator use already establishes the Linux Node flag compatibility floor. The blocker is candidate precedence, not the worker-propagation hypothesis.

### Verification

- Exact base/head: `0c3568c656dce2c3bda445eb9ce5f597c5a65c70..d8f22f8306b9361729016d422b390bb24490d548`
- `pytest -q tests/test_sync_core_runner_vitest.py`: 149 passed, 2 skipped, 1 warning
- `pylint pdd/sync_core/runner.py`: 10.00/10
- `python -m py_compile pdd/sync_core/runner.py`: passed
- `git diff --check`: passed
- Worktree remained clean; no changes, merge, or deploy performed.

PR #2126, opened 2026-07-17
fix(sync): revert unsupported Vitest address-space exception

gltanaka:
Closes #2125. Related: #2083, #2121, #2123.

## Why

#2121 set a Vitest-specific 4 GiB virtual-address ceiling solely as the claimed #2083 fix. Hosted Linux still reproduces the identical `bb513a…` SIGABRT at its merged base, the trusted private diagnostic contains no address-space/resource failure, and controlled 2/4/8 GiB outcomes do not track the ceiling. The branch is non-releaseable with this evidence-disproven exception silently retained.

This rollback deliberately does **not** claim to fix #2083. It removes only the unsupported virtual-address override and restores the inherited bounded supervisor default. The Vitest 4 GiB physical-memory cap, fail-closed missing-reporter classification, cgroup telemetry, process/output/time limits, no-retry behavior, and all gates remain unchanged.

## TDD evidence

- RED `1f477fd75`: `test_vitest_linux_command_binds_wasm_guard` failed with the exact 4 GiB-vs-default virtual-limit mismatch; pushed separately before GREEN.
- GREEN `d2f3cd387`: removes only the special `max_virtual_memory_bytes` override.
- targeted regression: 1 passed
- full `tests/test_sync_core_runner_vitest.py`: 149 passed, 1 skipped
- production pylint `pdd/sync_core/runner.py`: 10.00/10
- `py_compile` and `git diff --check`: passed

The full test-file pylint score remains its pre-existing 9.77/10 due unrelated baseline diagnostics; no new diagnostic was introduced.

PR #2128, opened 2026-07-17
fix(sync): isolate Vitest reporter descriptor from workers

### LINKED ISSUE #2124: security(sync): Vitest fork worker inherits trusted result descriptor
## Summary

The protected Vitest adapter gives checker-owned result FD 198 to the coordinator reporter, but an exact Node 22.23.1 / Vitest 4.1.10 Linux probe shows the fork worker retains that descriptor. Candidate test code can successfully inspect the descriptor, violating the intended coordinator-only observation boundary and potentially allowing forged reporter bytes.

This was discovered while investigating #2083 but is a separate trust-boundary defect and is not the cause of the `bb513a…` SIGABRT. PR #2123 was closed unmerged after hosted validation separated the two signals.

## Deterministic evidence

On a disposable Ubuntu 24.04 VM, launch the exact Vitest coordinator with a pre-opened inherited FD 198 and checker-owned worker `execArgv`. Two independent worker assertions establish:

- the checker-owned Node guard reaches the fork worker: **pass**
- `fs.fstatSync(198)` must throw in candidate code: **fail** (the descriptor is open)

No raw candidate output or credentials are required to reproduce.

## Acceptance criteria

- A real Linux Node 22.23.1 / Vitest 4.1.10 test proves candidate test/worker code cannot read, stat, duplicate, or write the checker result channel.
- The checker-owned coordinator reporter can still emit the complete bounded JSON result through its private channel.
- A deterministic forge attempt cannot normalize a failing/absent protected test to PASS.
- Preserve toolchain identity, immutable phase, supervisor termination evidence, timeout/process/memory limits, and installed-wheel behavior.
- Cover direct, protected sandbox, repeated-process, and installed-wheel execution.
- Mine PR #1993 and commits `23618b6a8`, `a7a748b86`, `a78d216c8`, `67669ba90`, and `a64cd4582` for the intended coordinator-only transport contract before implementation.

## Non-goals

Do not weaken reporter validation, reinterpret missing reporters, add retries, or fold this into the #2083 SIGABRT correction.

gltanaka:
## Summary

- keep trusted Vitest reporter FD 198 coordinator-only by forcing the checker-owned fork pool and closing the inherited fork copy in a checker-owned Node preload before candidate modules load
- bind the preload to the exact trusted FIFO device/inode and fail before candidate code on descriptor substitution
- reject candidate `poolOptions`, `execArgv`, and `env` authority paths, including empty values
- preserve the bounded reporter parser, immutable toolchain bindings, supervisor limits/termination evidence, and no-retry behavior

## Root cause

The supervisor intentionally opens FD 198 before executing the Vitest coordinator. The descriptor is not close-on-exec because the coordinator reporter must write it, so Vitest fork workers also retained the same open file description. Python-side CLOEXEC would remove the coordinator endpoint too, and Node 22/25 has no `fs.fcntlSync`. A worker-only `execArgv` preload is the first safe execution point: it validates the exact FIFO identity, closes only the forked copy, and runs before candidate setup/test modules.

## TDD

- RED: `b2988debf` (candidate worker can inspect FD 198; config authority paths accepted; trusted preload absent)
- GREEN: `4a0d8fe6d`
- base: `29faf8f53` (includes reviewed #2126 rollback; does not reintroduce the unsupported virtual-address ceiling)

## Test plan

- focused security/config/transport tests: 14 passed
- direct real Vitest 4.1.10 fork lifecycle on local Node 25: coordinator emitted complete JSON; worker `fstat`/write of FD 198 failed
- direct preload exact named-FIFO identity test: pass; mismatched inode rejected before candidate code
- full adapter file first pass: 156 passed, 1 skipped, 1 failed only because the new direct unit used anonymous-pipe metadata instead of the production named FIFO; corrected named-FIFO test passes
- hosted `unit-tests` provides exact Ubuntu / Node 22 / Vitest 4.1.10 / bwrap coverage

Fixes #2124


gltanaka:
## Pre-fix investigation

### Prior contract mined

- PR #1993 and commits `23618b6a8`, `a7a748b86`, `a78d216c8`, `67669ba90`, and `a64cd4582` establish a checker-owned coordinator reporter, private FD 198 transport, complete bounded drain, immutable toolchain bindings, and fail-closed missing/malformed reporter handling.
- Closed PR #2123 showed Vitest worker `execArgv` reaches the real fork worker, but its wasm-guard proposal was unrelated to the descriptor boundary and was correctly closed unmerged.
- Python `FD_CLOEXEC` is not a solution because the coordinator itself is reached through `exec`; Node 22/25 exposes no `fs.fcntlSync`.

### Sibling audit

- Jest uses the same result-FD pattern but forces `--runInBand`, so it has no candidate Jest worker process to inherit the channel.
- Playwright has a distinct anonymous-pipe + aggregate/snapshot protocol and a worker tree. It needs a real topology-specific proof rather than an inferred Vitest fix; follow-up #2129 tracks that bounded audit.
- Other standard adapters either do not use this Node worker topology or have separate transport authority.

### Convention followed

Keep the trusted endpoint at the coordinator, put generated control code in a checker-owned read-only root, validate the exact transport identity before use, reject candidate control-extension paths, and preserve fail-closed reporter interpretation. The new preload runs only in the explicitly forced fork pool, verifies the inherited FIFO device/inode, and closes that fork copy before candidate modules load.

gltanaka:
## TDD and end-to-end evidence

### RED — `b2988debf`

Focused reproduction on the pushed RED commit:

- 7 failed / 4 passed
- six candidate config paths (`test` and root `execArgv`, `env`, `poolOptions`) incorrectly executed as PASS
- checker-owned worker preload assertion failed with `StopIteration` because no worker preload was present

### GREEN — `4a0d8fe6d`

- focused security/config/transport selection: **14 passed**
- exact preload named-FIFO identity unit: **passed**
- malformed identity: Node rejected before candidate `-e` code
- direct real Vitest **4.1.10** fork lifecycle on local Node 25: worker `fstat` and write of FD 198 both failed; coordinator reporter emitted one complete bounded JSON document and exited 0
- `py_compile` for source and test: passed
- `git diff --check`: passed

Full adapter file ran to completion: **156 passed / 1 skipped / 1 failed** in the first pass. The only failure was isolated to the new test harness using anonymous `os.pipe()` metadata on macOS while production uses a named FIFO. After changing that test to the exact named-FIFO transport, the test passes. Hosted `unit-tests` is the authoritative exact Ubuntu + Node 22 + Vitest 4.1.10 + bwrap run.

No retries, reporter relaxation, result reinterpretation, supervisor limit increase, or #2083 claim is included.

gltanaka:
## Final local package evidence

- `uv run --with pylint pylint pdd/sync_core/runner.py`: **10.00/10**
- wheel build: **passed** (`pdd_cli-0.0.305.dev299-py3-none-any.whl`)
- isolated wheel install with all dependencies: **passed**
- installed `pdd --help` console-script smoke: **passed**
- source/installed-wheel `pdd/sync_core/runner.py` SHA-256 parity: `6bd5dac05bf6e8bbe7aa3e2858a4a23af602657ab93f822fb39b46b686ea4d43`
- installed wheel exposes the exact worker-preload helper/source: **passed**
- worktree clean; no unpushed commits; head `4a0d8fe6d2e5c9be236792effae96888c8b27a18`; base `29faf8f5350131ca818cf3e861f230184bfe4d89`

gltanaka:
## Critical security review — DON’T MERGE

Exact subject: 4a0d8fe6d2e5c9be236792effae96888c8b27a18  
Exact base: 29faf8f5350131ca818cf3e861f230184bfe4d89

I reviewed the coordinator-only transport contract from PR #1993 and commits 23618b6a8, a7a748b86, a78d216c8, 67669ba90, and a64cd4582. The forced forks pool, checker-owned CLI precedence, rejection-by-key-presence for root/test projects/workspace/pool/execArgv/env controls, early Node require preload, FIFO device/inode check, bounded coordinator JSON parser, finite supervisor limits, fail-closed termination classification, and no-retry behavior are preserved. However, the required transport proof is incomplete in security-relevant ways.

### P1 — the preload closes only descriptor 198 and does not establish that no duplicate channel descriptor remains

The preload checks and closes RESULT_FD only. If that descriptor is already EBADF it returns without looking for the expected FIFO under any other descriptor. The real sandbox test likewise checks fstat/write/open only at /proc/self/fd/198. It never enumerates /proc/self/fd and compares every open descriptor against the checker FIFO device/inode. Therefore the stated acceptance criterion that candidate code cannot duplicate or discover the inherited result channel is not proved. A duplicate/re-numbered descriptor would remain writable, and the candidate can read the preload to learn the expected device/inode and scan its descriptor table.

Concrete fix: before candidate code, enumerate every numeric /proc/self/fd entry (or the finite RLIMIT_NOFILE range), fstat each stable entry, close every descriptor matching the exact expected FIFO device/inode, and verify no matching descriptor remains. Preserve fail-closed behavior for non-EBADF/non-ENOENT errors. Add an exact Linux Node 22.23.1 / Vitest 4.1.10 sandbox test that parses the expected identity, scans every worker descriptor, attempts open/write through every matching entry, and proves there are none. The test must still prove the coordinator writes its complete bounded result.

### P2 — the forge and packaging/lifecycle acceptance tests are not complete

The real test writes the literal bytes forged to FD 198 and then reports a passing test. The separate parser unit supplies a synthetic return code 1. Neither is the required end-to-end forge adjudication: a worker should attempt a syntactically valid PASS frame while a protected test actually fails (and in a missing-reporter case), with the final trusted outcomes asserted as FAIL and non-PASS respectively. Also add the issue-required repeated-process test using fresh FIFO identities and an installed-wheel protected Vitest run; neither behavior is currently exercised. In the installed-wheel path, assert the preload is mounted read-only and cannot be replaced or modified from candidate code.

### Local validation

- Focused security/config/command selection: 13 passed, 1 skipped (real Linux toolchain skip)
- Full tests/test_sync_core_runner_vitest.py: 157 passed, 1 skipped in 249.94s
- pylint pdd/sync_core/runner.py: 10.00/10
- compileall: passed
- git diff --check: passed
- The full test module pylint command still exits 28 on its existing broad diagnostics; the added diff contains no over-100-character added line.

Hosted exact Linux/Node 22 validation has not run for this feature-branch-targeting PR yet; only auto-heal is present. That is not the reason for this verdict—the P1 transport-proof gap must be corrected first.

gltanaka:
## Codex adversarial review — PR #2128 @ `4a0d8fe6d2e5c9be236792effae96888c8b27a18`

**Verdict:** DON'T MERGE
**Findings:** 2 total (1 P1, 1 P2)

### P1 (blocking)
- [P1] The trusted FIFO remains reopenable by candidate code after FD 198 is closed — `pdd/sync_core/runner.py:4556`, `pdd/sync_core/supervisor.py:34`, `pdd/sync_core/supervisor.py:2509`, `pdd/sync_core/supervisor.py:3342`
  The preload validates and closes only the inherited numeric descriptor. The supervisor still bind-mounts the same mode-0600 FIFO with writable `--bind` at the fixed `/run/pdd-framework-observation` pathname, then opens it without unlinking or hiding that mount before it drops into the same candidate uid. Candidate test or product code can therefore call `fs.openSync('/run/pdd-framework-observation', 'w')` to obtain a fresh writer, then stat, duplicate, or write the checker-owned channel despite FD 198 being EBADF. The new real-runtime assertion checks only FD 198 and `/proc/self/fd/198`, so it cannot detect this bypass. Even where concatenated bytes fail closed instead of forging PASS, this violates the coordinator-only integrity boundary and gives candidate code a deterministic trusted-evidence corruption/denial primitive.
  **Suggested fix:** hand the coordinator a non-reopenable anonymous descriptor (or have trusted setup hide/unmount the named endpoint before any candidate code can run), and add exact Linux tests that reject opening/writing the canonical observation pathname as well as FD 198 while still proving one complete coordinator report.

### P2 (non-blocking)
- [P2] Candidate browser configuration is not sealed by the claimed fork-only policy — `pdd/sync_core/runner.py:1073`
  The new authority checks reject `pool`, `poolOptions`, `execArgv`, and `env`, but still accept `test.browser`. In Vitest 4.1.10, browser-enabled projects are routed to the browser pool before the configured `pool` is consulted, so `--pool=forks` is not unconditional fork-only enforcement. A static JSON config cannot provide the normal provider factory and will generally fail closed, so I did not find a direct reporter-forgery path through this control; it is nevertheless an unbound execution-mode escape and leaves the policy/test claim inaccurate.
  **Suggested fix:** reject presence of `test.browser` like the other execution controls and add empty/disabled/enabled browser-config regressions.

### Notes
- Vitest 4.1.10 source confirms `--execArgv=--require=<absolute-path>` is parsed as an array worker argument, CLI options override static config, and the fork worker receives that `execArgv`; Node `--require` runs before the worker entrypoint. The preload's BigInt device/inode comparison and EBADF handling are otherwise fail-closed for the inherited descriptor.
- The coordinator parser still requires complete supported identities/statuses and treats nonzero execution as FAIL; no retry, result reinterpretation, supervisor-limit increase, gate weakening, or #2121 virtual-address ceiling was reintroduced.
- Local exact-head evidence: focused security/config tests passed; full `tests/test_sync_core_runner_vitest.py` = **157 passed, 1 skipped**; `pylint pdd/sync_core/runner.py` = **10.00/10**; `git diff --check` clean; worktree clean.
- The skipped path is the authoritative Linux sandbox test. Exact-head Actions currently show only successful `auto-heal` and `heal`; there is no Node 22/Linux unit-test run. Local Node 25/macOS proof does not satisfy the issue's required Node 22.23.1/Vitest 4.1.10/Linux hosted acceptance even after the P1 is corrected.

gltanaka:
## Implementer response to security reviews

Review references: https://github.com/promptdriven/pdd/pull/2128#issuecomment-4998165817 and https://github.com/promptdriven/pdd/pull/2128#issuecomment-4998174293

### P1: duplicate or renumbered result-channel descriptors
Status: **FIXING in this PR**
Plan: before any candidate module runs, remove every worker descriptor matching the exact trusted channel identity and fail closed unless a complete post-close scan proves none remain.

### P1: named observation endpoint remains reopenable
Status: **FIXING in this PR**
Plan: replace or hide the path-backed worker-visible endpoint so candidate code cannot reopen `/run/pdd-framework-observation`, while preserving the coordinator-only writer.

### P2: end-to-end forge, repeated-process, and installed-wheel coverage
Status: **FIXING in this PR**
Plan: add valid-PASS forge attempts against real FAIL/missing evidence, fresh-identity repeated runs, and protected installed-wheel/preload-immutability coverage.

### P2: `test.browser` execution-mode escape
Status: **FIXING in this PR**
Plan: reject browser key presence at every supported static-config authority level and add empty/disabled/enabled regressions.

The corrected head will receive one exact-head critical re-review and authoritative Linux Node 22.23.1 / Vitest 4.1.10 hosted validation before merge.

gltanaka:
## Implementer response to both review rounds

Re: https://github.com/promptdriven/pdd/pull/2128#issuecomment-4998165817 and https://github.com/promptdriven/pdd/pull/2128#issuecomment-4998174293

### P1: fixed FIFO pathname remains candidate-reopenable
Status: **FIXING**

Plan: generalize the existing root-helper anonymous observation handoff so standard-framework reporters receive a parent-created anonymous pipe without mounting `/run/pdd-framework-observation` or any candidate-visible FIFO pathname. Preserve the existing bounded parent drain, supervisor lifetime/termination attribution, cgroup limits, and coordinator complete-JSON behavior.

### P1: preload closes only FD 198 and misses duplicate/re-numbered descriptors
Status: **FIXING**

Plan: make the worker preload enumerate the finite descriptor table, bind the expected observation identity, close every matching descriptor before candidate modules load, verify none remain, and fail closed on errors other than the explicitly safe disappeared/invalid-descriptor cases. Exact protected tests will scan the worker descriptor table and prove no matching channel identity remains.

### P2: candidate browser execution mode remains unsealed
Status: **FIXING**

Plan: reject root and `test` `browser` by key presence, including empty, false, and enabled values.

### P2: forge, lifecycle, and package acceptance coverage is incomplete
Status: **FIXING**

Plan: add protected tests for a syntactically valid forged PASS paired with an actual failing test, forged PASS with missing reporter, repeated runs with fresh observation identities, and an installed-wheel protected run that proves the preload is mounted read-only/nonreplaceable. The correction will not add retries, reinterpret results, change limits, expand into Jest/pytest/Playwright, or claim to fix #2083.

gltanaka:
Correction evidence at exact head 7157d02487eda71123eec45b29d18dcd4a518903

Root cause fixed: standard Vitest used a candidate-visible fixed FIFO pathname and the first worker preload closed only fd 198, so duplicated or renumbered references to the same endpoint survived. The corrected path generalizes the supervisor root-helper anonymous observation pipe beyond Playwright, passes only trusted device/inode identity to the preload, enumerates the finite descriptor table, closes every exact matching FIFO duplicate, verifies none remain, and fails closed on unexpected descriptor errors. Root and test browser controls are rejected by key presence.

TDD/history:
- b3269dec4 RED: 9 failed / 10 passed, reproducing duplicate-descriptor, browser-control, fixed-path, and standard-anonymous transport gaps.
- 61cf77b51 additional correction tests: forged valid PASS plus actual failure, missing reporter non-PASS, canonical path absence, zero matching worker descriptors, immutable preload, repeated endpoint/nonce authority, and adapter regressions.
- 7157d0248 GREEN implementation.

Validation at head:
- focused correction selection: 21 passed, 147 deselected
- full tests/test_sync_core_runner_vitest.py: 163 passed, 3 skipped in 227.88s
- full tests/test_sync_core_supervisor.py: 292 passed, 48 skipped in 16.93s
- py_compile: runner, supervisor, and both changed test files passed
- pylint runner.py supervisor.py: 10.00/10
- wheel build passed; packaged runner.py and supervisor.py byte-for-byte match source (SHA-256 a5dc23516f146a5b85567f0b174329feda4ba4af2d4ad5146ce23df88d771d4f and 769dda0a3d5e60e44c4df77654b33cb1ded54b7706a0179b638b8c767569e9d3)
- clean worktree and exact remote-head match
- scope versus base 29faf8f53 is only runner.py, supervisor.py, Vitest runner tests, and supervisor tests; no workflow, Jest, or pytest-adapter diff

The 3 real Linux/bwrap tests are platform-skipped on this macOS worker and remain for hosted Node 22/Linux orchestration. No cloud, merge, or deploy action was taken.

gltanaka:
## Codex exact-head correction re-review — PR #2128 @ `7157d02487eda71123eec45b29d18dcd4a518903`

**Code verdict:** MERGE

The correction closes both prior P1s and the requested P2 coverage:

- Vitest no longer supplies a named FIFO or mounts `/run/pdd-framework-observation`. The root helper creates the anonymous observation pipe, drains it under the existing bound, publishes a nonce/digest/size-bound root-owned artifact, and only then relays validated bytes to the parent endpoint.
- The worker preload binds the helper-provided pipe device/inode, enumerates `/proc/self/fd` with a finite 0–255 fallback matching the enforced `RLIMIT_NOFILE`, closes every matching FIFO descriptor, rescans, and fails closed if any copy remains or an unexpected descriptor error occurs. The checker-owned coordinator retains its writer.
- Root and `test.browser` controls are rejected by key presence for false, empty, and enabled values. Vitest 4 worker `execArgv` remains checker-owned and fork-only.
- Real-Linux tests now attempt a valid forged PASS alongside an actual failure and a killed/missing reporter; expected outcomes are FAIL and non-PASS. Repeated endpoints/nonces are fresh, and the installed-wheel path asserts site-packages execution plus read-only/nonreplaceable preload behavior.
- The generic anonymous helper path remains separate from the Playwright aggregate validation: an absent aggregate is accepted only for the standard anonymous mode, while a present aggregate retains its full snapshot/checker identity validation and descriptor protocol. Existing limits, cgroup/termination attribution, bounded output, cleanup, and no-retry semantics are unchanged.
- Scope is exactly `runner.py`, `supervisor.py`, and their Vitest/supervisor tests. No Jest/pytest adapter, workflow, retry, gate, address-space, or #2083 behavior changed. The original RED/GREEN commits remain in history; the correction adds new RED/coverage/GREEN commits without rewriting them.

Local exact-head verification:

- Focused Vitest security/config/forge selection: **21 passed, 3 Linux-only skipped**.
- Focused generic-anonymous/Playwright-supervisor selection: **3 passed, 1 Linux-only skipped**.
- `py_compile` for both source and test files: passed.
- `pylint pdd/sync_core/runner.py pdd/sync_core/supervisor.py`: **10.00/10**.
- `git diff --check`: clean; worktree and remote head match exactly.

Hosted Node 22.23.1 / Vitest 4.1.10 / Linux+bwrap execution remains a separate required merge gate. Exact-head Actions currently show successful `auto-heal` and `heal`, but not that authoritative hosted unit-test proof; this MERGE verdict is for the reviewed code and local evidence only.

gltanaka:
## Critical correction-cycle re-review — DON’T MERGE

Exact subject: 7157d02487eda71123eec45b29d18dcd4a518903  
Exact base: 29faf8f5350131ca818cf3e861f230184bfe4d89

The correction resolves the two direct code escapes from the prior reviews: the standard Vitest path now uses a helper-created anonymous pipe rather than the canonical FIFO mount; the worker preload closes every self-descriptor matching the exact pipe device/inode and post-verifies none remain; browser authority is rejected by key presence at both root and test levels; and valid-PASS forge fixtures now adjudicate actual FAIL and missing-reporter outcomes. The coordinator parser remains bounded and complete, the preload is mounted read-only, and the #2083 no-retry, finite-limit, timeout, telemetry, and fail-closed termination paths are unchanged.

Two required acceptance boundaries are still not completed on this remote head.

### P1 — installed-wheel protected Vitest behavior is never executed

The helper contains a PDD_REQUIRE_INSTALLED_WHEEL assertion, but .github/workflows/unit-tests.yml is unchanged from the exact base and no job sets that variable for a protected Vitest run. Package Preprocess Smoke still provisions only the Playwright protected runtime; its Vitest work is grammar-only. Thus the conditional is dormant and the issue-required installed-wheel observation boundary, preload immutability, coordinator report, and forge denial are not tested from the built artifact. Byte/grammar parity is not a substitute.

Concrete fix: provision the exact Vitest toolchain in the package job, copy the protected Vitest test outside the checkout, invoke it with the isolated wheel venv and PDD_REQUIRE_INSTALLED_WHEEL=1, and assert runner_module itself resolves under site-packages. The completed backup correction ffdbcb7dba8984ce43a672f81bcf6cbd4d8a3629 contains the missing workflow shape. Also pin setup-node to 22.23.1 rather than floating 22 so the stated Node 22.23.1 acceptance remains reproducible.

### P1 — anonymous does not yet prove non-reopenable through other process descriptors

The coordinator necessarily retains the anonymous pipe writer while the worker runs, the sandbox mounts /proc, and candidate code already demonstrates same-UID authority to signal its parent. The real test scans only /proc/self/fd and the removed canonical path. It never attempts /proc/<coordinator-or-other-visible-pid>/fd/<descriptor>. On systems where ptrace/proc permissions allow same-UID descriptor access, a worker could reopen the coordinator writer despite having no self duplicate. The implementation neither seals this with an explicit process policy nor behaviorally verifies the exact hosted boundary.

Concrete fix: in the exact Linux sandbox fixture, enumerate every visible non-self PID and accessible fd entry, compare identities to the expected device/inode, and attempt open/write through every match; assert none are accessible. Prefer enforcing a deterministic non-dumpable/different-credential or equivalent proc boundary rather than relying silently on the host Yama default, and add a trusted setup probe if that policy is the intended authority.

### P2 — the repeated-run test does not exercise repeated protected Vitest processes

The new repeated test creates three outer caller pipes, records their already-distinct identities, then monkeypatches _sandbox_command to fail during construction. It never creates the helper-owned inner observation pipe, launches a worker preload, or completes a Vitest report. It proves fresh parent nonces and Python os.pipe behavior, not the required repeated-process transport behavior.

Concrete fix: run the protected Vitest execution repeatedly with distinct repositories/temp roots and assert each run completes through a fresh authority while the worker observes no matching self or cross-process descriptor. At minimum add the runner-level repeated execution in ffdbcb7dba8984ce43a672f81bcf6cbd4d8a3629; the authoritative hosted test should exercise the real sandbox rather than only a capture mock.

### Deterministic local evidence

- Runner security/config/forge/termination focus: 26 passed, 3 skipped; the skips are the three real Linux/Node toolchain tests
- Supervisor anonymous/repeated/limit/termination focus: 5 passed
- pylint runner.py and supervisor.py: 10.00/10
- compileall: passed
- git diff --check: passed
- detached worktree remained clean

Only auto-heal/heal exists for this exact head. Exact Linux Node 22.23.1/Vitest 4.1.10 source and installed-wheel validation therefore remain hosted-only after these blocking coverage corrections.

gltanaka:
## Escalation response to correction-cycle re-review

Re: https://github.com/promptdriven/pdd/pull/2128#issuecomment-4998364228

### P1: installed-wheel protected Vitest lane and exact Node pin
Status: **FIXING**

Plan: add the isolated built-wheel protected Vitest run, assert the runner module is imported from `site-packages`, provision the exact Vitest 4.1.10 toolchain, and pin Node 22.23.1 for reproducible source and wheel lanes.

### P1: cross-process `/proc/<pid>/fd` reopenability
Status: **FIXING**

Plan: enforce an explicit coordinator process policy that denies same-UID proc-fd reopening, then behaviorally enumerate visible non-self processes/descriptors in the real worker and prove no matching observation identity can be opened or written. Host Yama defaults will not be treated as the authority.

### P2: repeated protected process coverage
Status: **FIXING**

Plan: replace the construction-only mock with repeated real protected Vitest executions using fresh repositories/temp roots and assert complete coordinator results plus no self/cross-process observation access each time.

This is the permitted substantive-review escalation. No retry, result reinterpretation, resource-limit change, or #2083 causal claim will be added.

gltanaka:
## Final critical escalation review — DON’T MERGE

Exact subject: 23186aedba16fc0445401313c2e553488982c2f1  
Exact base: 29faf8f5350131ca818cf3e861f230184bfe4d89

The escalation resolves most of comment 4998364228. The workflows pin Node 22.23.1 and Vitest 4.1.10, provision Vitest in the package job, execute the protected test outside the checkout under the isolated wheel interpreter, and assert both pdd and runner_module come from site-packages. The real repeated test runs three protected workers with fresh authority identities. Anonymous/self-descriptor closure, canonical-path denial, browser rejection, valid-PASS FAIL/missing adjudication, immutable preload checks, bounded coordinator JSON, finite limits, trusted termination telemetry, and no-retry behavior remain intact.

One P1 escape remains.

### P1 — the proc seal hides only /proc/<pid>/fd, leaving equivalent task/TID and pidfd duplication paths

The root inner supervisor bind-mounts the empty seal only over /proc/<coordinator-pid>/fd. Linux exposes the same process descriptor table through aliases including /proc/<pid>/task/<tid>/fd; for the thread-group leader this reaches the coordinator descriptors without traversing the sealed mountpoint. Thread IDs can also be addressed through proc paths even when they are not returned by the top-level directory listing. The real worker probe enumerates only /proc/<pid>/fd, so it reports success without testing these aliases.

The bind mount also does not change ptrace authorization. Where ambient same-UID ptrace policy permits it, pidfd_open plus pidfd_getfd can duplicate the coordinator writer without using the sealed pathname. Relying on the hosted machine Yama default is not an explicit, verified sandbox policy.

Concrete fix: establish and verify a coordinator non-dumpable policy after the uid drop and before Node starts, failing closed on setup or verification error, so proc aliases and pidfd_getfd are denied independently of path spelling. The independent green escalation 3def0d116fa2a6ba9bc298b0321ca6de329b39eb contains the PR_SET_DUMPABLE/PR_GET_DUMPABLE direction. If retaining the mount defense, also hide /proc/<pid>/task (including future thread entries), not only /proc/<pid>/fd. Extend the real worker probe to enumerate /proc/<pid>/task/<tid>/fd for every visible process/thread and attempt stat/open/write through each matching identity; add a pidfd_getfd denial probe or an equivalent trusted assertion of the kernel permission boundary.

### Deterministic local evidence

- Runner workflow/config/preload/forge/termination focus: 26 passed, 4 skipped; skips are the real Linux toolchain tests
- Supervisor anonymous seal/status/limit/termination focus: 12 passed
- pylint runner.py and supervisor.py: 10.00/10
- compileall: passed
- git diff --check: passed
- detached worktree remained clean

Only auto-heal/heal exists for this feature-branch-targeting exact head. Exact source and installed-wheel Linux runs remain hosted-only after the P1 proc-boundary correction.

gltanaka:
## Final P1 response

Re: https://github.com/promptdriven/pdd/pull/2128#issuecomment-4998578682

### P1: proc task aliases and `pidfd_getfd` remain authorized
Status: **FIXING**

Plan: retain the exact `/proc/<pid>/fd` mount seal as defense-in-depth, add a verified `PR_SET_DUMPABLE=0` coordinator policy before the Node exec to revoke ptrace-style proc/pidfd duplication authority across every alias, and expand the real worker probe to `/proc/<pid>/task/<tid>/fd`. Setup or verification failure will stop before candidate code. The hosted exact-head test must still prove complete coordinator output and all self/cross-process access denied.

gltanaka:
### Final exact-head security review

**Verdict: DONT MERGE**

Reviewed exact head `eedbd33f9cee3ecee8fc63531c4e3de5cb87001b` against base `29faf8f5350131ca818cf3e861f230184bfe4d89`.

#### Resolved from the prior P1

- The coordinator descriptor is sealed at `/proc/<pid>/fd`.
- The trusted pre-exec probe covers `/proc/<pid>/fd`, `/proc/<pid>/task/<tid>/fd`, top-level `/proc/<tid>/fd` aliases, and `pidfd_getfd`.
- `setpriv --ptracer none` is checked fail-closed and its restriction is preserved across exec.
- The protected worker performs the real post-exec proc scan, and the previous anonymous/self-FD, preload, browser, forge, finite-limit, termination, no-retry, installed-wheel, exact Node/Vitest, and repeated-run controls remain present.
- The exact mount seal is retained.

#### Blocking finding

**P2 - Hardened Yama modes are rejected even though they satisfy a stronger security policy.**

At `pdd/sync_core/supervisor.py:2564`, the wrapper requires `ptrace_scope` to equal exactly `1`. Linux Yama modes `2` and `3` are stricter than mode `1`: they restrict ptrace to administrators or disable attach entirely. The setpriv documentation also says `--ptracer` is meaningful only in mode `1`; outside mode `1`, the stronger kernel policy itself provides the denial. The new behavioral probe already verifies all required proc and pidfd denial paths, so rejecting modes `2` and `3` creates an avoidable release regression on hardened Linux hosts while adding no security benefit.

This is blocking compatibility behavior, not an acceptable fail-closed condition: protected Vitest becomes unavailable on a host where the required authority denial is demonstrably stronger.

Required correction: parse the sysctl as an integer, reject mode `0` and malformed/unreadable values, accept documented modes `1` through `3` (or values at least `1`), and retain the full behavioral probe as the final fail-closed gate.

References: [Linux Yama documentation](https://docs.kernel.org/admin-guide/LSM/Yama.html), [setpriv manual](https://man7.org/linux/man-pages/man1/setpriv.1.html).

#### Validation

- Focused runner security tests: **26 passed, 4 skipped**
- Focused supervisor security tests: **13 passed**
- Pylint: **10.00/10**
- Compileall: **passed**
- `git diff --check`: **passed**
- Worktree remained clean

No additional P1 finding remains at this exact head.

gltanaka:
### Final exact-head security re-review

**Verdict: MERGE**

Reviewed exact remote head `1eebbd1dd08ef7eaf406c445f61b0b50d033aba2` against base `29faf8f5350131ca818cf3e861f230184bfe4d89`. No blocking findings remain.

- The correction uses explicit base-10 integer parsing, accepts only Yama modes `1`, `2`, and `3`, and fails closed for unreadable, malformed, `0`, negative, and out-of-range values.
- The full behavioral probe remains intact for `/proc/<pid>/fd`, task/TID aliases, top-level TID aliases, and `pidfd_getfd`; the exact coordinator proc-fd mount seal and exec-stable `setpriv --ptracer none` gate are retained.
- The correction changes only supervisor policy parsing and its focused tests. Exact Node `22.23.1`, Vitest `4.1.10`, installed-wheel enforcement, protected copied runner, and repeated fresh-authority tests are unchanged.
- Hosted checks observed without rerun: `heal` and `auto-heal` both passed at this exact head.

Validation:
- Focused supervisor boundary: **21 passed**
- Focused runner/Vitest boundary: **26 passed, 4 skipped**
- Pylint: **10.00/10**
- Compileall and full PR `git diff --check`: **passed**
- Review worktree: **clean**

Residual uncertainty: the four real Linux Vitest cases are skipped on this macOS review host, so this review did not independently execute the kernel Yama behavior. The deterministic policy/boundary tests are green and the Linux workflow protections are retained unchanged.

gltanaka:
### Exact-head framed-descriptor security review

**Verdict: MERGE**

Reviewed exact remote head `de7d0ee9f111767bbee8be637f1e46f6fbbbf345` against base `29faf8f5350131ca818cf3e861f230184bfe4d89`. No blocking findings remain.

- Standard-anonymous and Playwright results share bounded framing but remain domain-separated: standard uses the fixed `pdd-standard-anonymous-observation-v1` digest, while Playwright requires its attested aggregate digest. Both require the fresh nonce, exact canonical result shape, authenticated nested candidate record, bounded stdout/stderr/observation, and matching observation size/hash.
- Candidate stdio and observation descriptors remain distinct from the parent control descriptors. The parent validates the complete result before forwarding observation bytes, captures cgroup telemetry before ACK, and requires exact ACK digest, helper exit status, terminal EOF, deadlines, and cleanup.
- Conflicting FIFO plus descriptor endpoints and Playwright without a descriptor endpoint fail during construction.
- The standard-only `setpriv --ptracer none` gate, Yama behavioral task/TID/top-level-TID/`pidfd_getfd` probes, and exact coordinator proc-fd mount seal remain enabled.
- The revised legacy authority replay test is legitimate negative cross-protocol coverage: the retired file-control helper cannot satisfy framed READY, returns a typed setup failure, and relays zero observation bytes. The supported framed path has separate positive result-binding and malformed/replay/timeout/ACK/cleanup coverage.
- Hosted checks observed without rerun: `heal` and `auto-heal` passed at this exact head.

Validation:
- Full deterministic supervisor suite: **306 passed, 30 skipped, 18 deselected**
- Focused runner/Vitest boundary: **26 passed, 4 skipped**
- Pylint: **10.00/10**
- Compileall and full PR `git diff --check`: **passed**
- Review worktree: **clean**

Residual uncertainty: real Linux-only cases were skipped on this macOS review host. An initial concurrent-load run hit two unrelated 100 ms legacy mock timing failures; both passed isolated and in the clean sequential full-suite rerun.

gltanaka:
## Exact-head local validation update

Head: de7d0ee9f111767bbee8be637f1e46f6fbbbf345

- Full supervisor suite: **306 passed, 48 skipped** in 27.06s
- Full Vitest adapter suite: **164 passed, 4 skipped** in 216.50s
- Unit-test workflow contract: **22 passed**
- Pylint on runner.py and supervisor.py: **10.00/10**
- Compileall for both source and both changed test modules: **passed**
- Wheel build: **passed** (pdd_cli-0.0.305.dev312-py3-none-any.whl)
- Isolated wheel import from outside the checkout: **passed**
- Source/installed-wheel SHA-256 parity:
  - runner: a5dc23516f146a5b85567f0b174329feda4ba4af2d4ad5146ce23df88d771d4f
  - supervisor: 9345627b3d75881add03216e73be8804330e5c37b8ce5f352239950c739b3940
- Installed-wheel protected Vitest target imported the isolated site-packages package and then **skipped only the Linux/bwrap body** on this macOS host.
- git diff --check: passed; worktree clean; local and remote heads match.

The exact-head framed-descriptor security re-review has no blocking code finding. Merge remains gated on the authoritative hosted Linux run with Node 22.23.1, Vitest 4.1.10, bwrap, and the copied installed-wheel test; current PR checks expose only heal and auto-heal, not that Linux integration lane.

gltanaka:
### Hosted regression follow-up

**Verdict: DONT MERGE**

This supersedes my earlier approval of exact PR head `de7d0ee9f111767bbee8be637f1e46f6fbbbf345`.

#### Blocking hosted failure

PR #2134 [Package Preprocess Smoke job 87801157139](https://github.com/promptdriven/pdd/actions/runs/29553620089/job/87801157139) failed all seven installed-wheel Playwright 1.55 config cases. Every case completed descriptor READY/START, then remained in `_run_playwright_descriptor_supervised` waiting for the result frame until the 90-second pytest timeout. The identical lane passed all seven cases at baseline `7157d0248` in [job 87791208796](https://github.com/promptdriven/pdd/actions/runs/29550297153/job/87791208796).

#### Causal change

The first and only change in `7157d0248..de7d0ee9f` that semantically alters the Playwright candidate launch is commit `23186aedb` (`fix(sync): seal Vitest coordinator proc descriptors`), specifically the unconditional seal synchronization added at current `pdd/sync_core/supervisor.py:1902-1909` and the unconditional seal/mount argument path at `pdd/sync_core/supervisor.py:3483-3488`.

Although Playwright resolves the seal flag to `0`, it still executes the new `seal_read, seal_write = os.pipe()` fork/release gate and altered inner-supervisor argv. The later `setpriv --ptracer none`, Yama probe, proc alias probe, and mount operation itself are correctly guarded by `standard_anonymous`; the shared-framing commit leaves the existing Playwright nonce/aggregate/result/ACK grammar equivalent. `runner.py` is byte-identical between the passing baseline and this head, and both hosted jobs used Node `v22.23.1`. This isolates the regression to the unconditional inner launch change introduced by `23186aedb`, not the toolchain or runner.

#### Required security-safe correction

- Preserve the proc-fd mount seal and Yama/ptracer defenses for standard anonymous Vitest.
- Restore Playwright to the pre-`23186aedb` inner coordinator launch path: do not allocate/read the seal release pipe, inject the seal/mount argv, or add the mount runtime solely for a Playwright launch.
- Prefer separate standard-anonymous and Playwright inner launch constructions, or make the entire seal handshake conditional before pipe creation/fork behavior. A runtime flag of `0` that still changes Playwright launch behavior is insufficient.
- Retain Playwright descriptor nonce, aggregate attestation, nested candidate status, observation hash/size, ACK, EOF, deadlines, cgroup telemetry, and cleanup unchanged.
- Add an assertion that a Playwright plan contains no standard-anonymous seal/ptracer/Yama machinery, and require the real installed-wheel Playwright Chromium lane to pass on the exact correction head.

#### Test coverage assessment

Local deterministic checks remained green because real Chromium cases are skipped on macOS, and the synthetic Linux descriptor-chain test uses Python fork fixtures rather than the actual Playwright/Chromium lifecycle. The mandatory installed-wheel hosted test is the correct regression test and it failed. This must be corrected and pass at the exact PR head before merge.

Remaining uncertainty: the hosted timeout kills the process before helper-side state is reported, so the precise kernel wait inside the new seal launch sequence is not observable from this log. The causal commit and altered line range are isolated by the passing baseline, identical runner/Node inputs, and elimination of the standard-only later changes.

gltanaka:
Hosted regression response for run 29553620089 / job 87801157139: status FIXING. The exact failure is accepted as blocking and is not the #2083 Vitest abort class. Independent review and implementation both isolate commit 23186aedb: Playwright aggregate launches were routed through the Vitest-only proc-seal inner supervisor topology even with the seal flag disabled. A TDD correction at local head 57df3f1076b0ab50aa77f0f40d5ab181df453d29 restores the pre-23186 unsealed inner launch only for Playwright while retaining standard-anonymous mount/ptrace/Yama/task/TID/pidfd/framing defenses. Local supervisor, Vitest, and Playwright suites are green; exact correction review and a fresh immutable hosted installed-wheel Chromium validation are required before merge. No timeout relaxation, retry, result reinterpretation, or production action will be used.

gltanaka:
Correction review verdict: MERGE for exact local head `57df3f1076b0ab50aa77f0f40d5ab181df453d29` (RED `5732e3e68f587be076b06633a9491d307d2dc9d9`, correction base `de7d0ee9f111767bbee8be637f1e46f6fbbbf345`).

Findings: none in the correction boundary. The Playwright aggregate path now uses a dedicated unsealed inner status supervisor that is byte-for-byte identical to the known-good pre-regression implementation at `7157d0248` (1,259 bytes; SHA-256 `a35afb7fe69478c9a0b065f84a91ce759fe97f86f1a69a8a55f4bdece9e92faa`). Protocol selection is parent-controlled by the validated aggregate. Nonce, aggregate/hash validation, authenticated nested termination status, cgroup placement/telemetry, ACK/EOF ordering, helper-exit agreement, and cleanup remain in the shared path. Standard anonymous execution retains the sealed supervisor, mount staging, `--ptracer none`, Yama/probes, and framing unchanged.

TDD/evidence:
- RED commit adds assertions that Playwright has neither the seal pipe nor seal marker; GREEN commit implements only that split.
- `tests/test_sync_core_supervisor.py -m "not real"`: 306 passed, 30 skipped, 18 deselected.
- Focused runner/Vitest boundary: 26 passed, 4 skipped, 138 deselected.
- Pylint: 10.00/10; compileall and `git diff --check de7d0ee9f..57df3f107` passed; review worktree clean.

Residual gate: this local verdict does not close the hosted blocker by itself. The exact corrected head must be pushed and the hosted Linux installed-wheel Playwright lane must pass all seven cases before merge/release.

gltanaka:
### Hosted standard-anonymous blocker

**Verdict: DO NOT MERGE `57df3f1076b0ab50aa77f0f40d5ab181df453d29`.**

PR #2135 validation SHA `0a7ffb3614102c4e79bce915146ea2236ded22ca` confirms the Playwright correction: all 7 installed-wheel cases passed. The immediately following exact installed-wheel Vitest case failed before the standard-anonymous READY frame with sandbox exit 125, trusted phase `scope-setup`, diagnostic `d6ef3492984d6c1486d7589c748c10bd28fd064ee17f413a8c1b8af6d0df4cea`: https://github.com/promptdriven/pdd/actions/runs/29554991891/job/87805130775

Root cause is the standard-only pre-READY feature gate introduced by `eedbd33f9` and retained through framed commit `3923a1024`: the root helper requires `setpriv --help` to advertise `--ptracer`, and the candidate argv uses `setpriv --ptracer none`. The hosted image is Ubuntu 24.04 (Noble), whose util-linux 2.39.3 `setpriv` does not provide `--ptracer`: https://manpages.ubuntu.com/manpages/noble/man1/setpriv.1.html. Playwright now intentionally skips this standard-only branch, explaining the exact pass/fail split. Commit `23186aedb` introduced the proc-fd seal and the hosted Vitest lane; `3923a1024` moved standard anonymous onto READY framing; `57df3f107` fixed only the Playwright topology and leaves this unsupported host-tool dependency intact. This is not the #2083 abort class.

Required safe correction: do not remove the ptrace invariant. Remove the host-version-dependent `setpriv --ptracer` argument and preflight, and apply `PR_SET_PTRACER` with pid 0 directly in the checker-owned Python anonymous wrapper after `setpriv` has dropped UID/GID and before the denial probe and exec. Fail closed on the `prctl` result; retain the coordinator `/proc/<pid>/fd` bind seal, Yama mode check, task/TID and `pidfd_getfd` probes, framed nonce/hash/ACK/EOF protocol, cgroup placement, and cleanup.

Missing regression coverage: exercise standard anonymous construction/execution with a Noble-compatible `setpriv` that has no `--ptracer`; assert no unsupported option or pre-READY feature probe is emitted, assert the wrapper performs and checks `PR_SET_PTRACER(0)` before its cross-process probes/exec, and require this exact installed-wheel Vitest lane to pass on the corrected head.

gltanaka:
### Exact-final hosted-blocker correction review

**Verdict: MERGE `83a1cb74f83a8bc823b6ac3687ac54c74e83d5a4`, subject to the mandatory exact-head hosted Ubuntu Noble installed-wheel lane.**

Findings: none.

Both observed hosted blockers are corrected:
- Authenticated argv: the standard-anonymous launch now emits the exact root-helper-authenticated Noble suffix `--reuid UID --regid GID --clear-groups --`; the stale `--ptracer none` mismatch is gone.
- Noble compatibility: the pre-READY `setpriv --help` feature gate and unsupported `setpriv --ptracer` argument are gone. The checker-owned anonymous Python wrapper calls and checks `prctl(PR_SET_PTRACER, 0)` after `setpriv` performs the UID/GID drop, before the Yama check, `/proc` task/TID and `pidfd_getfd` denial probes, and Node exec. A nonzero result fails closed.

The standard coordinator proc-fd bind seal and release ordering remain unchanged. Nonce/aggregate/hash/status authentication, ACK/EOF, cgroup placement and telemetry, cleanup, and the dedicated unsealed Playwright aggregate path are unchanged from `57df3f107`.

Review evidence:
- New ptrace/authentication/Playwright isolation focus: 13 passed.
- Full deterministic supervisor: 308 passed, 30 skipped, 18 deselected.
- Focused Vitest boundary: 26 passed, 4 skipped, 138 deselected.
- Full deterministic Playwright file: 264 passed, 16 skipped.
- Pylint 10.00/10; compileall and diff check passed; exact review worktree clean.

Residual uncertainty: macOS cannot execute the real Linux/Yama/cgroup path. The exact pushed SHA must pass the hosted Ubuntu Noble installed-wheel Vitest case and the seven installed-wheel Playwright cases before merge/release.

gltanaka:
### Exact remote merge-head critical review

**Verdict: MERGE `5ad0f7e128dbcbf422098924288fbce7a6b8a5e8`, subject to mandatory exact-head hosted Ubuntu Noble validation.**

Findings: none.

Reviewed merge `0e3bcb78cdd958ff3e28bf9ed33ca4fb1733559e` with parents `9d835ed0e3e8026483ac9c3337a6b9f445403aab` and previously approved Noble correction `83a1cb74f83a8bc823b6ac3687ac54c74e83d5a4`, plus test-only barrier commit `5ad0f7e12`.

Merge integrity:
- The actual and authenticated expected standard drop suffix remain exactly `--reuid UID --regid GID --clear-groups --`; no Noble-incompatible `setpriv --ptracer` option or feature probe returned.
- Checked `prctl(PR_SET_PTRACER, 0)` remains in the checker-owned post-drop wrapper before Yama, proc task/TID, `pidfd_getfd`, and Node exec; nonzero return still fails closed.
- Pre-READY setup telemetry is a strict canonical descriptor frame with exact keys, the per-run nonce, and an enum-only reason. Malformed nonce, extra fields, and non-allowlisted values are rejected; runner output accepts only typed trusted termination evidence and hashes diagnostic prose.
- The final barrier change consumes its test-only environment once per pytest process before candidate execution, so repeated-process cases join the outer barrier once without exposing controls to the sandbox.
- Coordinator proc-fd sealing/release, nested status authentication, result nonce/aggregate/hash, ACK/EOF, cgroup placement/telemetry, cleanup, and the dedicated unsealed Playwright aggregate path are preserved.

Evidence:
- Cross-branch merge focus: 11 passed.
- Full deterministic supervisor: 312 passed, 30 skipped, 18 deselected.
- Focused Vitest protocol/reporting/workflow: 30 passed, 5 skipped, 136 deselected.
- Pylint 10.00/10; compileall and diff check passed; exact review worktree clean.

Residual uncertainty: local macOS cannot execute the real Linux/Yama/cgroup or concurrent systemd scope path. The exact remote SHA must pass the hosted Noble installed-wheel Vitest, concurrent scope construction, and seven installed-wheel Playwright cases before merge/release.

gltanaka:
Hosted validation remains blocking; do not merge.

Exact implementation content was validated through 5ad0f7e128dbcbf422098924288fbce7a6b8a5e8. Current branch head 80697c26a20af0aebc24220d1f5142f54ea252eb adds only the auditable failed-correction commit and its revert, so its tree is byte-identical to 5ad0f7e12.

RED evidence: run https://github.com/promptdriven/pdd/actions/runs/29557285417 passed the focused protected-runner suite, then failed the exact synchronized three-process Vitest test; installed Playwright passed and installed Vitest reported exit 125, scope-setup, launch-exit.

Noble correction evidence: run https://github.com/promptdriven/pdd/actions/runs/29558032325 advanced both source and installed-wheel Vitest beyond launch-exit but timed out after 30 seconds with reporter missing and no cgroup memory/pids deltas. The exact synchronized concurrency test failed the same acceptance boundary.

One bounded causal correction removed the live coordinator proc-fd bind seal. Run https://github.com/promptdriven/pdd/actions/runs/29559311703 disproved that hypothesis: focused tests and installed Playwright passed, but installed-wheel Vitest again timed out after 30 seconds with reporter missing/no cgroup deltas, and synchronized source concurrency failed. The correction was reverted non-destructively because it weakened defense without fixing acceptance.

Local evidence on the attempted correction was green: supervisor 312 passed/48 skipped; Vitest 166 passed/5 skipped; workflow contracts 22 passed; pylint 10/10; compile and diff checks clean. This does not override the authoritative hosted failures.

Next work should add bounded trusted stage telemetry to localize whether the post-drop PR_SET_PTRACER denial probe or subsequent Node/Vitest worker launch is the noncompleting stage. Do not increase timeouts, add retries, relax reporter validation, or remove ptrace/proc defenses without that evidence.

gltanaka:
### Exact-head revert restoration review

**Verdict: MERGE `80697c26a20af0aebc24220d1f5142f54ea252eb` — no findings in the restored tree, subject to mandatory hosted validation.**

The experimental `80c9908bc` change relaxed the security boundary by removing the coordinator `/proc/<pid>/fd` seal and unifying the inner status-supervisor topology. Exact head `80697c26a` explicitly reverts that commit. The entire repository tree is byte-identical to previously approved `5ad0f7e128dbcbf422098924288fbce7a6b8a5e8`: all three resolve to tree `5dedebc7dc9cc3ad8c4bb18b3b067cca4a98e6a5`, and both the full-tree diff and security-sensitive source/test diff are empty.

Therefore the coordinator proc-fd seal, worker preload, checked post-drop `PR_SET_PTRACER(0)` plus Yama/task/TID/pidfd denial probes, dedicated unsealed Playwright topology, framing, limits, termination evidence, and cleanup are restored exactly. Deterministic evidence: clean worktree, exact remote-head match, empty equivalence diffs, and clean `git diff --check`.

Residual merge gate: run the authoritative exact-head hosted Ubuntu Noble / Node 22.23.1 / Vitest 4.1.10 installed-wheel and synchronized-concurrency validation, including the seven installed-wheel Playwright cases. Tree equivalence restores the approved code boundary but does not supersede the unresolved hosted acceptance failure.

gltanaka:
## Escalation telemetry RED

RED commit: `2c87b34bc`

The exact focused contract failed 8/8 new tests before implementation:

- no typed Vitest progress enum or bounded transport parser existed;
- the post-drop wrapper emitted no probe-complete or candidate-exec boundary;
- the checker-owned reporter/preload emitted no coordinator, worker, collection, or result-publication boundaries;
- timeout evidence could not carry allowlisted progress while rejecting candidate-controlled text.

Command: `uv run pytest -q tests/test_sync_core_runner_vitest.py -k 'progress_transport or progress_sources or timeout_reports_only_allowlisted_progress' --maxfail=20`

This commit changes tests only. It does not change timeouts, retries, reporter adjudication, sandbox defenses, result access, workflow behavior, or cloud state.

gltanaka:
### Exact-head bounded-progress security review

**Verdict: MERGE `c64d44c79929643e310cafacd2ffcb8befc8c0c7` — no blocking findings, subject to mandatory exact-head hosted validation.**

Reviewed delta from approved/restored `80697c26a20af0aebc24220d1f5142f54ea252eb` through RED `2c87b34bcf062a11731c7c8dd617fbd35a236313` and GREEN `c64d44c7`.

Security boundary:
- `supervisor.py:2699,2711-2714` adds only fixed, value-free progress writes after the checked post-drop PR_SET_PTRACER/Yama/task/TID/pidfd denial probe and immediately before candidate exec. The supervisor delta is exactly these five additive lines; proc-fd sealing, timeouts, limits, cleanup, and termination authority are unchanged.
- `runner.py:4562-4638` admits only six enum stages, a 256-record bound, prerequisite ordering, one terminal result, no trailing/partial/unknown records, and no candidate values. `runner.py:4641-4746` places fixed emitters only in the checker-owned reporter/preload; the worker writes its stage before closing every matching result descriptor and before candidate modules load.
- `runner.py:4892-4965` lets progress decorate only non-pass infrastructure detail. Trusted sandbox errors still outrank any report; timeout remains TIMEOUT; PASS still requires the terminal reporter JSON to pass the existing identity/status/return-code validator. Candidate stdout/stderr remains hashed rather than disclosed.
- Playwright is unchanged: the generated `seal_cross_process=False` wrapper is byte-identical to `80697c26a` (465 bytes, SHA-256 `d67768d9dabfe6f7efd435755c0be74accb10bdcf31bde03314678472817ea39`). No retry, timeout, resource, pool, reporter, or adjudication relaxation was added.

Grammar observations are acceptable residuals, not blockers: `splitlines()` also normalizes CRLF, but it still maps only an exact allowlisted enum and cannot carry data or affect PASS. The raw-JSON compatibility branch is reachable by portable unit harnesses only; every production standard-anonymous Vitest stream writes `post-drop-probes` before candidate exec, so production bytes cannot begin with `{`, even if a later candidate stage somehow recovered write authority. Both rely on the already-reviewed anonymous-FD/proc/ptrace capability boundary and do not expand it.

Evidence:
- RED focused contract: **8 expected failures**.
- GREEN telemetry plus anti-forgery/cleanup/worker-FD focus: **11 passed**.
- Full deterministic Vitest adapter: **174 passed, 5 skipped**.
- Full deterministic supervisor: **312 passed, 30 skipped, 18 deselected**.
- Pylint: **10.00/10**; compileall, diff-check, exact ancestry/head, and clean-status checks passed.

Mandatory residual: macOS skips the real Linux/Yama/cgroup cases. The exact head must run on hosted Ubuntu Noble / Node 22.23.1 / Vitest 4.1.10 and demonstrate which trusted post-READY stage is last before the existing noncompletion. This telemetry approval does not resolve or supersede the hosted merge blocker.

gltanaka:
Hosted correction evidence: validation run 29562437322 package job 87827601875 failed closed with Vitest progress transport stage is out of order. This is a telemetry regression, not the underlying Vitest localization: real concurrency disproved the assumed total order, while the generic error hid which trusted stage raced. RED c2e451613 proves the rejected stream must retain bounded enum-only observed/failing evidence; GREEN 014873c2d adds only that safe diagnostic and leaves rejection, record/byte bounds, timeouts, retries, result adjudication, and defenses unchanged. Local: focused 7 passed; Vitest 175 passed/5 skipped; supervisor 312 passed/48 skipped; pylint 10.00/10; compile/diff clean. Validation-only PR #2140 now has exact implementation head as an ancestor at 03708e69e; rerun 29563424794. Do not merge #2140.

gltanaka:
### Exact-head typed-order delta review

**Verdict: MERGE `014873c2d5a6312ae2742361e988af0333aa8457` — no findings in the delta from approved `c64d44c79929643e310cafacd2ffcb8befc8c0c7`.**

Reviewed files: `pdd/sync_core/runner.py` and `tests/test_sync_core_runner_vitest.py`; `pdd/sync_core/supervisor.py` is byte-unchanged from the approved baseline.

At `runner.py:4610-4641`, arbitrary/non-ASCII stage bytes still fail with the fixed generic allowlist error before interpolation. The new out-of-order detail is assembled only after `VitestProgressStage(...)` succeeds, from the deduplicated observed enum list (maximum six values) and one failing enum. It cannot include candidate prose, paths, secrets, counts, or unbounded values. The parser still raises `ValueError` before terminal result handling, so malformed/concurrent ordering remains fail-closed and cannot create PASS, change trusted termination precedence, or grant result-channel authority. Duplicate-worker handling and every accepted ordering rule are unchanged.

Full security invariants remain those approved at `c64d44c7`: anonymous FD authority, worker preload closure, coordinator proc-fd seal, checked PR_SET_PTRACER/Yama/task/TID/pidfd denial, reporter/result validation, timeout/resource bounds, no retry, Playwright isolation, and cleanup are untouched.

Evidence:
- RED `c2e451613`: exact new test fails on the prior generic error.
- GREEN focused telemetry/timeout/anti-forgery/cleanup selection: **10 passed**.
- Adversarial probe: unknown `candidate-secret` stage remains generic and undisclosed; out-of-order known stages report enum values only.
- Pylint: **10.00/10**; compileall, diff-check, exact ancestry, clean status, and exact remote-head guard passed.

Residual uncertainty: this bounded diagnostic still requires the exact-head hosted Ubuntu Noble / Node 22.23.1 / Vitest 4.1.10 run to observe the real concurrent ordering/noncompletion. This delta approval does not resolve that hosted acceptance gate.

gltanaka:
Hosted topology evidence from run 29563424794 package job 87830622482: post-drop-probes,candidate-exec,coordinator-start then result-published; worker-start and collection-complete were not observed. This proves the typed records are process-topology observations, not a mandatory total lifecycle. Escalation REDs: 87372e076 covers a worker write racing before reporter construction and the hosted optional-stage gap; 25de70abf covers duplicate and post-publication regression rejection. GREEN db291ba9c enforces linear wrapper prerequisites, independent post-exec coordinator/worker paths, optional worker/collection observations, and terminal coordinator publication. Unknown/partial/trailing/duplicate/regressive streams remain fail-closed; bounds, timeout/retry/adjudication/defenses unchanged. Local: focused 11 passed; Vitest 179 passed/5 skipped; supervisor 312 passed/48 skipped; pylint 10/10; compile/diff clean. Validation-only #2140 head 4bea1ed0a contains exact db291ba9c; rerun 29564456678. Do not merge #2140.

gltanaka:
### Exact-head concurrent-progress graph review

**Verdict: MERGE `db291ba9cff13b91f2c847a9e83a16c2db8d2f64` — no blocking findings in the delta from approved `014873c2d5a6312ae2742361e988af0333aa8457`.**

Reviewed files: `pdd/sync_core/runner.py` and `tests/test_sync_core_runner_vitest.py`. Supervisor, proc/ptrace/FD security topology, worker preload, reporter emitters, and Playwright code are byte-unchanged from the approved baseline.

The graph at `runner.py:4614-4647` correctly models the trusted pipe topology:
- wrapper order remains linear: `post-drop-probes` then `candidate-exec`;
- after exec, coordinator and fork-worker writes are independent, so `worker-start` may precede or follow `coordinator-start`;
- worker and collection stages are optional observations, matching hosted run `29563424794`, while collection/publication still require coordinator start;
- `result-published` is terminal: every subsequent progress stage is rejected before prerequisite/duplicate handling, and a result frame is accepted only immediately after publication.

No authority or PASS path expanded. Stage bytes must still convert to the six-value enum; arbitrary values remain generic failures. Non-worker duplicates remain rejected, repeated worker records are deduplicated in exposed evidence, and the unchanged 256-record bound caps the stream. Partial, unknown, trailing, premature/duplicate result frames, regressive stages, and post-publication worker/collection records all fail closed. The terminal JSON still passes the existing identity/status/return-code validator before PASS.

Evidence:
- RED `25de70abf`: focused graph suite **3 failed / 8 passed**, reproducing the two valid concurrent paths and post-publication rejection gap.
- GREEN focused progress/timeout/anti-forgery/cleanup: **14 passed**.
- Adversarial transition matrix covered both race orders, optional gaps, all prerequisite failures, every non-worker duplicate, stages after publication, premature/duplicate results, and the exact record ceiling: passed.
- Pylint: **10.00/10**; compileall, diff-check, exact ancestry/head, clean status, and exact remote-head guard passed.

Residual uncertainty: exact-head hosted Ubuntu Noble / Node 22.23.1 / Vitest 4.1.10 integration remains mandatory. This local graph approval does not resolve the underlying hosted noncompletion or replace that gate.

gltanaka:
## Exact post-#2138 integration remains blocked

Fresh never-merge validation #2143 combined exact current main `0c7a78e2b` with reviewed #2128 `db291ba9c` at merge `2772d7ddc`: https://github.com/promptdriven/pdd/actions/runs/29566177644

The integration passed sandbox/cgroup, protected pytest, signer, held-namespace FD cleanup, repo-bloat, public CLI, story, and auto-heal. It contains none of the discredited maxWorkers/V8/libuv caps.

Both independent real-Vitest boundaries then timed out at the unchanged 30-second child bound after trusted `post-drop-probes,candidate-exec,coordinator-start,result-published`, with zero OOM/OOM-kill/pids-max deltas:
- fresh source gate: diagnostic `1807a773…`;
- isolated installed wheel: diagnostic `d6d062ac…`.

This is post-reporter-publication noncompletion, not #2083 SIGABRT and not the stale capped #2140 tree. The current `reporter=missing` label is inconsistent with the trusted stage but retained evidence cannot prove the following result-frame write completed; timeout remains fail-closed. #2143 was closed unmerged and its ref deleted. Full evidence: https://github.com/promptdriven/pdd/pull/2143#issuecomment-5000775434

Do not merge #2128 yet. Next work must causally localize/fix coordinator post-result liveness without retry, timeout/resource increase, PASS reinterpretation, or security weakening.

gltanaka:
Hosted topology result from temporary PR #2147 (closed unmerged): https://github.com/promptdriven/pdd/actions/runs/29569468307

Exact composition was main `0c7a78e2b9da296318dce83c2be03fda7dc3344b` + this PR head `db291ba9cff13b91f2c847a9e83a16c2db8d2f64` + reviewed topology diagnostic, yielding validation head `afe7b289e9fc2cbf4e45b2e97072c5a4e716bc57`. Auto-heal and CodeQL passed without head movement.

Authenticated Vitest timeout evidence: reporter entered, terminal frame present, progress reached `result-published`, all cgroup deltas were zero, and `pre_kill_process_topology=one`. The one remaining process is the Vitest coordinator; no fork worker descendant remained. The hang is therefore after successful result publication in coordinator shutdown, not candidate execution, worker liveness, OOM, or PID exhaustion.

The same topology delta caused a separate installed-wheel Playwright regression: all 7 suffix cases blocked in descriptor-frame polling until the 90-second outer pytest timeout. The diagnostic branch is being corrected under the PR loop before any new integration validation. No retry or gate weakening is appropriate.

gltanaka:
Exact #2144 + #2128 hosted integration #2149 reached a new release-blocking state: the real Vitest gate exits in 15.16s (so the post-result timeout is fixed) but the authenticated terminal payload contains zero test identities and fails closed as `NOT_COLLECTED`. Held-namespace/FD-only cleanup passed 6/6. Evidence: https://github.com/promptdriven/pdd/pull/2149#issuecomment-5001948175

This is not the old timeout or #2083 SIGABRT. It is an ordering issue between concurrent reporter case-result delivery and immediate coordinator exit. No retry/timeout/cap/preload/adjudication weakening is planned; a focused PR-loop correction is being designed.

gltanaka:
Link correction for the exact #2149 terminal evidence comment: https://github.com/promptdriven/pdd/pull/2149#issuecomment-5002206397

gltanaka:
## Terminal module-result fix merged

PR #2150 merged into this branch at `95f92c2ea8af5762005d4592135e8d1188afcfb7` after exact-head critical review found no P1/P2 issues. Local adapter evidence: 194 passed, 6 skipped; hosted staging checks green. Review: https://github.com/promptdriven/pdd/pull/2150#issuecomment-5002696001

This fixes the hosted zero-identity cause by retaining awaited `onTestModuleEnd` snapshots. Remaining release gate is the ancestry-preserving restack through #1995 → #2097 → #2108 → #2128 and unchanged-head hosted source/wheel Vitest, Playwright, and FD/namespace validation. During #2144 composition, coordinator exit must be derived from authenticated snapshots/errors rather than Vitest’s spurious empty-terminal `process.exitCode=1`.

PR #2143, opened 2026-07-17
test(sync): validate exact post-cap Vitest integration

gltanaka:
## Never-merge hosted validation

This temporary PR validates the exact security integration after merged #2138. It must never be merged.

- exact base: `0c7a78e2b9da296318dce83c2be03fda7dc3344b`
- exact #2128 head: `db291ba9cff13b91f2c847a9e83a16c2db8d2f64`
- exact integration merge: `2772d7ddcfd788fdf1b06a3452e3fa3f73a4e72a`
- merge parents are the two exact inputs above
- discredited `--maxWorkers=1`, `--v8-pool-size=1`, and forced `UV_THREADPOOL_SIZE=1` are absent
- 30-second child timeout, 4 GiB physical/default virtual bounds, forks/preload, proc/ptrace/Yama/FD defenses, typed bounded progress, reporter fail-closed semantics, one attempt/no retry, held-namespace smoke, and locked fresh-interpreter Vitest gate are retained

Local evidence: 850 implementation tests passed, 51 skipped, 18 deselected; independent review added 113 passes and one Linux-only skip; pylint 10/10; AST/YAML/compile/diff/clean/exact-parent guards passed.

Hosted acceptance requires installed-wheel Vitest, concurrent Vitest construction, Playwright 7/7, protected Linux namespace/memfd/cgroup gates, focused runner, package/offline smoke, and the full unit suite. Close unmerged and delete this ref after terminal evidence.

PR #2144, opened 2026-07-17
fix(sync): terminate Vitest after trusted result publication

gltanaka:
## Summary

- write the complete checker-owned Vitest result through a checked synchronous `writeAll` loop
- bind coordinator `process.exit` when the sole trusted reporter module loads
- exit only after result publication, preserving Vitest's existing nonzero `process.exitCode`
- keep all timeout, retry, adjudication, sandbox, and supervisor cleanup policy unchanged

Fixes #2145.

Related, but explicitly **not closed by this PR**: #2083. That issue's pre-result `reporter=missing; SIGABRT` failure and approved sealed Node 22.22.3/22.23.1 causal A/B remain unresolved. This PR is only the prerequisite post-result shutdown-liveness correction.

## Root cause evidence

Three unchanged-head Package attempts on validation run [29564456678](https://github.com/promptdriven/pdd/actions/runs/29564456678) produced the same typed sequence and timeout:

1. [job 87833848158](https://github.com/promptdriven/pdd/actions/runs/29564456678/job/87833848158): `post-drop-probes,candidate-exec,coordinator-start,result-published`; timeout 30s; all cgroup deltas 0
2. [job 87837179277](https://github.com/promptdriven/pdd/actions/runs/29564456678/job/87837179277): identical typed sequence and outcome
3. [job 87840247129](https://github.com/promptdriven/pdd/actions/runs/29564456678/job/87840247129): identical typed sequence and outcome

Exact Vitest 4.1.10 package source shows `_testRun.end` awaits trusted `onTestRunEnd` before CLI `finally` awaits `ctx.close()`. `ctx.close()` awaits pool closure, while fork-worker graceful stop has a hard-coded 60s stop timeout. PDD's unchanged protected deadline is 30s. Successful typed result publication followed by the 30s deadline therefore localizes the defect to post-report coordinator/pool shutdown, not test execution, collection, or result transport.

## Safety invariants

- candidate config cannot add `reporters` or enable `coverage`; the command supplies exactly one checker-owned reporter
- candidate tests execute in fork workers, separate from the coordinator where exit is bound
- incomplete, zero-progress, or impossible writes throw before exit
- Vitest-set nonzero exit codes remain nonzero
- the supervisor retains ownership of worker/descendant cleanup after coordinator exit

## TDD evidence

At RED commit `18be57336`:

- held post-result interval: timed out after 2s
- forced 3-byte writes: reporter emitted only `{"t`
- zero-progress write: incorrectly returned exit code 0
[...]

PR #2146, opened 2026-07-17
fix(sync): authenticate Vitest timeout topology

gltanaka:
## Bounded timeout topology diagnostic

This sub-PR targets #2128 and does not change adjudication. It adds typed evidence required by the exact cap-free #2143 failures.

TDD:
- RED `d5c8da289f1dad59dca2ac2b4f9220cdfef97015`: tests-only contract; collection failed because typed topology did not exist.
- GREEN `08f0852d9528e7af002d5354e1dd7b7a6ddece2e`: exact four-file implementation.

Behavior:
- root-owned helper samples the owned candidate cgroup immediately before timeout kill and emits only `none|one|multiple`;
- timeout records require the enum; non-timeout records reject it;
- runner reports reporter entry only from allowlisted progress and terminal-frame presence only after strict framed parsing;
- TIMEOUT is returned before any JSON semantic adjudication and can never become PASS;
- no PIDs, counts, paths, argv, environment, or candidate prose are exposed.

Non-goals: no retry, timeout/resource increase, reporter reinterpretation, proc/ptrace/FD/preload/Playwright weakening, or causal fix yet.

Local evidence: supervisor 318 passed/48 skipped; non-real Vitest 184 passed/5 skipped; Playwright 264 passed/16 skipped; focused final 18 passed/4 Linux-only skipped; pylint 10/10; compile/diff/clean/live-head guards passed. Independent critical review added full Vitest 185 passed/5 skipped and supervisor adversarial 31 passed/4 Linux skips.

Hosted purpose: observe terminal-frame state and pre-kill candidate topology on the known post-`result-published` timeout, then make the smallest causal correction.

### DIFF
--- a/pdd/sync_core/runner.py
+++ b/pdd/sync_core/runner.py
@@ -55,6 +55,7 @@ from .supervisor import (
     InfrastructureFailurePhase,
     ImmutableBindingProof,
     PlaywrightSnapshotAggregate,
+    ProtectedProcessTopology,
     ScopeSetupFailureReason,
     SnapshotBindingProof,
     SupervisorLimits,
@@ -4491,19 +4492,31 @@ def _vitest_result(
 def _vitest_infrastructure_termination(
     result: subprocess.CompletedProcess[str], timeout_seconds: int,
     *, progress: tuple[object, ...] = (),
+    terminal_frame_present: bool = False,
 ) -> tuple[EvidenceOutcome, str]:
-    """Describe trusted no-reporter termination without trusting stderr prose."""
+    """Describe trusted Vitest termination without trusting stderr prose."""
     termination = getattr(result, "termination", None)
     kind = getattr(getattr(termination, "kind", None), "value", None)
     if kind is None:
         kind = "timeout" if result.returncode == 124 else (
             "signal" if result.returncode < 0 else "exit"
         )
-    fields = ["Vitest infrastructure termination: reporter=missing", f"kind={kind}"]
     trusted_progress: list[str] = []
     for stage in progress[:len(VitestProgressStage)]:
         if isinstance(stage, VitestProgressStage) and stage.value not in trusted_progress:
             trusted_progress.append(stage.value)
+    reporter_entered = VitestProgressStage.COORDINATOR_START.value in trusted_progress
+    terminal_present = (
+        terminal_frame_present is True
+        and VitestProgressStage.RESULT_PUBLISHED.value in trusted_progress
+    )
+    fields = [
+        "Vitest infrastructure termination: reporter=" + (
+            "entered" if reporter_entered else "not-entered"
+        ),
+        "terminal_frame=" + ("present" if terminal_present else "absent"),
+        f"kind={kind}",
+    ]
     if trusted_progress:
         fields.append("trusted_vitest_progress=" + ",".join(trusted_progress))
     if kind in {"exit", "sandbox-error"}:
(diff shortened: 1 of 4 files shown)


gltanaka:
Hosted consumer result from temporary PR #2147: https://github.com/promptdriven/pdd/actions/runs/29569468307

The topology field worked for the real Vitest timeout and localized it to the single coordinator after `result-published`. However, the same diagnostic delta introduced a release-blocking installed-wheel Playwright regression: all 7 config-suffix cases blocked in `_read_descriptor_frame_fd` / `_descriptor_poll` and hit the 90-second outer pytest timeout. The prior exact composition without this PR delta passed that package gate.

This PR must remain unmerged until a TDD correction preserves timeout topology for Vitest without changing the successful Playwright descriptor protocol. Do not retry or weaken outer/security gates.

gltanaka:
## Correction-cycle investigation and TDD evidence — 5255f2daf

**Investigation**
- Sibling descriptor supervision uses one helper for standard-anonymous Vitest and Playwright aggregate mode, but only Vitest needs candidate-cgroup topology telemetry.
- Prior framing commit `3923a1024` established the standard-anonymous protocol distinction; the `80c9908bc` / `80697c26a` pair confirms this correction must not weaken restored proc/FD sealing.
- Root cause: `08f0852d9` unconditionally required a candidate cgroup during pidfd timeout and unconditionally extended the descriptor candidate schema, regressing Playwright timeout kill/reap and its established four-field terminal record.

**TDD**
- RED `d10d02ca5`: 3 failed, 1 Linux-only skipped. Failures proved topology-free timeout raised before kill, helper mode selection was unconditional, and descriptor parsing lacked trusted protocol context.
- GREEN `5255f2daf`: pidfd timeout always kills/reaps/closes; topology is collected and serialized only for `standard_anonymous` Vitest. Descriptor parsing requires topology for standard Vitest and rejects it for Playwright aggregate mode.

**Validation**
- `tests/test_sync_core_supervisor.py`: 326 passed, 49 skipped.
- Non-real Vitest: 185 passed, 5 skipped.
- Non-real Playwright: 264 passed, 16 skipped.
- Supervisor pylint: 10.00/10; touched-test pylint completed with existing baseline diagnostics only.
- Compileall, diff-check, clean-tree, exact ancestry, and remote-head guards passed.

No timeout, retry, authentication, cgroup cleanup, proc/ptrace, FD, preload, or reporter-adjudication contract changed.

gltanaka:
Independent critical re-review of exact head `5255f2daf1c4dc9222eeb9ca818ca3b61928f408`: **MERGE — no correction blocker found.**

The two-commit delta binds timeout topology to the authenticated protocol mode: standard-anonymous Vitest requires the bounded enum, while a validated Playwright aggregate rejects it and retains its four-field timeout record. Playwright topology-free timeout still kills immediately, reaps exactly once, and closes the pidfd; nonce/digest/schema authentication, timeout precedence, retry/limits, ptrace/FD protections, and cleanup remain unchanged.

Evidence: focused mode/schema/Vitest/descriptor tests passed; full Playwright cross-contract run reached `277 passed, 21 skipped` with one unrelated failure, `test_playwright_external_node_modules_environment_is_available`, which reproduces identically on the previously reviewed pre-correction composition and is therefore not introduced by this delta. `compileall`, pylint, diff-check, static weakening guards, clean status, and exact live head/base checks passed. Residual uncertainty is the expected macOS gap for Linux-only real pidfd/cgroup/hosted sandbox tests.

gltanaka:
Closing unmerged as completed diagnostic work after the direct causal fix merged in #2144.

#2146 successfully authenticated the prior failure as a post-`result-published` timeout with only the Vitest coordinator remaining. #2144 now fixes that exact lifecycle by completing the trusted reporter write and terminating the coordinator after publication while preserving nonzero `exitCode` and supervisor-owned descendant cleanup.

This PR adds localization only; it does not alter adjudication, close the distinct pre-result SIGABRT in #2083, or provide a security property required by #2128. Shipping its four-commit runner/supervisor timeout-schema expansion would add protocol and Linux cgroup/pidfd test surface without release value. The correction cycle and critical review remain recorded here, including the Playwright regression and fix.

Decision: retain #2128 worker FD isolation/preload, retain #2144 as the shutdown fix, do not remove the preload, and close #2146 unmerged.

PR #2164, opened 2026-07-17
fix(sync): seal Vitest result authority before workers

gltanaka:
Closes #2083.

Root cause: the hosted A/B diagnostic proved that removing the Vitest worker preload restores lifecycle callbacks but exposes the protected reporter pipe to fork workers via direct, alias, and cross-process descriptor paths.

This change preserves the reporter authority in the Node coordinator and adds a checker-owned Linux N-API seal at reporter `onTestRunStart`, before Vitest creates its fork pool. The seal validates the expected FIFO identity and applies `FD_CLOEXEC` to every matching descriptor alias. It fails closed if validation/sealing cannot complete, removes the worker `--execArgv` preload, and packages the dependency-free addon into the wheel through an offline build hook.

Validation before PR:
- 20 focused coordinator/attestation/forge/lifecycle tests passed
- runner, build hook, and setup lint passed
- wheel build succeeded and contains the addon payload

Required hosted evidence: Linux Node 22 source and installed-wheel real-Vitest lifecycle; real fork/exec fd+alias isolation; existing forge/missing-reporter behavior; no preload command shape.

No release/deploy action is included.

Issue #2159, opened 2026-07-17
build: isolate Make bootstrap from inherited cross-worktree runtime PATH


gltanaka:
## Observed

In a PDD worktree after `direnv allow`, `CONDA_PREFIX` correctly ends in `pdd`, but `python`, `pip`, `conda run -n pdd python`, and `conda run -n pdd pip` all resolve through inherited `/private/tmp/gvs-.../.codex-runtime/...` entries placed ahead of PDD in `PATH`. Unsetting `VIRTUAL_ENV` does not correct it.

This makes Make bootstrap (`ensure-dev-deps`) install/check dependencies against another worktree runtime, yielding flaky or misleading lint/test results.

## Desired behavior

Make/bootstrap commands must resolve the selected Conda environment interpreter deterministically, independent of inherited `PATH` or `VIRTUAL_ENV` state. Prefer `$CONDA_PREFIX/bin/python -m pip` after validating the selected environment, or another explicit Conda-env executable resolution; avoid bare `pip`/`python` lookup.

## Acceptance criteria

- A regression test simulates a foreign runtime prepended to `PATH` and confirms PDD bootstrap uses PDD's interpreter/pip.
- `make lint`/dependency bootstrap remains deterministic in a fresh PDD worktree with inherited runtime entries.
- No secret/runtime paths are emitted in normal logs.

Discovered while validating PDD #2083; unrelated to its Vitest authority diagnosis.

gltanaka:
Additional reproduction detail: even `conda run -n pdd python` and `conda run -n pdd pip` can resolve through the foreign runtime, so the fix must avoid relying on ambient PATH after Conda activation. See duplicate #2160 for the independently captured reproduction.

gltanaka:
Additional reproduction during the #2083 Linux N-API validation worktree: after `direnv allow`, a broad test launch resolved `pytest` from a foreign GVS `.codex-runtime` on PATH rather than the selected PDD environment. The worker stopped only its own three duplicate processes and is continuing with the explicit PDD interpreter. This reinforces the requested bootstrap rule: never rely on inherited PATH for Python/pytest/pip in a worktree; resolve through the selected environment interpreter. No source or release artifact contamination was observed.

gltanaka:
Additional release-worktree reproduction (2026-07-18): inherited cross-worktree runtime precedence can survive `conda run -n pdd python`.

[...]

PR #2173, opened 2026-07-17
fix(sync): exercise native Vitest authority tests

gltanaka:
## Summary

Follow-up to #2164 for the hosted-only gate failures.

- preserves production coordinator authority code and source-only wheel rules; changes only the test harness and build-test dependency declaration.
- uses pytest `originalname` so parametrized native-addon negative tests no longer receive the text placeholder.
- gives Linux reporter tests real selected-Node headers, while retaining platform skips.
- makes the no-rehash attestation test assert each staged/attestation mutation happened.
- declares the existing `setuptools-scm>=8` build-system requirement in the dev extra used by the no-isolation wheel smoke.

## Root cause

The exact #2164 hosted run `29614087280` failed because the fixture compared parameterized `request.node.name` (for example `[wrong-identity]`) to an unparameterized allow-list, causing Node to load `portable test authority` and report `file too short`; reporter tests compiled the fake descriptor headers; and wheel smoke invoked `build --no-isolation` without the dynamic-version build dependency in the dev environment. The separate protected-ownership prerequisite for the three tracked paths is intentionally excluded: it must merge first and cannot self-authorize in this PR.

## Evidence

- Red before fix: `test_vitest_authority_wheel_build_tooling_is_available_without_isolation` failed; committed separately as `416804d02`.
- Green after fix: 5 targeted tests passed (wheel tooling, source-only wheel, phase-bound attestation, candidate aliases, no-rehash mutations).
- `compileall`, `py_compile`, `python -m pdd --help`, and `git diff --check` passed.
- Local macOS skips Linux-only live addon execution; next hosted Linux run is required for reporter/native artifact confirmation.

## Non-goals

No production authority relaxation, policy self-authorization, CI timeout change, merge, deploy, release, or tag.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,6 +124,7 @@ dev = [
     "pytest-asyncio",
     "pytest-timeout",
     "build",
+    "setuptools-scm>=8",
     "twine"
 ]
 
--- a/tests/test_sync_core_runner_vitest.py
+++ b/tests/test_sync_core_runner_vitest.py
@@ -535,6 +535,14 @@ def test_vitest_authority_wheel_is_source_only(tmp_path: Path) -> None:
     assert not any(name.endswith(".node") for name in names)
 
 
+def test_vitest_authority_wheel_build_tooling_is_available_without_isolation() -> None:
+    """Keep the source-only wheel smoke test's build requirements explicit."""
+    repository = Path(__file__).resolve().parents[1]
+    project = tomllib.loads((repository / "pyproject.toml").read_text(encoding="utf-8"))
+
+    assert "setuptools-scm>=8" in project["project"]["optional-dependencies"]["dev"]
+
+
 def test_vitest_coordinator_addon_rejects_unsupported_platform(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch,
 ) -> None:
--- a/tests/test_sync_core_runner_vitest.py
+++ b/tests/test_sync_core_runner_vitest.py
@@ -264,9 +264,10 @@ def _controlled_supervisor(
         "test_vitest_coordinator_precompile_requires_phase_bound_header_attestation",
         "test_vitest_coordinator_precompile_rechecks_phase_attestation_without_rehash",
     )
+    test_name = request.node.originalname or request.node.name
     if (
         not request.node.name.startswith("test_real_vitest_runs_copied_entrypoint")
-        and request.node.name not in native_authority_tests
+        and test_name not in native_authority_tests
     ):
         # The production authority deliberately rejects unsupported platforms
         # and requires a real Node distribution. Most adapter contracts use a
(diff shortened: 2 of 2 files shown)


gltanaka:
Investigation evidence:

- Hosted #2164 run `29614087280`: parameterized `wrong-identity`/`non-fifo` tests loaded a fixture text placeholder (`file too short`); reporter cases compiled fake descriptor headers; wheel smoke lacked `setuptools-scm` under `--no-isolation`; ownership rows are a separately merged prerequisite.
- Red TDD: `416804d02` added the no-isolation build-tool assertion; it failed before the fix.
- Green: 5 targeted Vitest/wheel tests passed; `compileall`, `py_compile`, `python -m pdd --help`, and `git diff --check` passed.
- Remaining platform evidence: macOS skips Linux-only live addon cases, so confirm the real artifact/reporter cases in this PR’s hosted Linux run. No production authority behavior was relaxed.

gltanaka:
## Codex adversarial review — PR #2173 @ b84b60ad95e3e62f27ef625bae1c4115cd961af4

**Verdict:** DON'T MERGE
**Findings:** 1 total (1 P1, 0 P2)

### P1 (blocking)

- [P1] Rebase the stacked head onto the current #2164 base — `tests/test_sync_core_runner_vitest.py:252-286`
  GitHub reports this PR as `CONFLICTING`/`DIRTY`: head `b84b60a` forked from `13a851f`, while its stated base is now `5d6753b`, and both modify the autouse native-authority fixture. It therefore cannot merge or receive a meaningful required Unit Tests result. The base independently normalizes a missing `request.node.originalname` with `request.node.name.split("[", 1)[0]`; retaining only the head's fallback would reintroduce the parameterized-name mismatch that caused the hosted `wrong-identity`/`non-fifo` tests to load the text placeholder rather than the real addon.
  **Suggested fix:** Rebase/cherry-pick the two #2173 commits onto `5d6753b`, resolve this hunk by retaining #2164's `originalname` then bracket-stripping fallback plus #2173's real-header/wheel/attestation changes, and rerun the hosted Linux authority tests.

### P2 (non-blocking)

- None.

### Notes

- Reviewed exact head, TDD commits `416804d`/ `b84b60a`, the failed hosted run `29614087280` job `87995195301`, and the setuptools-scm convention from #1052/`725d969`.
- Against the shared merge-base `13a851f`, the change keeps the production runner untouched; the source-only wheel rule remains source `.c` included and native `.node` excluded. The dev extra now supplies the pre-existing build-system `setuptools-scm>=8` requirement used by the no-isolation wheel smoke.
- Focused validation with the explicit PDD interpreter: 5 passed, 7 expected macOS Linux-only skips; `compileall` and `git diff --check` passed. The real selected Node headers resolve to the running Node distribution. The current shell's inherited foreign GVS Python lacks `setuptools-scm`, so its no-isolation wheel failure is environmental and is separately tracked by PDD #2159, not a PR regression.

PR #2179, opened 2026-07-18
fix(sync): close Vitest authority release gates

gltanaka:
Supersedes #2177 without rewriting its reviewed history.

Carries the reviewed #2164 Vitest authority fix and adds the PR-loop correction for the sole remaining hosted Unit Test failure.

The source-only wheel test now:
- excludes .git from its temporary source snapshot
- supplies a deterministic setuptools-scm test version
- preserves subprocess output on failure

Evidence:
- prior hosted red: #2177 Unit Test run 29620462195, job 88014334032
- focused test: passed 5 consecutive runs
- xdist (-n 2): passed

Issue: #2164

PR #2180, opened 2026-07-18
fix(sync): close portable Vitest authority release gates

gltanaka:
Supersedes #2179 without rewriting reviewed history.

Carries the reviewed #2164 Vitest authority fix and source-only wheel-test isolation correction. This successor also fixes a release-blocking macOS path issue found by the required full local module run: tempfile may return a path beneath /var, which is a symlink to /private/var. The hardened header policy correctly rejects symlink ancestry, so trusted protected-base and checked-head temp roots are now canonicalized before cloning.

PR-loop evidence:
- red: full local module had 15 failures with "Vitest header ancestor is not a regular directory"
- focused former failures: 3 passed
- deterministic symlink-root regression: passed
- full module: 189 passed, 13 skipped
- source-only wheel test: passed 5 consecutive runs and xdist -n 2
- git diff --check: passed

The symlink-ancestry security policy remains unchanged.

Issue: #2164


gltanaka:
Superseded by #2181 without rewriting this PR history; closing to prevent accidental merge.

PR #2181, opened 2026-07-18
fix(sync): close all Vitest authority release gates

gltanaka:
Supersedes #2180 after its security review identified one incomplete trusted-temp-root canonicalization.

Carries the reviewed #2164 Vitest authority fix, source-only wheel-test isolation, and portable trusted-temp-root corrections. All trusted Vitest temp roots are canonicalized before use:
- protected-base clone
- checked-head clone
- coordinator execution staging

PR-loop evidence:
- hosted red from #2177: source-only wheel test failed with hidden build diagnostics
- local red: full module had 15 macOS failures with "Vitest header ancestor is not a regular directory"
- focused wheel test: passed 5 consecutive runs and xdist -n 2
- deterministic regression aliases all four base/head clone and staging allocations, asserts addon staging is not a symlink: passed
- focused former failures: 3 passed
- final full module: 189 passed, 13 skipped
- git diff --check: passed

Security properties:
- TemporaryDirectory creates the roots; candidate state cannot select them
- candidate path handling is unchanged
- no-follow, inode/provenance, ownership, and ancestry checks are unchanged
- alias canonicalization prevents false rejection; it does not broaden accepted candidate paths

Issue: #2164

PR #2182, opened 2026-07-18
fix(sync): close Vitest authority and wheel release gates

gltanaka:
Supersedes #2181 after its final broad Unit gate exposed one remaining test-environment dependency audit.

Carries the reviewed #2164 Vitest authority and portable trusted-temp-root fixes. The sole new delta adds python-build's --skip-dependency-check to the source-only wheel-content test. The test still:
- runs the real build backend with --no-isolation
- requires successful wheel creation
- inspects the wheel for packaged C source
- rejects host-native .node artifacts

Why: the broad Unit environment intentionally does not install build-system requirements setuptools-scm and wheel, while the dedicated Package Preprocess Smoke job installs/builds the real package and passed. Dependency availability is not the assertion of this content test.

PR-loop evidence:
- hosted red #2181 run 29623536139: 14,517 passed; sole failure missing setuptools-scm>=8 and wheel during build dependency check
- reproduced locally with a Python runtime missing those build requirements
- corrected test passed in that dependency-missing runtime
- xdist -n 2 passed in pdd environment
- full Vitest module on prior exact source: 189 passed, 13 skipped
- all #2181 focused authority, package, CodeQL, story, CLI, bloat, and auto-heal gates passed
- git diff --check passed

Issue: #2164


gltanaka:
Closing before merge: concurrent #2164 landed on main with the wheel dependency-check correction plus newer native-signing authority. This branch is now history-divergent and must not overwrite main. Restacking only the missing trusted-temp canonicalization on current main.

Issue #1926, opened 2026-07-08
Transactional fingerprint finalization: enforce 'artifact write ⇒ fingerprint write' in one code path


gltanaka:
## Problem

Fingerprint write-back is the single largest sync failure class in this tracker: ~24 issues across 4 near-identical waves (#911–#913, #953–#956, #986 + #988–#990, #1005–#1007, plus #437, #430, #1211 → #1252 → #1290, #1305, #1021, #869, #871, #1106, #1317).

The mechanism is always the same: **every mutating command (sync, generate, example, update single-file, update --all, auto-deps, fix, CI heal) re-implements "save fingerprint on success" as a scattered side effect**, and every new code path either forgets it, writes null hashes (#437, #1305), skips it on some branches (#430), or writes to the wrong root (#1211, whose fix added kwargs that were never threaded — #1290). The consequences compound:

- Finished work re-flags as stale forever (consumer evidence: 377/788 fingerprints stale in pdd_cloud, promptdriven/pdd_cloud#2994; PDD's own PR automation shipped without fingerprints, promptdriven/pdd_cloud#1800).
- Heal loops re-trigger on their own output (#1305: CI auto-heal looped forever because example wrote null fingerprints).

Each fix has been local to one path, so the class regenerates with every new mutating feature.

## Proposal

Make "artifact write ⇒ fingerprint write" an enforced invariant with exactly one implementation:

1. A single finalization API (e.g. a context manager / `FingerprintTransaction`) that owns: compute hashes with the canonical algorithm, write artifact(s), persist fingerprint atomically (temp file + rename), with `project_root` and unit identity resolved once at entry.
2. All mutating paths (sync, generate, example, update, fix, auto-deps, CI heal) route through it. A mutating command **may not report success unless the fingerprint persisted** — failure to finalize is a command failure, not a warning.
3. A property-test harness in CI: for every mutating command, run it against a fixture repo and assert afterwards that `sync_determine_operation(log_mode=True)` reports `nothing`/synced for the touched unit. Any future mutating path that forgets finalization fails this suite by construction.

## Acceptance criteria

- [ ] One shared finalization code path; `save_fingerprint` call sites outside it are removed or delegated.
- [ ] Atomic persist (no partially-written/null-hash fingerprints observable on crash).
- [ ] Success exit requires persisted fingerprint; otherwise non-zero exit with explicit diagnostics.
- [ ] Post-command freshness property test covering every mutating command, running in CI.
[...]

Issue #254, opened 2026-01-04
pdd update (single file mode) incorrectly places output file


jamesdlevine:
Working on the pdd_cap:

Command: pdd update pdd/commands/generate.py 

* Prompt original location: 
  * prompts/commands/generate_python.prompt
* Updated prompt output location:    
  * prompts/generate_python.prompt

Core file manually attached due to error with report-core:

[pdd-core-20260104T054838Z.json](https://github.com/user-attachments/files/24420336/pdd-core-20260104T054838Z.json)

PR #2195, opened 2026-07-18
feat(sync): finalize fingerprints transactionally

Serhan-Asad:
Refs #1926

Part of the isolated Agentic Sync V1 integration program. This draft targets
`integration/agentic-sync-v1`; it must not be merged to `main` independently.

## What changed

- Add one `FingerprintTransaction`/`finalize_fingerprint` implementation that
  computes canonical hashes, rejects null prompt hashes, and commits via a
  same-directory temporary file plus `os.replace`.
- Make finalization and stale-run-report clearing failures typed command failures
  instead of warning-and-success paths.
- Delegate existing sync, operation-log, update, auto-deps, and pin-example
  finalization paths to the shared implementation, including buffered atomic
  state.
- Preserve the previous valid fingerprint on write/replace failures and keep
  subproject/root resolution anchored to the touched prompt.
- Add focused transactional, wrong-root, null-hash, stale-report, Click non-zero,
  and post-command freshness regressions.

## Validation

- `env PYTHONPATH="$PWD" conda run -n pdd python -m pytest -q tests/test_transactional_finalizer.py tests/test_auto_deps_main.py tests/test_operation_log.py tests/test_pin_example_hack.py`
  — 78 passed, 1 warning.
- `env PYTHONPATH="$PWD" conda run -n pdd python -m py_compile pdd/fingerprint_transaction.py pdd/auto_deps_main.py pdd/operation_log.py pdd/pin_example_hack.py pdd/sync_orchestration.py pdd/update_main.py`
  — passed.
- `git diff --check 131f86d83e7f2058af861b8ee7bde432bbbf5027..HEAD`
  — passed.
- Repository-wide Make test/lint/coverage/build targets could not bootstrap in
  the worker because package-index access was unavailable. Full-suite and package
  validation remain integration-gate items.

## Mutating-path inventory

Changed/delegated paths:

- `operation_log.save_fingerprint` is now the compatibility entry point and
  delegates to `finalize_fingerprint`; the generate/example decorator therefore
  fails if finalization fails.
- sync atomic state and `_save_fingerprint_atomic` delegate to the transaction.
- update single-file and repository modes delegate and no longer swallow failure.
- auto-deps delegates and requires stale report clearing before success.
- pin-example's private fingerprint helper is a thin delegate.

Remaining production calls to `save_fingerprint` in `sync_main.py`,
`metadata_sync.py`, and `ci_drift_heal.py` resolve through the same delegated
compatibility entry point. Calls retained in `auto_deps_main.py` and
`update_main.py` do the same. The remaining `Fingerprint(...)` uses in
`sync_determine_operation.py`/`continuous_sync.py` are state construction or
deserialization, not direct JSON persistence. An independent reviewer will verify
this inventory and identify any missed mutating path.

## Shared-file edits

- `pdd/operation_log.py`, `pdd/sync_orchestration.py`, `pdd/update_main.py`,
  `pdd/auto_deps_main.py`, and `pdd/pin_example_hack.py` are shared mutation/
  metadata paths.
- Matching prompt specifications were updated for every changed prompt-owned
  runtime module plus the new finalizer prompt.

## Migration risks

- Null-hash/path-resolution and stale-run-report-clear failures now propagate as
  `FingerprintFinalizeError`; callers that relied on warning-and-success behavior
  will observe a non-zero/typed failure by design.
- Concurrent writers still require the repository's existing locking discipline;
  atomic replacement prevents partial files but is not a multi-writer transaction.
- The broad every-command CI invariant and PDD artifact completeness require
  exact-head review before integration promotion.

## Rollback

Revert this PR's two-commit series from the integration branch. This restores the
previous per-call-site persistence behavior and its warning semantics; no data
migration is required because the fingerprint JSON shape is unchanged.


Serhan-Asad:
Correction for formal round 1: rebased the transactional-finalizer series onto frozen base `2cacc91f90759ff45f1ad976da3b773e1a5f07a5` and fixed the canonical-path regression in `test_update_preserves_subdirectory_structure_issue_254`. An explicit in-root code path now selects its declared `.pddrc` context only; the finalizer still requires the resulting canonical role to be that exact path.

Focused validation passed: 339 affected tests; 33 crash/skip pairing tests (184 deselected); the exact update regression; PDD JSON/source-hash validation (12 metadata + 3 run artifacts, with 2 obsolete run records intentionally absent); diff checks; and literal-worktree imports.

No approval or merge action taken.

Serhan-Asad:
## REVIEWED_HEAD

- PR: #2195
- Exact reviewed head: `fdd688affe4062594eb4a3bdcbfa8fd6bcedc16a`
- Frozen diff base: `2cacc91f90759ff45f1ad976da3b773e1a5f07a5`
- Local HEAD, remote `refs/pull/2195/head`, and the PR API head OID matched exactly.
- Worktree was clean before and after review.
- Freshness evidence: live `main` is now `2c917fde4f7d622ad0860b16ae9dbb36ae9cfbcb`, one commit after the common base, while `integration/agentic-sync-v1` remains frozen at `2cacc91f90759ff45f1ad976da3b773e1a5f07a5`. The intervening main commit changes only `scripts/release_video.py` and `tests/test_release_video.py`; neither path overlaps any of this PR's 50 changed paths. Per coordinator policy, I did not substitute the diff base.

## GUIDE_PATH_AND_SHA

- Guide: `/Users/serhanasad/Desktop/SF/pdd/docs/prompting_guide.md`
- SHA-256: `6c8b69c8801499d2993f1c8149f8991bbe361adbc11dedaf68c493a36b5b2ddc`
- Applied guidance: prompts remain the source of truth; `<pdd-dependency>` declarations must match architecture dependencies; behavioral and negative tests are part of the accumulated generation mold, not disposable external evidence.

## REASONS_NOT_TO_MERGE

1. **[P1] A caught nested same-unit exception can publish half-state.** `AtomicStateUpdate.__enter__` joins an active same-unit transaction, but the joined `__exit__` returns without marking the owner aborted. A direct exact-head reproduction removed the staged run report inside a nested transaction, raised and caught `RuntimeError`, then let the owner exit normally. Result: `report_exists=False` while the old fingerprint remained. That violates the required rule that any non-null exception at nested depth aborts outer publication and recreates the missing-report/fingerprint split this change is intended to prevent. See `pdd/fingerprint_transaction.py:218` and `pdd/fingerprint_transaction.py:244`.

2. **[P1] The advertised Windows fallback cannot reach the lock.** `_lock_path_for` unconditionally calls `os.getuid()` at `pdd/fingerprint_transaction.py:158` and `:164`; `os.getuid` is unavailable on Windows. Therefore every authoritative fingerprint/run-report read or mutation fails before the `msvcrt` fallback can operate. This conflicts with the explicit supported-Windows requirement in `pdd/prompts/fingerprint_transaction_python.prompt:55` and the project's Windows support statement in `docs/ONBOARDING.md:9`.

3. **[P2] Explicit-output update operations bypass the required unit lock.** `update_main` returns directly to `_update_main_locked` whenever `output` is truthy (`pdd/update_main.py:2046`); only the no-output path enters `AtomicStateUpdate`. Two concurrent updates targeting the same canonical source via `--output` can therefore read and mutate without the per-unit lock, race until late finalization, and both report success although only the last bytes/fingerprint survive. The update prompt requires locking before any read or mutation (`pdd/prompts/update_main_python.prompt:48`).

4. **[P1] The explicit post-command freshness acceptance test is absent.** Issue #1926 requires a CI property harness over every mutating command. The changed tests provide focused finalizer and command tests, but no parametrized harness covering the fixed command set (generate, example, update, fix, auto-deps, metadata/CI heal, sync, and pin) and asserting the post-command property for each successful return.

5. **[P1] The PDD evidence set is incomplete or stale.** The tracked CI-heal run report disagrees with its current metadata test hash/file set; the explicitly unignored sync-determine run report is deleted; the explicitly unignored update run report remains absent despite production changes; and the finalizer's PDD metadata maps only the 8-line importability test, excluding the 870-line behavioral/negative transactional suite from regeneration context. Several production modules also import the new finalizer without declaring the corresponding prompt/architecture dependency, including auto-deps, operation-log, sync, orchestration, and update paths. This leaves prompts, architecture, and verification artifacts unable to reproduce the reviewed implementation reliably.

## ACCEPTANCE_COVERAGE

- **Single transactional finalizer for all listed writers: PARTIAL.** The shared boundary is broadly routed and focused propagation tests pass, but the update explicit-output path bypasses the required early lock and dependency declarations do not fully describe the implementation.
- **Nested transaction abort semantics and atomic report/fingerprint publication: FAIL.** Reproduced a caught nested exception publishing report deletion while retaining the old fingerprint.
- **Portable lock/recovery and typed failure propagation: FAIL.** macOS-focused recovery/propagation tests pass, but the supported Windows path fails deterministically at `os.getuid()`.
- **Post-command freshness property test for every mutating command in CI: FAIL.** No complete command-set property harness exists.
- **Canonical unit/root handling, including absent report, wrong-root prevention, nested/custom roots, and issue #254 path preservation: PASS for exercised cases.** The focused suite and exact regression passed.
- **Crash/skip pairing and real report/fingerprint recovery: PASS for exercised local cases.** Thread/process/file recovery tests passed; this does not cure the caught nested-exception case.

## PDD_ARTIFACT_MATRIX

| Unit / artifact | Classification | Review result |
|---|---|---|
| `auto_deps_main` | REQUIRED_UPDATED | Prompt/code/example/test/meta hashes and absolute paths are current; finalizer dependency is missing from prompt/architecture. Run report is ignored/N/A. |
| `ci_drift_heal` | REQUIRED_UPDATED | Hashes and paths are current, but the tracked run report is stale versus metadata test hash/files. |
| `commands/maintenance` | REQUIRED_UPDATED | Code/meta/test/example evidence is current; direct finalizer use is not declared as a prompt/architecture dependency. Run report ignored/N/A. |
| `commands/modify` | REQUIRED_UPDATED | Hashes/paths and tracked 46-pass run are current; direct finalizer use is not declared as a prompt/architecture dependency. |
| `fingerprint_transaction` | REQUIRED_UPDATED | Prompt/code/example/meta hashes are current; PDD test mapping excludes the substantive 870-line transactional suite and records only an importability test. |
| `json_atomic` | REQUIRED_UPDATED / HUMAN_OWNED | No generator prompt by documented design; exercised by transactional tests. Directory-fsync portability remains a Windows risk in addition to the definite `getuid` failure. |
| `metadata_sync` | REQUIRED_UPDATED | Declared dependency, hashes/paths, and tracked 47-pass run are current. |
| `operation_log` | REQUIRED_UPDATED | Hashes/paths are current; lazy finalizer use is missing from prompt/architecture dependencies. Run report ignored/N/A. |
| `pin_example_hack` | REQUIRED_UPDATED | Declared dependency, hashes/paths, and tracked 2-pass run are current. |
| `sync_determine_operation` | REQUIRED_UPDATED | Hashes/paths are current, but its explicitly unignored tracked run report was deleted; lazy finalizer use also creates an undeclared dependency cycle with the finalizer. |
| `sync_main` | REQUIRED_UPDATED | Hashes/paths are current; finalizer dependency is missing from prompt/architecture. Run report ignored/N/A. |
| `sync_orchestration` | REQUIRED_UPDATED | Hashes/paths and focused crash/skip behavior are current; finalizer dependency is missing. Deleted run report is ignored/N/A by repository policy. |
| `update_main` | REQUIRED_UPDATED | Hashes/paths and the exact issue #254 regression are current; finalizer dependency is missing and the explicitly unignored run report is absent. |
| `docs/wave1_1926_artifact_matrix.md` | REQUIRED_UPDATED | Present but incomplete: it omits several affected generator-owned units listed above. |
| Hand-owned operator/docs changes | HUMAN_OWNED | Reviewed against the exact diff; they describe the intended protocol but do not disclose the Windows and nested-abort failures. |

All present changed JSON artifacts parsed successfully. Independent current-hash calculation for the 12 affected generator-owned units found no prompt/code/example/test/include hash mismatch; the blockers above concern behavioral correctness, dependency topology, test mapping, and tracked run-report freshness.

## HUMAN_REVIEW_STATUS

- GitHub reviews: 0
- Inline review comments: 0
- Existing issue-style PR comments before this review: 1, from Terra
- No human approval or human-authored review finding was present at review time.

## FAILURE_MODES

1. **Nested exception swallowed by the caller:** inner same-unit transaction stages report removal, raises, caller catches it, owner commits; observable result is a missing report paired with the old fingerprint.
2. **Supported Windows invocation:** the first authoritative fingerprint/report operation calls missing `os.getuid()`; locking and recovery never start, so mutating commands fail before finalization.
3. **Concurrent explicit-output updates:** two `pdd update --output ...` operations for one canonical unit bypass the early unit lock, race their reads/writes, and can both return success with only the last writer represented.

## PERFORMANCE

The new state layer has bounded payload/journal sizes and chunked copying, and I found no new unbounded CPU loop or network/subprocess work in the finalizer. It does add multiple staged writes, backup copies, journal rewrites, file fsyncs, and directory fsyncs per publication. More importantly, the same per-unit lock is held across long generation/update/sync workflows, so even read-only status calls for that unit can block behind LLM or command execution for minutes. Cross-unit operations remain independent. This is material expected latency/lock contention, but the verdict is driven by the correctness and evidence failures above.

## VALIDATION

- Mandatory sandbox gate: exact probe write failed with `Operation not permitted`; the probe remained absent.
- Mandatory GitHub gate: live `gh api` access succeeded without printing or inspecting `GH_TOKEN`.
- Exact-head integrity: local HEAD = remote PR head = `fdd688affe4062594eb4a3bdcbfa8fd6bcedc16a`; clean worktree; `git diff --check` clean.
- Changed JSON parse check: passed.
- Literal imports: all affected production modules resolved from this exact worktree.
- Generator hashes/paths: 12 affected units independently recalculated with no hash mismatch.
- Affected pytest selection: **647 passed, 1 unavailable under the mandated read-only sandbox, 107 warnings**. The sole failure was the real npm/Jest TypeScript E2E reaching its final write to `.pdd/meta/calculator_typescript_run.json`, which the sandbox correctly denied; it was not an assertion failure, so I make no claim that the full suite is green.
- Exact issue #254 regression: **1 passed**.
- Direct exact-head nested-abort reproduction: **failed the required invariant**, yielding `report_exists=False` with the old fingerprint still present.
- Mocked/stubbed boundary: many command/LLM tests patch external execution. Real local coverage included filesystem transactions, concurrency, subprocess crash recovery, and hash/path calculations. Hosted-provider and full writable-repository E2E were not run.

## REMAINING_RISKS

- `json_atomic` also opens/fsyncs directories directly, which is not demonstrated portable on Windows even after the definite `getuid` problem is addressed.
- Symlinked ancestors are rejected; on macOS, units beneath `/tmp` fail because `/tmp` resolves through a symlink to `/private/tmp`.
- No overall CI status is inferred from the focused local selection, and the real TypeScript E2E could not complete its publication step in the required read-only environment.
- The coordinator still owns any later readiness/base decision; the non-overlapping live-main advancement was assessed only as freshness evidence.

## VERDICT

CHANGES_REQUIRED

PR #2205, opened 2026-07-18
Enforce prompt repair deadline for blocking providers

### LINKED ISSUE #2045: Enforce --max-prompt-repair-seconds across blocking LLM calls
## Summary

`pdd checkup --prompt-repair` does not enforce `--max-prompt-repair-seconds` as a hard wall-clock limit when a synchronous LLM call blocks.

This surfaced during the staging3 E2E for promptdriven/pdd_cloud#3225 / PR #3232. The deployed command used:

```
pdd checkup <prompt> --prompt-repair best-effort \
  --max-prompt-repair-rounds 1 \
  --max-prompt-token-growth 1000 \
  --max-prompt-repair-seconds 120
```

It ran until pdd_cloud's outer 900-second subprocess timeout killed it.

## Production evidence

- pdd_cloud head: `5f54abc77249ed08516d3259e3db314a60d2e0c4`
- pinned PDD SHA: `eb2e739645b1233e6538d5dfac8e5b31c3caa796`
- staging job: `IdOqHkILm4NJkK16fvqV`
- Cloud Run execution: `pdd-executor-job-pprjn`
- stage: `source_set`
- outcome/reason: `BOUNDED_EXIT / global_deadline_exhausted`
- configured inner bound: 120 seconds
- effective outer bound and observed duration: about 900 seconds
- no accepted edit or commit; no recorded cost

pdd_cloud safely terminated the process group, released the 5,000-credit reservation, restored the account to 15,000 credits, removed the label, and left the PR head unchanged.

## Root cause

At the pinned PDD SHA, `run_prompt_repair_loop()` checks elapsed monotonic time before calling synchronous `change()`. That calls `llm_invoke()`. The budget is checked again only after `change()` returns.

If the provider operation blocks, control cannot return in time to enforce `max_seconds`. The flag currently limits when another round may start, but does not bound an active repair call.

## Expected behavior

`--max-prompt-repair-seconds N` should bound the complete repair operation, including every active provider request. At expiry, PDD should:

1. cancel or terminate the active operation;
2. return promptly with a stable timeout classification;
3. reject partial edits as successful output;
4. preserve available usage/cost evidence;
5. never start another round;
6. remain non-interactive in CI/hosted execution.

A small documented cleanup grace is reasonable. A 120-second bound must not permit an active request to continue until an unrelated 900-second supervisor timeout.

## Provider/auth interaction

This path uses `change() -> llm_invoke()` provider routing. It is separate from pdd_cloud's managed `codex exec` OAuth lane. Missing compatible provider credentials must fail promptly and must not start a browser/device flow in non-interactive execution.

The evidence proves the hard-timeout gap. It does not conclusively identify which provider/network operation blocked.

## Implementation direction

Any design is acceptable if it establishes a true hard bound, for example:

- async provider calls wrapped in `asyncio.wait_for`;
- request timeouts propagated through `change()` and `llm_invoke()`;
- a supervised child process whose process group is terminated at the deadline;
- provider timeouts plus a process-level fallback.

Use monotonic time for enforcement. Timing out a future while its underlying thread continues running or can later write output is insufficient.

## Acceptance criteria

- [ ] A blocking fake provider call is interrupted near the configured limit.
- [ ] The command does not wait for that call's natural completion.
- [ ] Timeout has a stable machine-distinguishable exit/result.
- [ ] No later repair round starts.
- [ ] Partial edits from a timed-out call are not reported as successful.
- [ ] Available usage/cost evidence is retained; unavailable cost is not reported as measured zero.
- [ ] Missing/invalid credentials fail promptly in non-interactive mode without browser/device authentication.
- [ ] Successful best-effort and interactive behavior remain compatible.
- [ ] Tests cover blocking before edits, blocking after a partial edit, provider exception, cancellation, and completion just below the deadline.
- [ ] Installed CLI/subprocess tests prove the boundary, rather than only directly testing the loop.
- [ ] Documentation states that the option is a hard end-to-end deadline.

## Relationship to pdd_cloud

pdd_cloud#3225 requires bounded low-risk source repair under one global deadline. PR #3232 will retain its own source-stage supervisor timeout as defense in depth. Once a fixed PDD SHA is pinned, staging should prove the CLI timeout fires before the cloud fallback.

PDD #2027 concerns structured `detect --stories` results and does not cover repair timeout enforcement.

dexhunter:
Fixes #2045

## Summary

- enforce one monotonic deadline across the complete prompt-repair loop
- supervise each legacy `change() -> llm_invoke()` call in a killable worker process group
- terminate provider descendants, reap the worker, and restore the original prompt on timeout
- preserve cost and model evidence from completed calls, including a primary call completed before same-round extraction times out
- keep ordinary `change()` callers and the authenticated CLI provider path unchanged
- synchronize the source prompts, user documentation, and regression tests

## Why

The previous loop checked elapsed time only between synchronous operations. A provider call that blocked inside `change()` could therefore outlive `--max-prompt-repair-seconds` until an unrelated outer supervisor killed the command.

The new task-local provider override leaves `change()` orchestration in the parent while moving only the active provider call behind a process boundary. This preserves extraction and accounting behavior while giving the repair loop a process tree it can terminate at the absolute deadline.

## Evaluation

Metric: `prompt_repair_deadline_overshoot_ratio` (lower is better).

- current-main baseline: `0.666493` to `0.706871`
- candidate: `0.030195` to `0.031754`
- conservative improvement: at least `95.2%` lower deadline overshoot

The strict evaluator covers blocking before edits, partial writes, descendant cleanup, a second-round timeout, a same-round extraction timeout, fast completion, near-deadline completion, provider exceptions, non-interactive credential failure, and the CLI path. Candidate metrics are suppressed unless all semantic gates pass.

The process-isolation direction was explored with Weco autoresearch, then hardened with the additional rollback, accounting, compatibility, and process-cleanup gates used here: [public trajectory](https://dashboard.weco.ai/share/Nzb2U1wuxlBSyilVLjtu2J6Eyar21S9-).

## Validation

- `uvx ruff check --fix pdd/change.py pdd/prompt_repair.py tests/test_prompt_repair.py`
- `pytest -q tests/test_prompt_repair.py tests/test_change.py tests/test_change_call_site_and_retry.py tests/test_change_main.py --timeout=60` (`204 passed, 2 skipped`)
- strict evaluator passed twice on the final candidate (`0.030195`, `0.031754`)
- both new deadline regressions fail on the unmodified baseline after waiting about five seconds and pass with this patch


dexhunter:
Re-checked this against current `main` (`8691065c`):

- The branch still merges cleanly. The four commits that landed on `main` since its base (#2319, #2320, #2318, #2324) touch none of the six files this PR changes, so there is no interaction to re-validate.
- `Unit Tests` has never actually run on this PR. It is sitting in `action_required` on all four commits, which is GitHub holding workflow runs on a fork PR pending approval. There is also an `auto-heal` run showing `startup_failure`, which looks repo-side rather than anything to do with this branch.

I deliberately did not try to reproduce the CI job locally — it provisions a protected sandbox and an identity-bound Node/Vitest toolchain that I can't stand up faithfully — so I'd rather not claim a green result I can't actually back. The focused pytest results in the description were run when this was filed, and none of the upstream commits above touch that code.

Glad to refresh the branch onto current `main` or narrow the diff if either would make it easier to look at.

PR #2263, opened 2026-07-21
Default Claude Code to Opus 5 and support explicit Fable 5

DianaTao:
## Summary

- Adds the PDD CLI side of the Opus 5 / Fable 5 rollout tracked in [promptdriven/pdd_cloud#3965](https://github.com/promptdriven/pdd_cloud/issues/3965), paired with [pdd_cloud#3852](https://github.com/promptdriven/pdd_cloud/pull/3852).
- Preserves the ordinary PDD model default from main: `DEFAULT_BASE_MODEL` remains `None`, so normal `llm_invoke` selection stays catalog-based.
- Defaults only the Anthropic / Claude Code agentic path to `claude-opus-5`.
- Supports `claude-opus-5` and `claude-fable-5` as distinct Anthropic models across direct LiteLLM invocation, Claude Code print/interactive routing, audit attribution, catalog testing, and fallback cost accounting.
- Removes every Opus 5 → Fable 5 compatibility alias. An optional `anthropic/` prefix is stripped only where Claude Code requires a bare model ID; the selected model identity is otherwise preserved.
- Adds separate bundled direct-Anthropic rows:
  - `claude-opus-5`: $5/M input, $25/M output, adaptive reasoning, 1M context.
  - `claude-fable-5`: $10/M input, $50/M output, adaptive reasoning, 1M context.
- Keeps both unranked (`platform-default`) and tries an explicitly requested Claude 5 model first rather than allowing strength interpolation to replace it.
- For authenticated cloud generation, forwards the invocation-scoped model in the `generateCode` payload while omitting the field when no model is selected.
- Exposes a request-local model scope for the paired cloud endpoint so nested generation, completeness, and continuation calls share the selection without process-global environment mutation.

Anthropic documents the two current API IDs and their distinct pricing in the [model overview](https://platform.claude.com/docs/en/about-claude/models/overview).

## Prompt associations

- Synchronizes `agentic_common`, `routing_policy`, `core/cli`, `llm_invoke`, `generate_model_catalog`, `llm_model_csv`, and `model_tester` prompt contracts with independent Opus 5 / Fable 5 routing.
- Refreshes the dependency-aware PDD fingerprints for every changed managed prompt/code/test association.

## Scope

[...]

PR #2268, opened 2026-07-21
chore(sync): preauthorize Fable contract transitions

DianaTao:
Phase A authorization for the exact Fable contract updates in #2263.

This branch changes only `.pdd/verification-profile-rotations.json`, replacing the two consumed Fable prompt transitions with exact final-byte authorizations:
- `generate_model_catalog_python.prompt`: `a086fd…` → `3cd583…`
- `llm_invoke_python.prompt`: `15c51e…` → `387ffe…`

Merge this prerequisite before the Phase B prompt/profile/code changes in #2263.

Validation: JSON parse and `git diff --check`.


DianaTao:
Sol finding resolved at `caef6f8aac932d83f6b92c480a05c9a24ad63867`.

- Accepted: the authority rows were correct, but the audit still modeled the older replay table.
- Root cause: it compared against replay-era rows and used future `HEAD` for historical replay.
- Fix: audit now checks 52 exact raw ordered survivors plus the two appended Fable successors, including exact source/target/profile hashes; replay is pinned to immutable `eb1fc0e`. Existing focused negative tests cover changed source/profile and malformed prompt/profile binding rejection.
- Validation: `ruff check tests/test_sync_core_pdd_rollout_policy.py`; focused authority audit (3 passed); focused consumed-replacement/source-state suite (7 passed); `git diff --check`. Full policy-suite run also exposed pre-existing historical-ref failures where old commits lack the current policy paths; no authority bytes were changed to mask those failures.

Please re-review this exact head.

DianaTao:
Accepted Sol review finding: the audit had weakened historical replay coverage by moving the story proof and replay base to the Phase-A snapshot.

Root cause: the test refactor used the current authority snapshot rather than preserving the approved immutable replay pair.

Fix: c728b05130c435256ef069db0eaaab1a0029d566 restores the story assertion against `e10bd9b3` (including the agentic-common row/count) and pins replay to `e10bd9b3 -> eb1fc0e2`; authority bytes are unchanged.

Validation: fetched and verified both immutable commits; focused audit/replay/story tests: 3 passed; `ruff check` and `git diff --check` passed. Complete policy collection has 93 cases; this shallow local worktree ran 89 successfully. The remaining 3 branch-only local-fetch fixtures cannot run here because Git rejects updating a ref from a shallow source (`shallow roots are not allowed to be updated`); the clone-based fixture passed.

Requesting Sol exact-head re-review of c728b05130c435256ef069db0eaaab1a0029d566.

DianaTao:
## Validation plan for this Phase-A authorization

This PR changes only the protected authorization history and its policy test; it does not change the Fable runtime, catalog, prompt, metadata, or verification-profile bytes. Validation is therefore split into authority checks here and behavioral/provider checks in Phase B (#2263).

1. **Exact history check:** prove that the 52 surviving schema-2 rows are byte-identical and retain their relative order, exactly two consumed rows are removed, and the two Fable authorization rows are appended.
2. **Exact binding check:** recompute and assert the current source prompt/profile hashes, the two Phase-B target prompt hashes, and the target profile hash. Mutating any source, target, or profile binding must fail.
3. **Real-loader replay:** fetch the immutable historical objects and load the actual transition `e10bd9b3… -> eb1fc0e…`; require 469 expected/loaded profiles, zero invalid reasons, zero unaccounted paths, and coverage `1.0`. This is intentionally a real transition, not a same-commit no-op.
4. **Current-base dormancy:** prove the two successors remain dormant against the current protected base; Phase A must not change effective runtime/profile behavior.
5. **Regression suite:** run `python -m pytest tests/test_sync_core_pdd_rollout_policy.py -q`, focused verifier/consumed-replacement tests, `ruff check`, and `git diff --check`.
6. **After Phase A merges:** rebase #2263 onto that protected-main SHA, run its catalog/runtime/provider E2E plan, then repin and validate pdd_cloud before it leaves Draft.

I will post the exact commands, commit, and pass/fail results for this plan before changing this PR out of Draft.

DianaTao:
## Why this is Phase A and Phase B (plain English)

Think of the protected verification profile as a signed travel authorization.

- **Phase A (#2268)** says, “we have reviewed the *exact* future Fable prompt/profile bytes, and we authorize that specific move.” It deliberately leaves the running code and live profile untouched. That gives reviewers a chance to reject the authorization without any behavior changing.
- **Phase B (#2263)** makes the actual code, catalog, prompt, metadata, and profile change. The verifier will accept it only if it matches the exact authorization that already landed in protected `main`.

Putting both in one PR would let the change authorize itself. Splitting them prevents that: Phase A is the locked approval; Phase B is the thing that consumes that approval. The cloud PR stays Draft for the same reason at the next layer: it must pin the final protected-main PDD commit, not a temporary feature-branch commit.

DianaTao:
## Validation-plan execution — exact head `caef6f8aac932d83f6b92c480a05c9a24ad63867`

Executed locally in the exact Phase-A worktree:

- `git diff --check` — passed.
- `python -m ruff check tests/test_sync_core_pdd_rollout_policy.py` — passed.
- Changed authority/replay/dormancy tests — **3 passed**. This includes the real immutable replay path, not a same-commit replay.
- Generic consumed-replacement/source-state verifier suite — **7 passed**.
- Exact policy suite: `python -m pytest tests/test_sync_core_pdd_rollout_policy.py -q` — **89 passed, 4 failed, 1 warning** in 160.75s.

The four full-suite failures were investigated individually. One clone case passed when rerun in isolation. The other three reproduce only because this local checkout is shallow: their fixture tries to create `protected-preauth`, and Git rejects updating that ref from shallow roots. This is a test-environment/history-availability failure, not an authority/runtime failure; it does not occur in CI's full-history checkout. No policy assertion or authority byte was weakened to hide it.

Remaining gate: keep this PR Draft until Sol's exact-head review clears the corrected historical replay, then mark it ready and require the normal full CI lanes. Phase-B/provider E2E remains intentionally pending until this authority merges.

DianaTao:
Correction to the prior validation-execution comment: the exact pushed and validated head is `c728b05130c435256ef069db0eaaab1a0029d566` (it includes the restored immutable `e10bd9b3 -> eb1fc0e2` replay and historical story assertion). The earlier `caef6f8a` identifier was the immediately preceding test-audit commit; do not treat it as the current review target. All stated focused/full-suite evidence applies to the final `c728b051` head unless explicitly labeled otherwise.

DianaTao:
Accepted final Sol finding: the restored historical c566 assertion retained a stale comment claiming the replay base was current main.

Root cause: comment drift after restoring immutable e10 replay coverage.

Fix: 6777db4dcd27bc190f41f297dbaa73b007a91ffc changes only that comment to state the immutable e10 historical basis; no authority/runtime bytes changed.

Validation: ruff, diff check, and focused story/authority/replay suite: 3 passed. Requesting final exact-head Sol review of 6777db4dcd27bc190f41f297dbaa73b007a91ffc.

PR #2273, opened 2026-07-22
test(update): stabilize include dependency fixture

gltanaka:
## Summary

- create the include dependency before rooted SHA256 cache priming
- preserve the changed=True behavior, reason assertion, and exact hash-call assertion
- avoid production hash/cache changes

## Release blocker

Cloud release run `run-20260721-232839-4597a4d1a` failed `tests/test_update_main.py::TestIsCodeChanged::test_fingerprint_include_deps_changed` because coalesced parent mtimes can reuse the directory listing cached before `preamble.md` exists.

## Validation

- exact failing test: 10/10 passed in fresh pytest processes
- `TestIsCodeChanged`: 6 passed
- `tests/test_update_main.py`: 136 passed
- Ruff critical selectors (`E9,F63,F7,F82`): passed
- `git diff --check`: passed

## Scope

Test-only fixture ordering change. No production behavior, assertions, skips, or unrelated formatting changed.

### DIFF
--- a/tests/test_update_main.py
+++ b/tests/test_update_main.py
@@ -2235,10 +2235,9 @@ class TestIsCodeChanged:
 
         code_file = tmp_path / "module.py"
         code_file.write_text("def foo(): pass\n")
-        current_hash = calculate_sha256(code_file, tmp_path)
-
         dep_file = tmp_path / "preamble.md"
         dep_file.write_text("updated preamble content\n")
+        current_hash = calculate_sha256(code_file, tmp_path)
 
         mock_fp = MagicMock()
         mock_fp.code_hash = current_hash


gltanaka:
## Investigation summary

Sibling review found two rooted-hash setup cases in `TestIsCodeChanged`: the matching-hash case creates its only file before hashing, while this include-dependency case alone created another trusted-root file after the initial hash. No production sibling requires a change.

Prior history:
- `aeb2c3bc1` introduced include dependency change detection.
- `ec59ce9ad` established that unreadable dependencies intentionally return `changed=True` with an unreadable reason.
- `8f288cc4d` introduced the test fixture under the earlier unrooted hash contract.
- `e10bd9b3d` / PR #1971 rooted both hashes at the project and added the exact two-call assertion while introducing bounded directory indexing.

Convention followed: construct the complete trusted-root fixture before priming rooted path/hash lookup, then retain exact behavior and call assertions. Scope remains the single test fixture; no cache/hash implementation changes.

gltanaka:
## Cloud red to local green evidence

- Red: release cloud run `run-20260721-232839-4597a4d1a`, task 15, failed `tests/test_update_main.py::TestIsCodeChanged::test_fingerprint_include_deps_changed` when the cached pre-creation directory listing yielded `include dependency unreadable`.
- Green: the exact test passed 10/10 in fresh pdd-environment pytest processes after fixture reordering.
- Proportionate coverage: `TestIsCodeChanged` 6/6 passed; `tests/test_update_main.py` 136/136 passed.
- Static/patch checks: Ruff `E9,F63,F7,F82` passed; `git diff --check` passed.

Commit: `496ea34c9`.

gltanaka:
## Codex adversarial review — PR #2273 @ 496ea34c99d35a374082800f57db30c19345a099

**Verdict:** MERGE

### P1 (blocking)
- None.

### P2 (non-blocking)
- None.

### Notes
- The fixture now creates both trusted-root files before the first rooted `calculate_sha256` call primes the mtime-keyed directory index. This removes the cloud coalesced-mtime dependency without changing production code.
- Behavior coverage is preserved: the test still requires `changed is True`, the `include dependency changed` reason, and the exact ordered hash calls for `(code_file, tmp_path)` and `(dep_file, tmp_path)`.
- Adversarial validation: a forced coalesced-directory-mtime simulation reproduced the old-order miss and confirmed the reordered fixture hashes the dependency; the exact test passed 10/10 in fresh processes; all 6 `TestIsCodeChanged` tests passed; `git diff --check` passed.
- Review preconditions held after a fresh fetch: clean worktree, no unpushed commits, and local/remote PR HEAD both equal the reviewed SHA.

PR #2281, opened 2026-07-22
test(change): accept word ordinal retry exhaustion

gltanaka:
## Summary
- recognize spelled attempt ordinals from first through twentieth in retry-exhaustion judgments
- share the bounded ordinal grammar across condition-first and action-first forms
- cover the exact cloud output plus negation, contradiction, unrelated-condition, and clause-binding adversarial cases

## Release evidence
- Base: `bff563ebf76732592270ce25b6689019528b5bf7`
- Cloud run: `run-20260722-072651-bff563ebf`
- Exact compliant output: `The runner should attempt the pipeline a maximum of 3 times. If the third attempt also fails with a connection error, the runner must propagate the connection error to the caller.`

## Validation
- RED: exact regression failed with `Missing retry exhaustion condition and fallback action.`
- `pytest -q tests/test_change_call_site_and_retry.py -k "word_ordinal or common_word_ordinals"` (12 passed)
- `pytest -q tests/test_change_call_site_and_retry.py::TestDeterministicChangeJudges` (151 passed)
- `pytest -q tests/test_change_call_site_and_retry.py` (151 passed, 2 skipped)
- `ruff check --select E9,F63,F7,F82 tests/test_change_call_site_and_retry.py`
- bounded performance corpus: 12,000 assertions + 200 long near-matches in 0.229s (<5s bound)
- `git diff --check`

Issue #1932, opened 2026-07-08
Master: Continuous artifact sync — prompts, code, docs, and tests stay in sync automatically (supersedes #894)


gltanaka:
## Goal

**Everything stays in sync automatically — no matter what changes (doc, prompt, code, test) or who changes it (human, coding agent, pdd itself, a manual hotfix PR).** Manual `pdd update --all` / `pdd sync` become recovery tools, not routine chores. In the steady state, drift is captured the moment it happens, healed at the next natural boundary (commit, PR, merge, nightly), and never silently accumulates.

This supersedes #894, whose Definition of Done has shipped (see the [audit](https://github.com/promptdriven/pdd/issues/894#issuecomment-4912477090)): `pdd update --all`, include-doc handling, CI drift detection + auto-heal (required check), `pdd change` preflight healing, and no-arg global `pdd sync` all exist. **Yet drift persists** — pdd_cloud measured 377/788 fingerprints stale (promptdriven/pdd_cloud#2994) and had to build its own nightly reconciler that deliberately avoids calling `pdd sync`. The tools exist; the guarantee doesn't. This epic is about the guarantee.

## Why drift persists: six systemic causes

From a sweep of ~180 sync/drift issues in this tracker plus consumer-side evidence in pdd_cloud:

1. **Fingerprint finalization is a scattered side effect, not a transaction.** ~24 issues in 4 near-identical waves (#911–#913, #953–#956, #986 + #988–#990, #1005–#1007; #437, #430, #1211→#1252→#1290, #1305). Every mutating command re-implements "save fingerprint on success"; every new path forgets or botches it; heal loops re-trigger on their own null output (#1305). → #1926
2. **The dependency graph is modeled twice and both models are incomplete.** Prompt `<include>` regex vs `sync_order.py` parser vs architecture.json `dependencies` disagree (#733, #1061, #881 open); PRD, docs, and out-of-tree tests were never edges at all (#882 — PRD sync appears broken from day one, a result tuple checked as if it were LLM text; corroborated by #1625, #1903). Each missing edge was added reactively after a stale-artifact incident. → #881, #1931
[...]

PR #2282, opened 2026-07-22
test(sync): bound continuous sync CLI subprocesses

gltanaka:
## Summary
- bound every issue #1932 CLI subprocess to 30 seconds and isolate it in a process group
- kill and reap the full process group on command timeout or an external pytest-timeout interruption
- separate the unconditional `reconcile --heal` refusal/metadata invariant from the five-consumer classification loop
- give only the multi-process classification acceptance case a measured 90-second aggregate budget

## Hosted failure triage
PR #2281 run 29933294255 attempt 2 reached the final (sixth) child in the doc and prompt cases, then the suite-wide 60-second timeout interrupted `subprocess.communicate()`. The code and test cases passed. Local cases took ~14.5s each normally and ~19s each while four copies competed concurrently, supporting cumulative startup/contention rather than a deterministic `--heal` deadlock.

The split preserves all four drift classifications across reconcile/sync/update/change/CI, and separately preserves all four `--heal is disabled` and fingerprint-immutability assertions. The child timeout is lower than both aggregate limits and now has deterministic cleanup diagnostics.

## Validation
- focused affected tests: 9 passed
- focused xdist (`-n 2`) repeated 3 times: 8 passed each run
- corrected focused xdist including cleanup regressions: 10 passed
- full file: 42 passed (before final cleanup-regression addition; final focused run covers that addition)
- Ruff critical selectors (`E9,F63,F7,F82`): passed
- `git diff --check`: passed

## Scope
Test infrastructure only; no product behavior changed.

Related to #1932 and release blocker observed on #2281.

### DIFF
--- a/tests/e2e/test_issue_1932_continuous_sync_guarantee.py
+++ b/tests/e2e/test_issue_1932_continuous_sync_guarantee.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 
 import json
 import os
+import signal
 import subprocess
 import sys
 from pathlib import Path
 from typing import Callable, Dict, List
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 import pytest
 
@@ -17,6 +18,7 @@ from pdd.sync_determine_operation import get_pdd_file_paths
 
 
 REPO_ROOT = Path(__file__).resolve().parents[2]
+COMMAND_TIMEOUT_SECONDS = 30.0
 
 
 def _env() -> Dict[str, str]:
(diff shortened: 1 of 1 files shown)


gltanaka:
MERGE — critical P1/P2 review complete.

Exact base: bff563ebf76732592270ce25b6689019528b5bf7
Exact head: b833a4314be69c9a752a1343478bff9c479b7a0a

Findings: 0 P1, 0 P2. The Popen wrapper starts an isolated POSIX session, bounds every child at 30s, kills the full process group and reaps the direct child on both subprocess TimeoutExpired and external pytest-timeout-style BaseException, preserves timeout stdout/stderr diagnostics, and re-raises external interrupts. Normal completion is reaped by communicate(). Linux and macOS share this POSIX path; Windows falls back to direct-child kill and is outside the required CI platforms. The split retains all four classification/ledger assertions and all four disabled-heal/fingerprint-immutability assertions; only the four multi-command classification cases receive the scoped 90s marker. Diff is one test file with no product changes.

Reviewer validation:
- focused serial: 10 passed, 33 deselected (79.84s)
- focused xdist -n 2: 10 passed (45.07s)
- real macOS descendant probe, subprocess timeout: passed; full process group terminated, no live grandchild
- real macOS descendant probe, signal-driven external interrupt: passed; original interrupt preserved, no live grandchild
- git diff --check: passed
- exact scope: tests/e2e/test_issue_1932_continuous_sync_guarantee.py only

Residual uncertainty: the long hosted Linux unit job is still running; all completed hosted checks are green. This review found no merge-blocking issue.

PR #2316, opened 2026-07-25
Fix ChatGPT subscription generation routing

gltanaka:
## Summary

- route `chatgpt/*` model smoke tests and normal generation through LiteLLM's Responses API
- bridge Codex login credentials and apply the existing ChatGPT SSE output patch before setup/model tests
- send ChatGPT Responses requests as list-form input built from the final messages, preserving schema-in-prompt instructions
- prevent failed ChatGPT Responses requests from falling through to the unsupported chat-completions endpoint
- retain the existing direct OpenAI GPT-5 Responses behavior and hosted-budget admission checks

## Root cause

`pdd setup` and `pdd generate` exercised different LiteLLM paths. The setup/model test could validate the Codex subscription, while generation called `litellm.completion()` for `chatgpt/gpt-5.6-sol`. That targeted `/backend-api/codex/chat/completions`, which returned a browser-only Cloudflare challenge. PDD then fell back to Gemini despite the ChatGPT subscription being valid.

The Codex subscription backend works through `litellm.responses()` and requires list-shaped `input`.

## Tests

- `pytest -q tests/test_model_tester.py tests/test_codex_subscription.py` — 71 passed
- focused ChatGPT/Codex/Responses regression selection — 43 passed
- existing OpenAI Responses API tests — 4 passed
- `python -m py_compile pdd/llm_invoke.py pdd/model_tester.py`
- `git diff --check`

## Live verification

Ran an editable install against an authenticated Codex subscription:

```bash
PDD_MODEL_DEFAULT=chatgpt/gpt-5.6-sol \
PDD_FORCE=1 \
pdd --local --strength 1.0 generate success_python.prompt \
  --output /home/glt/pdd/codex_responses_smoke.py
```

The command completed without the Cloudflare response or Gemini fallback:

```text
✓ Step 1 (generate): Cost: $0.000000, Model: chatgpt/gpt-5.6-sol
```

### DIFF
--- a/.pdd/meta/llm_invoke_python.json
+++ b/.pdd/meta/llm_invoke_python.json
@@ -1,9 +1,9 @@
 {
-  "pdd_version": "0.0.309.dev0",
-  "timestamp": "2026-07-24T23:44:59+00:00",
-  "command": "test",
-  "prompt_hash": "34e03947c072722466560349d32bde6f285a797ad533cb8b049f59abc08a6ca7",
-  "code_hash": "c32efbee0533eddfa7e2f0776b033fe44c3b7117de0955846eb218eb16803ee3",
+  "pdd_version": "0.0.310.dev0",
+  "timestamp": "2026-07-28T16:11:20.872917+00:00",
+  "command": "fix",
+  "prompt_hash": "9e4ef520d92fda9138e300f399b918d217117daf35c906e552a94849e009928f",
+  "code_hash": "4865533bb177323ee5dde136f7c7e9081a56d7d5e03bc0daa1779021022b3be9",
   "example_hash": "3e66413118ec2f8e8940e7a835932920309aeeb775d23ce728d78299fb99e2ec",
   "test_hash": "c11062141398737eb949d11e117c14f7e82492d00c4ca4921ac227364a21d8c4",
   "test_files": {
--- a/.pdd/verification-profiles.json
+++ b/.pdd/verification-profiles.json
@@ -7955,7 +7955,7 @@
       "prompt_path": "pdd/prompts/llm_invoke_python.prompt",
       "language_id": "python",
       "required_requirement_ids": [
-        "CONTRACT-SHA256:09e5140c01bbf8136f4c487c873c816f8e75db75412c11794a8b7ea47259cf3c"
+        "CONTRACT-SHA256:10129606f47d4301052490b7767acc08d8fc713e48bcb2b867efadf2063f8d1e"
       ],
       "obligations": [
         {
@@ -7964,7 +7964,7 @@
           "validator_id": "threshold-ed25519",
           "validator_config_digest": "threshold-ed25519-v1",
           "requirement_ids": [
-            "CONTRACT-SHA256:09e5140c01bbf8136f4c487c873c816f8e75db75412c11794a8b7ea47259cf3c"
+            "CONTRACT-SHA256:10129606f47d4301052490b7767acc08d8fc713e48bcb2b867efadf2063f8d1e"
           ],
           "artifact_paths": [
             "pdd/prompts/llm_invoke_python.prompt"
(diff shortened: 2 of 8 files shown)


gltanaka (commented):
## Changes requested

GitHub does not permit the PR author to submit a formal `REQUEST_CHANGES` event, but these are blocking findings. The single-call Responses routing is needed and the controlled focused tests support it; the current head is not merge-ready.

**P1 — preserve the fix in PDD’s source of truth.** The new behavior in `pdd/llm_invoke.py:5685-6057` and `pdd/model_tester.py:191-272` is not reflected in either checked-in prompt copy. `pdd/prompts/llm_invoke_python.prompt:180-183` still says only OpenAI GPT-5 uses `responses()` and every other model uses completion/batch completion; `pdd/prompts/model_tester_python.prompt:17-21` still requires exactly one `litellm.completion()` call. `model_tester` also has no correct architecture registration/baseline for `pdd/model_tester.py`. The repository’s own check confirms this: `python -m pdd.ci_drift_heal --dry-run --diff-base origin/main...HEAD --modules llm_invoke model_tester --json` exits 1 with `llm_invoke` = `CONFLICT` and `model_tester` = `UNBASELINED`. A future sync can therefore regenerate the unsupported endpoint and undo this fix. Update the canonical/package prompt copies plus architecture/metadata required by the sync workflow, and make the focused drift check clean.

**P1 — the supported batch boundary still uses the broken endpoint.** `use_responses_api` is gated by `not use_batch_mode` at `pdd/llm_invoke.py:5692-5700`; a `chatgpt/*` batch consequently reaches `litellm.batch_completion()` at `pdd/llm_invoke.py:6234-6237`. A focused reproduction with `PDD_MODEL_DEFAULT=chatgpt/gpt-5.3-codex` recorded `responses_calls=0`, `batch_completion_calls=1`. That retains the same unsupported chat-completions/Cloudflare failure for a public `llm_invoke` mode and conflicts with the original first-class-provider contract that every caller can use the subscription. Route batch items through the Responses path, or fail closed with an actionable unsupported-mode error; add a regression proving no ChatGPT request reaches completion or batch completion.

**P2 — make the changed subscription tests hermetic.** In this dedicated worktree, `pytest -q tests/test_model_tester.py tests/test_codex_subscription.py` gives 69 passed / 2 failed because `test_fallback_reaches_chatgpt_when_anthropic_key_missing` and the rewritten structured-output test inherit the project’s legitimate `PDD_MODEL_DEFAULT` from `.env` and never reach their mocked ChatGPT row. Pin/delete `PDD_MODEL_DEFAULT` explicitly in each test as appropriate. With `PDD_MODEL_DEFAULT=claude-sonnet-4-6`, the same suite passes 71/71.

The contributor’s live `pdd generate` smoke is useful but only covers an unstructured single request. After the fixes, please also record a real structured subscription invocation and the setup/model-test boundary changed here; batch can remain mocked if a real batch would consume unnecessary allowance.

gltanaka:
Follow-up review on `eeae8fefb` — approved after two fix/review cycles.

Resolved the original routing/prompt drift and batch safety issues, then addressed adversarial findings: the ChatGPT setup smoke now rejects empty Responses output, and code-generator text/image inputs retain `input_text` + `input_image` structure in the Responses request.

Validation: focused model tester/subscription suite (75 passed); Responses/batch invocation selection (14 passed); syntax and diff checks; `ci_drift_heal` reports `llm_invoke` and `model_tester` IN_SYNC. Current PR-head CodeQL and auto-heal checks are green.

Residual risk: provider calls are mocked; no live ChatGPT subscription smoke was run. I cannot submit a formal GitHub approval because this account is the PR author.

gltanaka:
Final review/fix loop is complete. The PR head has been rebuilt on current `main` with the Sol-high-approved product patch (stable patch ID `704cb9725100988db76214d6e6c1efa30c149940`).

Validation performed on the equivalent final tree:
- 482 focused product tests passed (`test_codex_subscription`, `test_model_tester`, `test_llm_invoke`)
- 217 protected rollout/verifier tests passed
- exact pre/post rollout loader probes passed (manifest/profile mismatches: 0; coverage: 1.0)
- authenticated ChatGPT subscription smoke passed with `chatgpt/gpt-5.6-sol`, result `PDD_PR2316_SMOKE_OK`, finish `completed`
- adversarial Sol-high review approved after substantive correctness and lifecycle fixes

Prerequisite governance/test-state changes were isolated and merged through #2355, #2356, #2357, #2359, #2360, and #2361. Awaiting this head’s full hosted CI before merge.

PR #2324, opened 2026-07-25
fix(cloud-batch): use Gemini 3.6 Flash for release validation

gltanaka:
## Summary
- replace the stale `vertex_ai/gemini-3-flash-preview` Cloud Batch default with `vertex_ai/gemini-3.6-flash`
- update the regular and sync regression overrides so every Batch execution path uses the same current model
- add a regression contract that rejects stale preview pins

## Release impact
Unblocks the v0.0.309 final cloud gate. The active staging catalog rejects the old preview identifier; the current catalog contains `vertex_ai/gemini-3.6-flash`. No video work is included.

## TDD evidence
- RED: `conda run -n pdd python -m pytest tests/test_cloud_batch_job_templates.py::test_cloud_batch_execution_paths_pin_current_release_default_model -q` failed against the three old preview pins.
- GREEN: `conda run -n pdd python -m pytest tests/test_cloud_batch_job_templates.py -q` — 22 passed.
- `bash -n ci/cloud-batch/entrypoint.sh tests/regression.sh tests/sync_regression.sh` — passed.
- `git diff --check origin/main..HEAD` — passed.

## Scope
Only the Batch entrypoint, two regression harnesses, and their focused contract test change.

PR #2286, opened 2026-07-23
fix(test): recognize let-propagate fallback action

gltanaka:
## Context

Cloud Batch run `run-20260722-151736-2cf006b37` failed task 31 on explicit bounded fallback guidance:

> If the 3rd attempt still fails with a connection error, `run_pipeline` must let the connection exception propagate to the caller.

After #2283 recognized the ordinal exhaustion and cause tail, the deterministic fallback-action grammar still did not recognize this affirmative terminal action.

## Change

- Add a test-oracle-only action production for `let` + an explicit error/exception subject + terminal `propagate`.
- Accept either a bare terminal propagation or an explicit caller/user/client/upstream/calling-code destination.
- Preserve modal-subject handling, negation and continuation rejection, bound matching, and clause isolation.
- Explicitly leave nested Markdown action bullets unsupported; the exact Cloud form is a top-level bullet and is covered directly.
- Add the exact full Cloud output and focused positive, semantic-reversal, unrelated-condition, unbounded, and nested-isolation controls.
- No product, prompt, provider, runtime, or workflow changes.

## Validation at rebased head `989268742` on base `c5cb6dab5`

- Exact Cloud-output regression: 1 passed.
- Exact Cloud/top-level-bullet focused selection: 23 passed.
- Adjusted original reviewer matrix: 15/15 matched; the documented pre-existing unsupported inverse form is excluded.
- Nested scope/isolation matrix: 9/9 matched.
- Full module: 196 passed, 2 skipped.
- Ruff and `git diff --check`: passed.
- Rebase equivalence to approved head `9f4a79d92`: binary diff SHA-256 `679adf9cc0e16f433cd40efec2c4078692691696480ff41fafdd0c6ce3739e1f`; stable aggregate patch ID `8fbdade9b7017494383cc511925dc8d5002e54da`; changed-file blob `c6b39b9ce2d3333cad83504ace38fdfbf4b67f3a`.

TDD evidence: initial red commit before rebase had 4 expected failures and 8 passing controls. Correction red commit reproduced the four semantic false positives and two nested false negatives before nested support was explicitly removed from scope.

No paid/real LLM test was rerun locally. Fresh merged-main Cloud Batch remains the authoritative release gate.

## Risk

Natural-language grammars can false-negative novel equivalent wording. This production intentionally favors soundness: it requires literal `let`, an explicit error/exception subject, terminal `propagate`, and no restrictive or redirecting tail. Nested action bullets remain unsupported to avoid weakening conditional-ancestor isolation.

PR #2283, opened 2026-07-22
fix(test): parse ordinal retry cause tails

gltanaka:
## Context

The final `2026-07-21` release-candidate Cloud Batch run `run-20260722-114232-5237e0689` failed only the real-LLM retry safety assertion in task 31. The generated guidance declared a maximum of 3 attempts and explicitly required `run_pipeline` to propagate the final exception after the third attempt failed due to a connection error, but the deterministic judge returned `Missing fallback action`.

## Root cause

The bound ordinal exhaustion grammar consumed `If the 3rd attempt also fails` but not its `due to a connection error` cause tail. The remaining condition text prevented the clause-safe parser from binding the following explicit propagation action.

## Change

- Extend only condition-first bound-ordinal exhaustion to consume a bounded optional `with|due to ... error|exception|failure` cause tail.
- Add the exact full cloud-output paragraph and focused propagation/re-raise positives.
- Preserve negative controls for missing, vague, availability-only, negated, contradicted, mismatched-bound, and unrelated-condition fallback actions.
- Keep existing inverse `when` behavior unchanged; ordinal `once`/`after` expansion is intentionally out of scope.

Only `tests/test_change_call_site_and_retry.py` changes; there are no runtime, prompt, provider, or workflow changes.

## Validation

- Exact cloud-output reproduction: passed.
- Adjusted cause-tail matrix: 24/24.
- Deterministic judge suite: 173 passed.
- Full module: 173 passed, 2 skipped.
- Ruff critical selectors and `git diff --check`: passed.
- Critical review at exact head `4964c00f0e2a01beb048499a1b745c7537ce2636`: MERGE, 0 P1 / 0 P2.

No paid/real LLM test was rerun locally. Fresh merged-main Cloud Batch remains the authoritative release gate.

## Risk

Natural-language grammar can overmatch. The new tail is limited to the existing bound-ordinal condition-first production, `with` or `due to`, an optional article, at most four modifier words, and an error/exception/failure noun. Existing bound matching, clause adjacency, negation, action rejection, and contradiction checks remain intact.

PR #2278, opened 2026-07-22
fix(test): parse retry failure cause tails

gltanaka:
## Context

The 2026-07-21 release candidate `765b65cc713b197c20bb19d43c269de99349f61c` failed Cloud Batch task 31 even though the generated retry contract was compliant:

> If all 3 attempts fail due to connection errors, `run_pipeline` must propagate the final connection error to the caller.

Evidence: `gs://pdd-stg-ci-results/run-20260722-033259-765b65cc7/results/task_31.json` and `task_31.log` (699 passed, 1 failed). This is a narrow follow-up to #2193 and merged PR #2197 (`f22d100134bd4a7e99d88666b607d55493c41a7b`).

## Root cause

The deterministic exhaustion grammar recognized `all 3 attempts fail`, but left `due to connection errors` in the condition suffix. The clause-safe fallback parser correctly refused to interpret the later action through unrecognized text, yielding a false negative.

## Change

- Add the exact Cloud sentence as a positive regression.
- Add negative controls for continued retry, mere error availability, success, and negated propagation.
- Allow only a bounded optional `due to ... error(s)|exception(s)|failure(s)` tail after the existing all-N-attempts-fail form, then retain the existing adjacent action parser.

Only `tests/test_change_call_site_and_retry.py` changes. Runtime code, prompts, provider behavior, and cloud automation are unchanged.

## Validation

- New regression before fix: 1 failed, 4 passed.
- New regression after fix: 5 passed.
- `TestDeterministicChangeJudges`: 131 passed.
- Full module with real-test flags unset: 131 passed, 2 skipped.
- Ruff critical rules: passed.
- `git diff --check`: passed.

No paid/real LLM test was rerun locally. The authoritative end-to-end verification is the fresh full merged-main Cloud Batch release gate.

## Risk

Natural-language grammar can overmatch. The expansion is restricted to the existing numeric all-attempts-fail production, a `due to` marker, at most four modifier words, and an error/exception/failure noun. Existing action adjacency, rejection, negation, contradiction, and success checks remain intact.

PR #2328, opened 2026-07-26
fix(release): run the cloud gate continuously instead of during the release

gltanaka:
## Problem

The release was taking ~11.5h because `make cloud-test` — a ~2h Cloud Batch full regression, and the thing that blocks a release — ran for the **first time during the release**. Failures were therefore discovered mid-release, and each one restarted the run behind a full fix/review/merge cycle.

Reconstructed from the v0.0.309 orchestrator session:

| window | hours | what |
| --- | --- | --- |
| 06:21–12:12 | 5.9 | derive candidate, first gate run, merge four already-open blocker PRs |
| 12:12–14:25 | 2.2 | gate red (Gemini 3.6 Flash) → fix → PR #2324 → merge |
| 14:26–16:25 | 2.0 | forced re-validation, then gate red again (runtime digest cache) |
| 16:25–17:25 | 1.0 | fix → PR #2326 → merge |
| 17:26–17:53 | 0.5 | final gate run, tag, publish |

The middle **5.2h** is what this PR removes. Cadence over the two weeks this pattern held roughly halved, from ~1.0 to ~2.0 days per release.

## Change

**Continuous gate.** `cloud-test-main.yml` runs `make cloud-test` every 4h and on dispatch. Deliberately *not* per-push — the gate takes ~2h while main takes a commit every ~7 minutes, so per-push runs only pile up. It never cancels in flight either: `submit.sh` has no `gcloud batch jobs delete` on any path, so cancelling abandons 77 Batch tasks rather than stopping them.

**Release consumes, does not discover.** `check-release-cloud-green` refuses any candidate without a recent green run for that exact SHA, failing closed on gh auth failure, non-JSON, truncated JSON, empty history, future-dated runs (clock skew), and non-finite freshness bounds.

**Proven ancestors are releasable.** Requiring the tip deadlocks the gate — measured over 30 days, main had only 67 quiet ≥2h windows and 4 days with none at all. `check-release-branch` now requires containment in `origin/main` rather than equality, and reports how far behind the tip it is releasing. `release` still tags `HEAD`, so the commit that ships is always the commit that was proven.

**Prose-judge tests demoted.** `tests/test_change_call_site_and_retry.py` judges free-form LLM prose with regex oracles, so a rephrasing flips it with no product regression behind it — four release-blocking PRs (#2278, #2281, #2283, #2286) were spent widening those patterns mid-release. The two real-LLM cases now run in an advisory lane and report through the collected result JSON without gating. The 850 lines of deterministic judge tests stay blocking.

## Ships disarmed

`CLOUD_GREEN_GATE_ARMED := 0`. The workflow needs a GCP service account (`pdd-cloud-test-runner@prompt-driven-development-stg`) that **does not exist yet** — the existing WIF provider pins `attribute.workflow_ref`, so it cannot reuse auto-heal's. Arming before it exists would refuse every candidate and block all releases. `make release` warns loudly until armed.

To arm: provision the SA → confirm one green run → set `CLOUD_GREEN_GATE_ARMED := 1`.

## Review

Two independent adversarial reviewers ran against this diff. They found 4 P1s and 8 P2s, all fixed here:

- **Advisory logs broke the result-artifact allowlist** — `verify-result-identities.py` allowlists the results prefix exactly, so a `task_*`-prefixed advisory log aborted *every* run with a credential-boundary error even on a fully green fleet. Reproduced by the reviewer. Renamed outside that namespace, with a test that drives the real validator.
- **`cancel-in-progress` orphaned Batch jobs** rather than saving spend.
- **Exact-tip matching made the gate nearly unobtainable** → ancestor matching.
- **Clock skew failed open** — a future-dated run gave `age=-87600h` and exit 0.
- `inf`/`nan` freshness bounds; env-overridable via `?=`; a `RELEASE_SHA` flag that did nothing; advisory results invisible to the collector; image rebuild every run; 150-min timeout below the ~152-min worst case.

## Verification

352 tests pass across the affected suites, 464 across everything asserting on the changed files, 17,007 collect under `--strict-markers --strict-config`. Shell syntax + shellcheck clean, pylint 10.00, workflow YAML validated. Both fail-opens verified closed by execution; ancestor selection verified against real git history.

### DIFF
--- a/.github/workflows/cloud-test-main.yml
+++ b/.github/workflows/cloud-test-main.yml
@@ -1,0 +1,133 @@
+name: cloud-test-main
+
+# Runs the full Cloud Batch gate (`make cloud-test`) against main continuously,
+# so the release consumes a commit that is already proven green instead of
+# discovering failures on release day.
+#
+# Why this exists: `make cloud-test` is the gate that blocks a release, and
+# until now it ran for the first time during the release itself. That made
+# every latent failure a release blocker found at the worst moment. For
+# v0.0.309 the gate went red twice mid-release; each red restarted the run
+# behind a full fix/review/merge cycle. The two discoveries, their fixes, and
+# the re-validation they forced cost about 5.2 of that day's 11.5 hours.
+# Releases went from a ~1.0-day cadence to ~2.0 days over the two weeks this
+# pattern took hold.
+#
+# This workflow does NOT gate merges. It is deliberately post-merge: the gate
+# takes up to two hours, and blocking every PR on it would trade one throughput
+# problem for a worse one. It reports, and `make check-release-cloud-green`
+# refuses to release a commit it has not blessed.
+#
+# ---- One-time setup required before this is useful ----
+# The dispatcher service account below must exist and be impersonable by THIS
+# workflow ref. The existing WIF provider pins `attribute.workflow_ref`, so a
+# new workflow file cannot reuse auto-heal's service account. See
+# docs/contributors/pdd-cli-release-process.md ("Cloud gate setup").
+
+# Deliberately NOT triggered per-push. The gate takes up to 2h; main receives
+# ~15 commits/day at a 7-minute median gap, so a push trigger can never keep up
+# and would only pile up abandoned work. A timer blesses whatever main's tip is
+# when it fires, which is what the release actually needs.
+on:
+  schedule:
+    # Every 4 hours. Also catches drift that no commit caused — model catalog
+    # changes and provider-side behaviour shifts have both gone red here with
+    # no commit to blame.
+    - cron: '17 */4 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  id-token: write   # Workload Identity Federation to GCP
+
+# Never cancel a run in flight. submit.sh submits four Cloud Batch jobs (77
+# tasks) and then polls; it has no `gcloud batch jobs delete` on any path, so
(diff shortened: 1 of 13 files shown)


gltanaka:
Added `test(sync): stop pinned-base guards banning every new file`.

Both PRs in this stack failed unit tests on the same two assertions, while `auto-heal` passed and main stayed green. Root cause is independent of this change:

`test_sync_rollout_repair_executes_the_actual_protected_transition` and `test_release_video_opt_out_uses_only_actual_base_owned_paths` build a manifest from a pinned historical base and then assert the **whole tree** is clean. `manifest._ownership_rules` loads ownership only from the protected base tree, and those bases (`dec539aa8`, `c93332e9b`, both 2026-07-25) are frozen. So any file added afterwards can never have a rule in them.

That made the pair a standing ban on adding files to this repo. It went unnoticed because the only additions to main since those pins were two `.pdd/meta/*` files already covered by the repair bridge — these are the first PRs since to add genuinely new files.

Fix scopes the two blanket assertions to paths that existed at the pinned base. The tests' stated purpose is unchanged and the `records` assertions carrying it are untouched. No coverage lost: auto-heal builds the same manifest with `base_ref=origin/main`, a base that moves, and that is where a genuinely unowned new path is caught — confirmed by this stack, where auto-heal correctly rejected these files until #2329 landed their rules.

132/132 in `test_sync_core_pdd_rollout_policy.py` pass locally.

PR #2356, opened 2026-07-28
test(sync): accept schema-3 retirement diagnostics

gltanaka:
## Summary

- extend the protected substitution-test regex with two exact schema-3 retirement diagnostics
- preserve the JSON-only boundary for the subsequent #2316 Phase A authority change

No production, verifier, policy, profile, prompt, metadata, or architecture bytes change.

## Validation

- protected verification-profile suite on current main: 81 passed
- same suite with exact JSON-only Phase A policy overlaid: 81 passed
- Sol-high adversarial review approved `ea0ebcaa95479bfaad4e707805f5b3e77192a0f4`

Test prerequisite for #2316; does not close it.

### DIFF
--- a/tests/test_sync_core_verification_profiles.py
+++ b/tests/test_sync_core_verification_profiles.py
@@ -2150,7 +2150,9 @@ def test_estimate_contract_rotations_reject_substitution(
                 "candidate (?:requirement transition "
                 "(?:lacks protected authorization|rules are duplicated or ambiguous)"
                 "|removed unconsumed protected requirement transition"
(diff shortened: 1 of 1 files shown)


gltanaka (commented):
Sol-high adversarial review approved exact commit ea0ebcaa95479bfaad4e707805f5b3e77192a0f4. The anchored regex adds only two exact fail-closed schema-3 diagnostics and still requires VerificationProfileError. Formal self-approval is not permitted by GitHub.

PR #2368, opened 2026-08-04
edit prompting guide: frontend prompt and SVG units

### LINKED ISSUE #2367: Update prompting_guide.md for better frontend `.prompt` file abstractions
I'm going to update the `prompting_guide.md` to specify how to write '.prompt' files for frontend components.

Currently, just having a natural language text .prompt file for each frontend code file is not good for representing frontend layout and appearance. We want a better source-of-truth "abstraction" for frontend files, than just the natural language prompt files.

> For now, the idea is to use both `.svg` files and a `.prompt` file to correspond to frontend components. The `svg` files will provide visual abstractions that are easier to iterate on.

## Solution and scope

Update `docs/prompting_guide.md` only. Add clear instructions for choosing frontend requirements files, describing states and screen sizes, using SVG safely, separating requirements from examples and test results, preventing multiple generators from writing the same code file, and checking behavior, appearance, and accessibility after regeneration.

niti-go:
## Summary

- document prompt + measured semantic SVG source bundles for frontend units whose spatial hierarchy or appearance matters
- make brownfield visual views structurally exact but decoratively representative: preserve measured geometry, fold, responsive modes, token colors, type hierarchy, and controls while simplifying repeated ornament and artwork
- define a coherent frontend unit as one canonical generator that may produce one or several target code files
- show one-writer-safe `.pddrc` routing: one context/code path per target, with member contexts repeating the same canonical UI prompt path
- require target manifests and target-specific contracts inside a shared UI prompt instead of per-file TypeScript/React wrapper prompts
- distinguish visual UI units from prompt-only providers, hooks, metadata layouts, and controllers
- document semantic SVG contents, context-only surrounding UI, same-viewport overlay review, and behavior/visual verification
- reconcile the quickstart, scaling, and naming guidance with the frontend exception to ordinary one-prompt/one-file organization

## Validation

- `git diff --check upstream/main...HEAD`
- documentation-only diff: `pdd/docs/prompting_guide.md`
- reviewed all existing prompting-guide sections for conflicting one-prompt/one-file language

Closes #2367


niti-go:
Updated the frontend SVG guidance to use durable semantic names: `<state>[--<variant>]__<layout-mode>.ui.svg` (for example, `ready__split-pane.ui.svg`) instead of fixed pixel dimensions. The prompt now records the layout and when it applies; SVG `viewBox` is scalable drawing coordinates, not a required display size. Exact dimensions remain explicit only for fixed-output contracts such as social cards, video, print, or fixed displays, while concrete viewports are verification parameters.

This keeps frontend prompt source responsive and reusable across surfaces instead of binding its identity to today's resolutions or device labels.

niti-go:
Decision: frontend source filenames describe what a view means, not the device or display where it happens to appear.

This PR now says that names keep the stable state and layout arrangement (for example, `ready__split-pane.ui.svg`). They must not include 2D/3D, AR/VR, phone/desktop, orientation, DPI, pixels, or aspect ratio. If the arrangement truly changes, name that arrangement instead.

The file extension and its tooling distinguish SVG, raster, and future 3D sources. Exact viewport or aspect details stay in the prompt’s view contract. No existing files were renamed: the current semantic naming already fits this rule.

niti-go:
Updated the guide so frontend prompt folders use short, plain names for the interface they describe—for example, `account-header` or `project-sharing-panel`—instead of long names copied from the current source-file path. The folder and prompt file use the same name, making the right prompt easier to find even after code moves.

The existing SVG rule remains: names such as `stacked` and `split-pane` describe what people see arranged on screen. Names such as `mobile` or `landscape` only describe one current device or screen shape, so they become misleading when the same arrangement is used elsewhere.

PR #2370, opened 2026-08-05
Add Language Validity Gate to code_generator_main

agarwal-ishaan:
## Summary

Adds a new gate (5b, in `code_generator_main`'s validation pipeline) that checks generated content is syntactically valid Python before any downstream gate — architecture conformance, public-surface regression — assumes it can be parsed.

**Problem**: a prompt declared as Python (via the `_python.prompt` filename convention) whose generated output was actually a different language (e.g. the model produced HTML instead of Python) would silently get written to a `.py` file with no validation catching the mismatch.

## What changed

- `pdd/prompts/code_generator_main_python.prompt`: added the `5b. Language Validity Gate` requirement (source of truth for the behavior below), renumbered the existing `5b`/`5c` gates to `5c`/`5d`, updated the `<pdd-interface>`, Deliverables, and `<architecture_sync>` include selector accordingly.
- `pdd/code_generator_main.py`:
  - `LanguageMismatchError(click.UsageError)` — a peer of `ProseOutputError` (not a subclass). Carries structured attributes (`prompt_name`, `output_path`, `language`, `model_name`, `total_cost`, `raw_output_excerpt`, `syntax_error`) but deliberately has **no** `repair_directive` property — it hard-fails immediately rather than feeding an automatic retry loop.
  - `_verify_language_validity()` — Python-only (no-op for every other declared language), via `ast.parse`, with a `textwrap.dedent` retry to avoid false positives on indented/fence-extraction leftovers.
  - Wired in immediately after the existing Prose/Empty-Output gate and before architecture conformance.
- `tests/test_code_generator_main.py`:
  - 6 new tests (`TestVerifyLanguageValidityHelper`, `TestLanguageMismatchErrorGate`) covering the helper directly and the gate through `code_generator_main`.
  - Fixed 6 pre-existing tests whose mocked "generated code" was a plain English placeholder string (e.g. `"Updated code"`) that never needed to parse as real Python before this gate existed — swapped for minimal valid-Python equivalents (e.g. `"def updated(): pass"`) with no change to what each test actually verifies.

## Test plan

- [x] `pytest tests/test_code_generator_main.py` — full suite passes (379 passed)
- [x] New gate verified to fire before architecture conformance runs
- [x] Verified the target file is left untouched when the gate raises
- [x] Verified the gate is a no-op for non-Python declared languages, including on content that would never parse as Python


agarwal-ishaan:
Verified locally as well

image

agarwal-ishaan:
Good to merge, all test cases pass, rebased with new master.

gltanaka (changes requested):
Request changes — the problem is real and a Python language gate is needed, but this head is not merge-ready.

P1 — `pdd/code_generator_main.py:482-487`: the dedent retry validates different bytes from the bytes later written. For uniformly indented source, `ast.parse(original)` raises `unexpected indent`, `_verify_language_validity` returns `None` because `ast.parse(textwrap.dedent(original))` succeeds, and `code_generator_main` still writes the original invalid source at line 5838. I reproduced a successful first generation whose saved `.py` file does not parse. Propagate the successfully parsed candidate into `generated_code_content` or reject the original; add an integration assertion that `ast.parse(output.read_text())` succeeds.

P1 — `pdd/code_generator_main.py:5609-5612`: skipping the language gate whenever a non-empty output exists leaves a data-loss path. The downstream public-surface syntax guard runs only when the previous public surface is non-empty (`pdd/code_generator_main.py:2685`). I reproduced an existing private/script-only `.py` file being successfully overwritten by HTML, leaving syntax-invalid Python. Run the validity gate for every generated Python candidate before any write, preserve the existing file on failure, and update Requirement 5b plus the existing-file test. This also avoids the current fallback path where an `ArchitectureConformanceError` raised inside the compatibility block can lose attempted model/cost metadata because line 5751 catches only `PublicSurfaceRegressionError` and `TestChurnError`.

P1 — `.pdd/verification-profile-rotations.json:676-699`: this Phase B changes the protected rotation policy while consuming the row installed by #2382. `docs/ci.md:143-152` requires Phase B to leave Phase-A policy unchanged and forbids combining replacement authority with consumption. The protected base still has the older code-generator row, and the appended retirement is what removes the ambiguous active identity; candidate-local exact-state logic in `pdd/sync_core/verification.py:2948-3073` then accepts the final stationary state. Use separate protected governance steps: land a verifier recovery prerequisite if needed, land the retirement as a policy-only change with prompt/profile bytes unchanged, land any consumed-state verifier prerequisite, then rebase this Phase B so it changes only the preauthorized prompt/profile bytes and product/test files without changing rotation authority.

Validation: exact head `0f0307013137e267d7bae71a80ac14870c61de7d`; full diff reviewed. PDD Python 3.12.12: `tests/test_code_generator_main.py` 380 passed; new-gate selection 7 passed; current-profile reconciliation 1 passed; `py_compile` and `git diff --check` passed. The two overwrite failures above were independently reproduced by both the synthesis owner and the adversarial reviewer. CI is green, but the new tests encode the unsafe dedent behavior and the existing-file test asserts only that some non-language exception occurs, so green tests do not disprove either blocker.

Contributor validation is not sufficient end-to-end for this safety boundary: the PR evidence and tests call mocked internals, with no safe Click/`pdd generate` boundary proving nonzero exit and byte-identical target preservation. Please add that deterministic CLI-level check. Also rebase/coordinate with overlapping open PR #2374 after these changes.

gltanaka (changes requested):
Delta review of `77feedb0c5bea13a529336cd026564133af4e7c5` (`delta-since-0f030701`): the two data-loss findings are resolved, but this head is still not merge-ready.

Resolved: `_verify_language_validity` now returns and writes the exact candidate that parsed; every Python candidate is validated before any write; existing private/script-only files remain byte-identical on failure; failed-attempt model/cost metadata is retained; and the new deterministic Click boundary tests prove non-zero failure plus preservation. I independently reproduced those fixes.

P1 — `.pdd/verification-profile-rotations.json:640-648` and `:688-698`: Phase B still self-authorizes its corrected prompt bytes. The protected row installed by #2382 authorizes `e06c…` with profile `af2b…`; this candidate rewrites that protected authority to its new `a803…` prompt and `d581…` profile, rewrites the retirement replacement, consumes those bytes, and re-pins the verifier at `pdd/sync_core/verification.py:177-184` in the same PR. `docs/ci.md:143-152` explicitly says byte drift after Phase A requires a new protected Phase A and forbids editing the digests in Phase B. Leave protected rows/verifier bindings unchanged here; land the necessary byte-bound governance recovery and fresh authority as separate protected changes, then consume those unchanged bytes in a later Phase B.

P1 — `pdd/prompts/code_generator_main_python.prompt:103-104`: the source-of-truth contract is impossible. Line 103 requires `LanguageMismatchError.repair_directive`; the next line forbids that property and requires terminal/no-retry behavior. Runtime currently exposes the property, but `sync_main`, `sync_orchestration`, and subprocess routing do not catch/parse `LanguageMismatchError`, so the directive is dead and behavior remains terminal. Choose one contract. The smallest change consistent with existing behavior is to remove the new directive requirement/property/constant; if retry is intended, remove the terminal requirement and wire the exception through all retry/parser boundaries with tests.

Validation on the current head with the PDD Python 3.12 environment: `tests/test_code_generator_main.py` 385 passed; focused helper/direct-gate/Click boundary selection 12 passed; current-profile reconciliation 1 passed; `py_compile`, product-module pylint, and `git diff --check` passed. CI is green. No live provider was used or needed for these deterministic safety checks.

Verdict: BLOCK until the protected transition is sequenced correctly and the prompt contract is made internally consistent.

agarwal-ishaan:
P1 — .pdd/verification-profile-rotations.json:640-648 and :688-698: Phase B still self-authorizes its corrected prompt bytes. The protected row installed by https://github.com/promptdriven/pdd/pull/2382 authorizes e06c… with profile af2b…; this candidate rewrites that protected authority to its new a803… prompt and d581… profile, rewrites the retirement replacement, consumes those bytes, and re-pins the verifier at pdd/sync_core/verification.py:177-184 in the same PR. docs/ci.md:143-152 explicitly says byte drift after Phase A requires a new protected Phase A and forbids editing the digests in Phase B. Leave protected rows/verifier bindings unchanged here; land the necessary byte-bound governance recovery and fresh authority as separate protected changes, then consume those unchanged bytes in a later Phase B.

This is a test case fix, I can craete two seperate PRs for this fix, if you think it is needed, but we have updated the hash values correctly in the code.

gltanaka (dismissed):
Delta review of current head `2663a5a1a7b66a3ce05c9cd2f4703886b0b3c815`: the LanguageMismatchError contract blocker is resolved, but the protected rollout blocker is not, so this still should not merge.

Resolved: `pdd/prompts/code_generator_main_python.prompt:97-103` now defines one unambiguous terminal contract, and `pdd/code_generator_main.py:440-479` removes the unused `repair_directive` property and constant. Independent checks confirm the exception has no directive. The earlier functional fixes also remain correct: every generated Python candidate is gated, the exact dedented candidate that parsed is written, invalid output preserves existing public and private/script-only files, and Click-boundary tests cover the behavior.

P1 — `.pdd/verification-profile-rotations.json:640-648` and `:688-697`: this update re-pins the same protected Phase-A authority yet again, now from the prior candidate values to prompt `882876…` and profile `6e5891…`; compared with protected `origin/main`, the authorized target is still `e06cce…` / `af2beb…`. The PR also consumes those rewritten bytes and re-pins their accepted stationary digests in `pdd/sync_core/verification.py:175-184`. `docs/ci.md:143-152` explicitly says byte drift after Phase A invalidates the transition, forbids editing the digests in Phase B or combining replacement authority with consumption, and requires a new separately protected dormant Phase A row. Passing current-profile reconciliation here only shows that the self-consistent candidate matches its own rewritten authority; it does not establish protected authorization.

Action: keep this Phase B from rewriting protected rotation rows or verifier bindings. Land the needed governance recovery/fresh byte-bound authority as separately protected prerequisite change(s), then rebase and consume those unchanged authorized prompt/profile bytes in a later Phase B.

Validation on exact current head with PDD Python 3.12: focused helper/direct/Click language-gate tests 12 passed; independent no-directive and dedent/parse checks passed; full `tests/test_code_generator_main.py` passed (385 tests); current-profile reconciliation passed; `py_compile`, product-module pylint, and `git diff --check` passed. All current GitHub checks are green. No live provider call was necessary for these deterministic safety and governance checks.

Verdict: CHANGES REQUESTED. The code-path issue is addressed; protected rollout sequencing remains a merge blocker.

PR #2373, opened 2026-08-06
fix: point pdd-change Step 6 dependency context at <pdd-dependency>, not <include>

### LINKED ISSUE #1807: Make prompt files properly use <pdd-dependency> tags and <include> tags so we can trust our architecture.json
### pdd change Step 6 currently builds its dependency context from `<include>` tags. It should use `<pdd-dependency>` tags instead.

We need to update the prompting_guide.md to learn how to use `<include>` tags properly and `<pdd-dependency>` tags properly. This way, we can trust that our prompt files use `<pdd-dependency>` tags properly. Then, we can trust our `architecture.json` to provide correct information to be fed as context to pdd commands like pdd-change Step 6.

Here's the difference I think we should highlight between the two tags: 
1.  `<pdd-dependency>` is to declare architectural module dependencies. It shows that a module really depends on the module. **If a prompt file 'customer_service' contains `<pdd-dependency>payment_status_python.prompt</pdd-dependency>`, and we update `payment_status_python.prompt` then we definitely might want to update `customer_service`**
2. `<include>` tags are simply used to give context to the model in the prompt. If a prompt file 'customer_service' contains <include> tags like:
```
<include>docs/prompting_guide.md</include>
<include>context/refund_policy_examples.md</include>
<include>docs/customer_support_tone.md</include>
```
These are just context given to the LLM during time of reading the prompt. These help the LLM write the module, but they are not modules that `customer_service` necessarily depends on. **Even if we update `refund_policy_examples.md`, that doesn't mean we need to update the `customer_service` module.** 

## Why this matters

`pdd` commands like `pdd-change` require understanding the module dependencies in the codebase, so that it knows which prompt files might need updating/changing when working on a feature. If pdd change decides to update one module's prompt, then it needs to know which other modules' prompts it may need to update as well. Currently, pdd change scans all prompt files for `<include>` tags, and from that, builds up a large plain markdown string to show as dependency context. **However, `<include>` tags are not what define module dependencies; `<pdd-dependency>` tags do.**

The current markdown string injected into pdd change prompts looks like this: 
```
## Module Dependency Graph

  When a module is modified, all modules that depend on it (directly or transitively) may also need updates.

  ### Modules and their dependents (modules that will be affected if changed):

  - **module_a** → affects: module_b, module_c
  - **module_x** → affects: module_y

  Total modules tracked: 42
```
It actually limits to only the top 30 modules and shows at most 10 dependents, so this is very limiting context for large codebases.

If we can instead use the architecture.json, then our markdown string could look something like this instead:

### The prompting_guide.md already distinguishes somewhat between the two tags: 
It directly says:

```
<pdd-dependency>: Declares architectural dependency (updates architecture.json)
<include>: Injects content into prompt for LLM context (does NOT affect architecture)
Use both when appropriate - they serve different purposes
```

So maybe our `prompting_guide.md` should just be slightly tweaked to explain this difference with regards to how it affects prompts and have some examples:

````
1.  `<pdd-dependency>`: Declare architectural module dependencies (updates architecture.json). It shows that a module really depends on the module. **If a prompt file 'customer_service' contains `<pdd-dependency>payment_status_python.prompt</pdd-dependency>`, and we update `payment_status_python.prompt` then we definitely might want to update `customer_service`**
2. `<include>`: Injects content into prompt for LLM context (does NOT affect architecture). These are simply used to give context to the model in the prompt. If a prompt file 'customer_service' contains <include> tags like:
```
<include>docs/prompting_guide.md</include>
<include>context/refund_policy_examples.md</include>
<include>docs/customer_support_tone.md</include>
```
These are just context given to the LLM during time of reading the prompt. These help the LLM write the module, but they are not modules that `customer_service` necessarily depends on. **Even if we update `refund_policy_examples.md`, that doesn't mean we need to update the `customer_service` module.**
````

agarwal-ishaan:
## Summary
- `pdd change` Step 6 builds its module-dependency context by scanning `<include>` tags (`build_dependency_graph()`), but `<include>` tags are LLM context, not declared architectural dependencies -- so the graph both misses real `<pdd-dependency>`-declared dependents and fabricates false ones from stylistic includes.
- Updates the source prompts to specify a new `sync_order.build_dependency_graph_from_architecture()` that reads `architecture.json`'s `dependencies` field (already populated from `<pdd-dependency>` tags by `architecture_sync.py`), and wires Step 6's `_build_dependency_context()` to use it instead.
- **Prompts only in this PR.** The corresponding code regeneration (`pdd generate --incremental` on `pdd/sync_order.py` and `pdd/agentic_change_orchestrator.py`) is pending in a follow-up PR -- Anthropic API access is currently blocked by a billing issue, and the fallback Gemini model hits its output-token ceiling trying to patch the ~4000-line orchestrator file. Once resolved, the code PR will implement exactly what's specified here.

Fixes #1807

## Test plan
- [ ] Follow-up PR regenerates `pdd/sync_order.py` and `pdd/agentic_change_orchestrator.py` from these prompts via `pdd generate --incremental`
- [ ] Unit tests for `build_dependency_graph_from_architecture()` and the updated `_build_dependency_context()`
- [ ] Manual repro: a module declaring `<pdd-dependency>` on another without `<include>`-ing it shows up correctly as a dependent; a module `<include>`-ing another purely for style context does not

### DIFF
--- a/pdd/prompts/agentic_change_orchestrator_python.prompt
+++ b/pdd/prompts/agentic_change_orchestrator_python.prompt
@@ -321,14 +321,16 @@ Before the step loop begins, load `.pddrc` configuration to provide context keys
 - Step 13 receives: issue_content, step8_output, worktree_path, files_to_stage, issue_title, sync_order_script, sync_order_list, impacted_tests, base_branch, **head_branch** (the worktree's actual current branch, derived from its HEAD and defaulting to `change/issue-{N}`; it is a `change/issue-{N}-job-*` fallback when the canonical branch was locked under clean restart — Step 13 pushes and opens/updates the PR on this branch rather than a hardcoded `change/issue-{N}`. The existing-PR guard also matches `change/issue-{N}-job-*` heads so a later normal run does not open a duplicate), **user_story_summary** (literal `None` when no story was generated/attempted, otherwise a PR-body markdown section naming generated `user_stories/story__*.md` artifacts and linked prompts), **manual_review_lines** (aggregated `MANUAL_REVIEW:` lines from Step 9 JSON/prose, Step 10 synthesized `associated_docs_conflicts`, and Step 11/12 review-loop JSON/prose manual-review entries; rendered in the PR body's "Manual Review Required" section. The Step 12.5 gate no longer contributes here — a gate failure now hard-blocks before Step 13 rather than surfacing as an advisory line), **clean_restart** (`"true"` or `"false"`) — Step 13 uses this to decide whether to force-push (`--force-with-lease`) and whether to look for an existing open PR to update in place rather than creating a new one
 
 % Dependency Context (Before Step 6)
-Before Step 6 runs, build a formatted dependency graph to help identify transitively affected modules:
+Before Step 6 runs, build a formatted dependency graph to help identify transitively affected modules. This must reflect declared architectural dependencies (`<pdd-dependency>` tags, as recorded in architecture.json), not the `<include>` context graph — see issue #1807: `<include>` tags are LLM context and are not a reliable signal of "what needs updating if this module changes."
 
-Helper: `_build_dependency_context(prompts_dir: Path, quiet: bool = False) -> str`:
-- Build dependency graph using `build_dependency_graph(prompts_dir)`
+Helper: `_build_dependency_context(architecture_path: Path, quiet: bool = False) -> str`:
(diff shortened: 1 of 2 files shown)


agarwal-ishaan:
Consolidating into a single PR per request -- superseded by #2376, which contains this PR's commits plus the rest of the fix. Closing without merging (nothing is lost; #2376's branch includes these exact commits).

Issue #2256, opened 2026-07-21
reconcile ignores explicit nested .pddrc test mapping and reports a missing inferred test


niti-go:
## Problem

`pdd reconcile --module <name> --strict --json` can ignore an explicit dev-unit test mapping in a nested project `.pddrc` and instead infer a different nonexistent test path. The unit is then reported as `missing_artifacts: test`, preventing prompt/code/test fingerprint reconciliation even though the configured test exists.

## Reproduction

1. In a nested PDD project, configure a dev unit whose aliases include the prompt/code/module name and whose explicit artifacts map:
   - prompt: `prompts/src/workers/runtime/failure_classification_Python.prompt`
   - code: `src/workers/runtime/failure_classification.py`
   - test: `tests/test_failure_classification.py`
   - example: `examples/failure_classification_example.py`
2. Confirm the configured test file exists.
3. Run `pdd reconcile --module failure_classification --strict --json` from the nested project root.
4. Observe that the JSON resolves test to `tests/pdd_executor/test_failure_classification.py` and fails with `missing fingerprinted artifacts: test`, rather than using `tests/test_failure_classification.py` from `.pddrc`.

## Expected behavior

Explicit artifact mappings must take precedence over inferred conventional paths for reconcile/sync/baseline. The JSON result should report the configured test and reconcile the same dev unit selected by its aliases.

## Actual behavior

Module selection recognizes the configured unit name, but artifact resolution substitutes an inferred test path from the code namespace. This makes the configuration internally inconsistent and blocks strict reconciliation.

## Proposed solution

Centralize dev-unit resolution so module alias selection and artifact path selection return one resolved configured unit. Only infer a conventional test when the selected unit has no explicit test artifact. Add a nested-project regression with an explicit non-conventional test path; assert `sync`, `reconcile`, and `baseline` all resolve the identical prompt/code/test/example set and that JSON reports the configured test path.

Issue #1968, opened 2026-07-09
sync: repair directive does not converge on annotation-level surface drift (object vs Any, union broadening) — retry regenerates identical violation


Serhan-Asad:
## Summary

During staging E2E validation (promptdriven/test_repo#4248, sync job `RHdp2mgIQYMZvwdfqZtj`, pdd 0.0.299.dev147 from the e2e stack), the sync surface-contract gate correctly rejected a regeneration for annotation-level drift — but the **repair directive retry regenerated the exact same drift**, so the repair loop cannot converge on this class of violation.

## Evidence

Module `extensions/github_pdd_app/src/clients/secrets_client`:

```
Architecture conformance failed; retrying generation with repair directive
(PublicSurfaceRegressionError): Public surface regression
signature_detail: {"symbol": "_list_secrets_with_retry",
  "expected": "[function] (client: secretmanager.SecretManagerServiceClient, request: dict[str, object], *, parent: str) -> list[Any]",
  "actual":   "[function] (client: secretmanager.SecretManagerServiceClient, request: dict[str, Any] | secretmanager.ListSecretsRequest, *, parent: str) -> list[Any]",
  "source": "pdd-interface"}
signature_detail: {"symbol": "list_secret_metadata",
  "expected": "[function] (filter_str: str = '', max_results: int | None = None) -> list[dict[str, object]]",
  "actual":   "[function] (filter_str: str='', max_results: int | None=None) -> list[dict[str, Any]]",
  "source": "pdd-interface"}
```

The declared `<pdd-interface>` says `dict[str, object]`; the regenerated code consistently writes `dict[str, Any]` (and broadens the `request` parameter to a union). The repair directive told the generator about the violation; the retry produced the identical annotations again and the gate failed the module.

Reproduce: `cd extensions/github_pdd_app && pdd --context clients sync src/clients/secrets_client` (on promptdriven/test_repo main + PR #4254's branch).

## Why this is universal

`object` vs `Any` (and equivalent-but-broader unions) is the most common LLM annotation preference drift. Any user whose prompt declares one spelling will hit a hard, repair-proof gate failure whenever regen prefers the other — with no code-behavior difference at all. The repair loop's one retry is wasted because nothing forces the generator to treat the declared annotation as verbatim-binding.

## Fix directions (either/both)

[...]

Issue #1900, opened 2026-07-07
sync: enforce the prompt-declared interface as the surface contract with self-repair, instead of prose BREAKING-CHANGE permits


Serhan-Asad:
## Problem

The public-surface gate added in #1012 (commit `8f68c5e9f`) compares the **old generated code vs the newly generated code**, so it cannot distinguish an intended interface change from generator drift. The tie-breaker is a prose `BREAKING-CHANGE:` directive in the prompt body — a permission slip that no command writes automatically, so the standard `pdd change → pdd sync` flow fails whenever a module's interface legitimately evolves.

## Live evidence (promptdriven/pdd_cloud#2971, 2026-07-07)

Agentic sync job `sXz0zXHU5w4EQ1RLhVVt`, pdd `0.0.297.dev1`:

- `pdd change` correctly added `get_secret_with_enabled_fallback` to the prompt's `__all__` and structured interface section (surface 13 → 14) — PR promptdriven/pdd_cloud#2986. No prose directive emitted (change never does).
- `pdd sync` regeneration **also mutated the signature of `list_secret_metadata`** — an existing public function the prompt edit never touched. That is generator drift, the exact #1010/#1012 disease.
- Gate raised `PublicSurfaceRegressionError` and the run hard-failed:

```
signature_changed: list_secret_metadata)
pre surface size: 13
post surface size: 14
removed: <none>
```

Note: the **addition alone would have passed** (gate fires only on removals/signature changes). The drift is what tripped it — so teaching `pdd change` to auto-emit a `BREAKING-CHANGE:` directive would NOT have saved this run: change could only have declared the addition, not drift it never intended.

## Two structural gaps

1. **Intent has no first-class home.** The prompt already declares the interface — `__all__` plus per-symbol signature entries — and `pdd change` already maintains that section. But the gate never reads it; it only diffs code-vs-code and demands prose permits.
2. **Violations dead-end instead of self-repairing.** The gate ships `repair_directive` / `PDD_REPAIR_DIRECTIVE` plumbing, but the #2971 run shows a single-shot hard fail (whether the loop engaged in the agentic path is tracked in the companion pdd_cloud issue). More fundamentally, a code-vs-code gate has no known-correct answer to repair toward; a prompt-declared contract does.

## Expected behavior

1. **When the prompt declares an interface** (`__all__` and/or structured signature entries), sync validates the generated surface **against the prompt declaration** — not against the previous generation:
[...]

PR #2374, opened 2026-08-06
Split code_generator_main into pdd/conformance/ (six gate units)

agarwal-ishaan:
Extracts the four generation safety gates out of `code_generator_main` into `pdd/conformance/`, and pins their behaviour with a user story that survives the reorganisation.

**`pdd/code_generator_main.py`: 5,879 → 2,033 lines.** Prompts: 71,531 → 51,628 bytes (−27.8%).

## Why

Four gates run before PDD writes any generated file to disk — prose-instead-of-code, public-surface regression, test churn, and architecture/interface conformance. They are the reason you can re-run `pdd sync` on a mature module without fear. They were tangled into a 5,879-line module alongside path resolution, cloud routing, git time-travel and post-processing, so `sync_main`, `cmd_test_main` and `agentic_test_generate` had to import the whole generation pipeline just to *catch* a gate exception.

The issue puts the risk better than I can paraphrase: a safety check that stops firing does not produce an error — it produces a successful-looking run.

## The six units

| module | lines | prompt ratio | owns |
|---|---|---|---|
| `gate_errors` | 390 | 23% | the four typed exceptions, churn nonce |
| `directives` | 197 | 30% | `BREAKING-CHANGE` grammar, env flags |
| `surface` | 1,613 | 14% | `__all__` resolution, surface + signature snapshots, `@dataclass` ctor synthesis |
| `test_churn` | 275 | 30% | churn gate, threshold, test-path taxonomy |
| `interface_check` | 729 | 19% | architecture + `<pdd-interface>` conformance |
| `declared_surface` | 800 | 24% | public-surface regression gate + annotation reconciler |

Dependency graph is acyclic: `gate_errors`, `directives`, `surface` are leaves; `test_churn` and `interface_check` sit on `gate_errors`; `declared_surface` sits on all four.

**This started as nine units and was consolidated to six.** Four had no consumer outside the group. `surface`/`signatures`/`dataclass_signatures` were one AST walk cut in three — with the `__all__`-precedence rule stated in *two prompts required to agree forever*; merging made that invariant structural. `annotation_reconcile` was one private function serving one caller, and reads the same declared-interface map the gate reads.

**All six prompts sit inside the prompting guide's 10–30% prompt-to-code band**, after the condensing pass in `2fb8c6fe8` brought `gate_errors`, `directives` and `test_churn` down from 44%, 52% and 61% respectively.

## Extraction, not regeneration

Every moved definition is an **exact source slice**. This code carries fixes accumulated across #1012, #1662, #1663, #2256, #1900, #1558, #1612, #1968, #1446, #1903 and #1649. The story's premise is that behaviour must not change — re-deriving 2,500 lines of AST logic from prose would gamble precisely what the change exists to protect.

## Orchestrator surgery — three hazards that had to move together

- **H1** — the self-include selector trimmed to 21 symbols. A moved symbol left there would make the next `pdd sync` regenerate the monolith and *silently revert the split*.
- **H2** — declarations removed from both the `<pdd-interface>` block and the `architecture.json` entry. Declaring a symbol the module no longer defines fails its own conformance check.
- **H3** — all 91 moved symbols re-exported via **redundant aliases** (`X as X`), so every existing `from pdd.code_generator_main import ...` call site keeps resolving with zero consumer changes. A plain import would read as symbol removal under this project's own public-surface rules.

## Two things Phase B taught us that the paper design couldn't

**A missing dependency.** Both `declared_surface` functions consult `test_churn`'s test-path taxonomy to exempt test files from the surface gate. Recorded in `architecture.json`, the prompt, and R1. The alternative — a second copy of the taxonomy — is the two-copies-must-agree hazard the `surface` merge existed to remove.

**A re-export limit.** `test_churn_block_stamps_parent_nonce_from_fd` resets a once-per-process cache via `cg._CHURN_NONCE_READ = False`. Rebinding a re-export alias cannot reach the original module's global, so the reset silently stopped working. Two-line test change; observable behaviour unchanged. This is the seam flagged in the original Phase A notes as the one that fails quietly rather than loudly — it did exactly that.

## Prompt/code signature drift — found in review, fixed

Ten declared signatures disagreed with the code they are supposed to regenerate:

| module | n | worst case |
|---|---|---|
| `surface` | 5 | `_effective_patch_targets` was an entirely different function; two snapshots declared `file_path` where the code takes `patch_targets` |
| `interface_check` | 2 | `_verify_architecture_conformance` omitted required `arch_path` and `verbose`; `_verify_pdd_interface_signatures` had different arity *and* order |
| `declared_surface` | 3 | argument order, a default, `str` vs `Optional[str]` |

Not a documentation defect: `code_generator_main:1761` calls that function with `arch_path=` and `verbose=`, and `declared_surface` passes `patch_targets=` at four call sites, so `pdd sync` on those units would emit a module failing at its first call. Worse, `declared_surface` R5 judges declared symbols *against the declaration*, so a wrong declaration becomes the enforced shape.

Signatures were back-propagated from code, never the reverse. `validate_conformance_prompts.py` now parses each module's AST and requires declared signatures to equal real ones — verified by re-introducing a drifted signature and confirming FAIL.

## A latent `main` bug this surfaced

Adding `<contract_rules>` to the orchestrator exposed it: a bare `<include>` written in prose paired with the real closing tag in Dependencies and swallowed 7,328 characters — the whole rules block. `contract_ir`'s section scan is non-greedy but non-overlapping, so the block was invisible to `pdd contracts check` while a line-anchored search still found it. Green validation, zero coverage evidence. Fixed, plus a validator check mirroring `contract_ir`'s own regex.

## Verification

- **`pdd detect --stories`** → **PASS** (strength 0.9). Failed at default 0.2, which surfaced two genuine gaps — error precedence when several gates fail, and repair-directive content — both now numbered rules.
- **Full 489-file suite vs a `main` worktree**, identical files and order: branch 146 failed / 15,462 passed; main 130 / 15,478. **17 branch-only failures, all in the two digest-pinned files**, and all 17 fail identically at the branch point `eacf2a0f6` — they predate this work. One test passes here that fails on `main`.
- **Manual end-to-end**: generated a module, deleted a public function, gate refused with `removed: f_to_c` and the file stayed byte-identical; added one `BREAKING-CHANGE: remove f_to_c` line and the same regeneration succeeded.
- `validate_conformance_prompts.py` OK · `prompt_lint` 0 issues · 106 contract rules parsing · `pytest -m story` 17 passed.

## Digest-pinned suites

Editing any managed prompt moves digests that PDD pins in `verification.py` constants and `.pdd/*.json` rows, and adding tracked files requires an ownership entry whose own sha256 is pinned in `manifest.py`. Both are handled on this branch:

- `b4bcf77e5` registers the conformance inventory in `.pdd/sync-ownership.json` and `.pdd/expected-managed.json`
- `abfdd28b4` retires the superseded codegen fingerprint
- `b0a6adba8` retires the active codegen rotation

with the accompanying `pdd/sync_core/manifest.py` and `pdd/sync_core/verification.py` changes. `test_sync_core_pdd_rollout_policy.py` and `test_sync_core_verification_profiles.py` are being re-verified against this state.

## Not in scope

`.github/workflows/unit-tests.yml` is reverted to `main` (`944ab6d04`). The manual-dispatch trigger and the 60→360 minute timeout extension are unrelated to the conformance split and belong in their own change.

## Prompt/code drift logged at the start — all resolved

The three divergences recorded in `0420974be` are closed:

- **D1/D2** — the pre-split prompt claimed repair-directive wording the code never emitted. `gate_errors_python.prompt` now states the actual strings (`Public surface regression repair required.`, `Test churn repair required.`) as part of the documented cross-process message contract.
- **D3** — `TestChurnError.adopted_human` is now declared in both the prompt interface and `architecture.json`, with a rule that it only annotates the raised error and never changes the gate decision.

The related open question — whether the repair directive should name the `BREAKING-CHANGE` escape hatch — resolves as *deliberately not*: the prompt now directs intentional declared changes to the `<pdd-interface>` declaration rather than to `BREAKING-CHANGE`, which is consistent with `declared_surface` R7 (a marker must not excuse a still-declared symbol) and R14. The old prompt wording predated the #1900 declared-interface work.

## Ordering vs #2370

Based on `main`, so the Language Validity Gate is not included. This lands first; #2370 is re-targeted on top afterwards, with `LanguageMismatchError` authored into `conformance/gate_errors_python.prompt` alongside the other four.


agarwal-ishaan:
## Status update — Phase B complete, prompts condensed, verified against a `main` baseline

*(This comment supersedes earlier size figures, which predated the prompt condensing in `2fb8c6fe8`.)*

Since this PR was opened as a prompt-only split, it now also ships the extracted code.

### Scope: nine prompt units → six units + code

Four of the original nine had no consumer outside the group. `surface`/`signatures`/`dataclass_signatures` were one AST walk cut in three, with the `__all__`-precedence rule stated in **two prompts required to agree forever** — merging made that invariant structural. `annotation_reconcile` was one private function serving one caller.

`pdd/code_generator_main.py`: **5,879 → 2,033 lines.** Prompts: **71,531 → 51,628 bytes (−27.8%).**

### The six units

| module | lines | prompt ratio | owns |
|---|---|---|---|
| `gate_errors` | 390 | 23% | the four typed exceptions, churn nonce |
| `directives` | 197 | 30% | `BREAKING-CHANGE` grammar, env flags |
| `surface` | 1,613 | 14% | `__all__` resolution, surface + signature snapshots, `@dataclass` ctor synthesis |
| `test_churn` | 275 | 30% | churn gate, threshold, test-path taxonomy |
| `interface_check` | 729 | 19% | architecture + `<pdd-interface>` conformance |
| `declared_surface` | 800 | 24% | public-surface regression gate + annotation reconciler |

**All six sit inside the prompting guide's 10–30% band.** Acyclic dependency graph: `gate_errors`, `directives`, `surface` are leaves; `test_churn` and `interface_check` build on `gate_errors`; `declared_surface` on all four.

### Extraction, not regeneration

Every moved definition is an exact source slice. This code carries fixes from #1012, #1662, #1663, #2256, #1900, #1558, #1612, #1968, #1446, #1903 and #1649; re-deriving 2,500 lines of AST logic from prose would gamble exactly what the story exists to protect.

### Verification

| | |
|---|---|
| Full 489-file suite vs a `main` worktree, identical order | branch-only failures confined to the digest-pinned suites |
| `pdd detect --stories` | **PASS** (strength 0.9) |
| Gate suites (`issue_1900`, `1558`, `1968`, `1903`, `cmd_test_main`, `agentic_sync_runner`) | 858 passed, 0 failed |
| `test_code_generator_main.py` | 373 passed |
| `pytest -m story` | 17 passed · gate `story-regression-ok` |
| `validate_conformance_prompts.py` · `prompt_lint` · `contract_ir` | OK · 0 issues · 100 rules |

One test passes here that fails on `main`.

**Manual end-to-end**, exercising the extracted gates rather than asserting about them: generated a module, deleted a public function → `Public surface regression: removed: f_to_c`, file byte-identical afterwards; added one `BREAKING-CHANGE: remove f_to_c` line → the same regeneration succeeded. Blocked and allowed with a single directive line as the only difference.

### Three findings worth a reviewer's attention

**1. Prompt/code signature drift — 10 corrections.** The `<pdd-interface>` blocks disagreed with the code they regenerate. `_verify_architecture_conformance` omitted required `arch_path`/`verbose`; `_effective_patch_targets` was an entirely different function; two snapshots declared `file_path` where the code takes `patch_targets`. Not cosmetic — `code_generator_main:1761` passes those keywords, so `pdd sync` on those units would emit a module failing at its first call, and `declared_surface` R5 judges declared symbols *against the declaration*, making a wrong declaration the enforced shape. Back-propagated from code; `validate_conformance_prompts.py` now compares declared signatures against each module's AST and fails on drift.

**2. A latent bug on `main`.** A bare `<include>` written in prose pairs with the real closing tag and swallows 7,328 characters — `contract_ir`'s section scan is non-greedy but non-overlapping. It hid an entire `<contract_rules>` block from `pdd contracts check` while a line-anchored search still found it: green validation, zero coverage evidence. Fixed, with a validator check mirroring `contract_ir`'s own regex.

**3. A dependency the paper design missed.** Both `declared_surface` functions consult `test_churn`'s test-path taxonomy to exempt test files. Recorded in `architecture.json`, the prompt and R1 — the alternative is a second copy of the taxonomy, the two-copies-must-agree hazard the `surface` merge removed.

### Digest-pinned suites

The sync/verification work has since landed on this branch — `b4bcf77e5` registers the conformance inventory in `.pdd/sync-ownership.json` and `.pdd/expected-managed.json`, and `abfdd28b4`/`b0a6adba8` retire the superseded codegen fingerprint and rotation, with the accompanying `manifest.py` and `verification.py` changes.

`test_sync_core_pdd_rollout_policy.py` and `test_sync_core_verification_profiles.py` are being re-run against this state; I'll follow up with the result rather than assert it here.

### Not in scope

`.github/workflows/unit-tests.yml` was reverted to `main` in `944ab6d04` — the manual-dispatch trigger and the 60→360 minute timeout belong in their own change.

gltanaka (changes requested):
Tier A, full-diff correctness review at `00fb6e86d5d88abb17e4477980612d3cc7f38ad0`.

The split is needed and the checked-in extraction is largely sound: all 93 definitions removed from the monolith resolve in exactly one new module, the moved implementations have no substantive AST drift, all six prompt/code/architecture pairs are present, and the focused gate suite is green. Two merge-blocking boundaries are not covered by the current tests or contributor E2E, however.

1. **P1 — preserve append-only verification history** (`.pdd/verification-profile-rotations.json:408,688`).

   The candidate inserts a rotation inside the protected `requirement_rotations` prefix and mutates an existing protected retirement replacement. The production loader explicitly requires both candidate arrays to retain the protected arrays as prefixes (`pdd/sync_core/verification.py:2764-2770`). Reproducing the real merge transition with:

   ```sh
   /opt/homebrew/Caskroom/miniforge/base/envs/pdd/bin/python -c "from pathlib import Path; from pdd.sync_core import build_unit_manifest, load_verification_profiles; m=build_unit_manifest(Path.cwd(), base_ref='origin/main', head_ref='origin/pr/2374'); print(m.invalid_reasons); load_verification_profiles(Path.cwd(), m)"
   ```

   builds a clean manifest (`()`) and then deterministically raises `VerificationProfileError: candidate retirement history is not append-only`. The current rollout tests exercise `HEAD`/`HEAD` or frozen historical pairs, so CI remains green while missing the actual protected-base-to-candidate transition.

   Keep every `origin/main` rotation and retirement unchanged as the candidate prefix, append the new authorized history using the staged/bootstrap mechanism, and add a regression that calls `load_verification_profiles` for the actual merge base and candidate head.

2. **P1 — make the compatibility re-exports regenerable** (`pdd/prompts/code_generator_main_python.prompt:115`; required implementation is `pdd/code_generator_main.py:60-164`).

   R10 and the deliverables require the redundant-alias compatibility surface, but the prompt's only self-include selects eleven function definitions and none of the six import stanzas. Its `<pdd-interface>`/`architecture.json` entry declares only the five orchestration functions. Selecting that self-include yields no `from .conformance...` block and no `_is_test_output_path`, `_snapshot_public_surface`, or `_verify_public_surface_regression` alias. Those aliases are still imported by production callers including `one_session_sync.py`, `agentic_test_generate.py`, and `server/routes/prompts.py`, while the declared conformance gates cannot catch their loss.

   Include the exact compatibility block in regenerating context, make `validate_conformance_prompts.py` enforce the complete required alias set (including underscore helpers), and add a self-generation/context regression proving all aliases survive before importing the real callers. Without this, the next full sync can silently undo the compatibility portion of the split—the central failure mode this PR is intended to prevent.

Validation performed independently:

- `python scripts/validate_conformance_prompts.py` — PASS.
- 624 focused generator/gate/history/story tests — PASS in the `pdd` Conda interpreter.
- Additional broader run reached 1,251 passed / 1 skipped before interruption in an incorrectly inherited virtualenv; I did not count it as authoritative.
- Direct protected-base profile load — FAIL as above.
- Prompt self-include extraction — confirms all required compatibility aliases are absent.
- `git diff --check`, compile, and error-only lint checks — PASS; worktree clean.

The contributor's manual write-blocking/override E2E is useful evidence for the current checked-in gate behavior, but it does not exercise either the actual merge transition or regeneration of the orchestrator compatibility surface, so it is not sufficient for this PR as submitted.

Non-blocking cleanup while updating artifacts: the `architecture.json` descriptions for `gate_errors`, `directives`, and `test_churn` still say 52–53% and above the guide's band, contradicting the current approximately 25–30% ratios and the PR description.

gltanaka (changes requested):
Review of updated head faee09f069b16b5268b778f7263e584903d5c8d6.\n\nThe two blockers from my prior review are resolved: the real origin/main -> HEAD profile transition now loads cleanly at 100% coverage, and the code-generator prompt now carries and validates all compatibility re-exports. The architecture ratio descriptions are also corrected.\n\nOne substantive blocker remains:\n\n[P1] Re-pin the sync-ownership policy digest after adding the compatibility-export test ownership row.\n\n.pdd/sync-ownership.json now hashes to 2b4cee2ca317ff8283bc4d681dbc75b7b73cc704b3f62c1e122c20c1a7da1e38, but pdd/sync_core/manifest.py:339-342 still pins 048e43ef59bf5402929f354c9bf8cefc4fe6235976cefa65246198e3eea115e2. Because _sync_rollout_repair_ownership_rules requires that exact digest, the mismatch disables the repair bridge and eight historical .pdd/meta/* paths lose their effective ownership. This is reproduced by:\n- tests/test_sync_core_pdd_rollout_policy.py::test_sync_rollout_repair_metadata_bridge_stays_ordinary\n- tests/test_sync_core_pdd_rollout_policy.py::test_sync_rollout_repair_ownership_pin_tracks_the_actual_policy_file\n\nThe required Unit Tests workflow is therefore red: 7 failed, 16,386 passed, 35 skipped, 1 xfailed. The other five failures are stale diagnostic-regex expectations in test_estimate_contract_rotations_reject_substitution. All five adversarial cases still fail closed with VerificationProfileError: candidate requirement transition retirement rules are duplicated or chained. Please update the expected diagnostic without weakening the rejection behavior.\n\nValidated locally:\n- origin/main -> HEAD manifest/profile load: 475 profiles, no invalid reasons, no unaccounted paths, coverage 1.0\n- scripts/validate_conformance_prompts.py: passes, 6 prompts and architecture consistent\n- focused remediation tests: 2 passed, 2 failed on the ownership digest/bridge\n- all eight substitution cases: 3 passed, 5 failed only because the expected regex omits the new fail-closed diagnostic\n\nPlease re-pin the ownership digest, reconcile the diagnostic expectations, and rerun the complete Unit Tests workflow before re-review.

gltanaka (approved):
Re-review of head 3b0f214b88cf614ae1ca5f27e24da215c8e06bed: approved.\n\nThe update resolves the remaining ownership-digest blocker and scopes the retirement-chain exception to the reviewed code-generator successor without widening candidate authority.\n\nValidated:\n- Fresh independent Tier A delta review: MERGE — no findings\n- Real origin/main -> HEAD manifest/profile load: 475 units and profiles, no invalid reasons, no unaccounted paths, coverage 1.0\n- tests/test_sync_core_verification_profiles.py: 81 passed\n- 11 focused adversarial and merge-transition tests: passed\n- scripts/validate_conformance_prompts.py: 6 prompts and architecture consistent\n- Required GitHub checks: all passing, including full Unit Tests, public CLI regression, story regression, package preprocess smoke, and repo-bloat E2E\n- Clean worktree and unchanged reviewed head\n\nThe contributor end-to-end gate evidence plus deterministic prompt, history, ownership, and CI coverage is sufficient for this extraction.

PR #2376, opened 2026-08-06
fix: pdd-change Step 6 should use <pdd-dependency> tags, not <include> tags

### LINKED ISSUE #1807: Make prompt files properly use <pdd-dependency> tags and <include> tags so we can trust our architecture.json

agarwal-ishaan:
## Summary

`pdd change` Step 6 built its module-dependency context by scanning `<include>` tags (`build_dependency_graph()`), but `<include>` tags are LLM context, not declared architectural dependencies. A module can truly depend on another via `<pdd-dependency>` without ever `<include>`-ing it, and can `<include>` another module's example purely for style without depending on it at all -- so the old graph both missed real dependents and fabricated false ones.

1. **Prompts** (`pdd/prompts/sync_order_python.prompt`, `pdd/prompts/agentic_change_orchestrator_python.prompt`): specify a new `build_dependency_graph_from_architecture()` function and wire Step 6's dependency context to use it.
2. **`pdd/sync_order.py`**: implements `build_dependency_graph_from_architecture(architecture_path) -> Dict[str, List[str]]`, reading `architecture.json`'s `dependencies` field (populated from `<pdd-dependency>` tags by `architecture_sync.py`) instead of scanning `<include>` tags. Generated via `pdd generate --incremental`.
3. **`pdd/agentic_change_orchestrator.py`**: `_build_dependency_context()` now takes `architecture_path` and calls the function above instead of the include-based `build_dependency_graph(prompts_dir)` -- the actual Step 6 behavior fix.

## Test plan
- [x] `pytest tests/test_sync_order.py` -- 52 passed
- [x] `pytest tests/test_agentic_change_orchestrator.py` -- 265 passed
- [x] Manual repro (`pdd-issue-1807-repro/`): `_build_dependency_context()` now correctly shows `payment_status -> affects: customer_service` (previously missing) and no longer fabricates `legacy_wizard -> affects: onboarding`
- [x] Confirmed no regression: sibling modules (`agentic_sync`, `sync_graph_order_consistency`) that also import from `sync_order` still import cleanly

Fixes #1807


gltanaka (approved):

gltanaka (changes requested):
Thanks for consolidating the earlier prompt and implementation PRs. The architecture-based direction is sensible, but this version has two merge-blocking correctness problems:

1. `architecture.json.dependencies` is not currently a `<pdd-dependency>`-only data source. `merge_auto_deps_includes_into_architecture()` intentionally writes module/example `<include>` targets into the same array, and existing tests require that behavior. Consequently, reading this field cannot prove that an include-only module edge is excluded. The new Step 6 test hand-writes JSON without the include edge, so it does not exercise the real prompt → architecture → Step 6 path.

2. `build_dependency_graph_from_architecture()` uses `extract_module_from_include()` for architecture identities. That helper drops directories and rejects `_LLM.prompt` files. This loses real declared dependencies in the current repository. For example, `commands/gate_python.prompt -> gate_python.prompt` collapses to `gate -> gate` and is discarded as a self-edge; dependencies such as `agentic_update_python.prompt -> agentic_update_LLM.prompt` are also omitted.

Please resolve the dependency provenance/semantics and use a path-preserving architecture identity that supports LLM modules. Add regressions for:

- a real prompt-to-architecture-to-Step-6 flow distinguishing `<pdd-dependency>` from context-only `<include>`;
- path-qualified modules with the same basename;
- `_LLM.prompt` dependencies;
- root plus nested architecture files.

The focused suites pass (317 tests), but they cover only flat `_python.prompt` names and do not exercise these production shapes. Follow-up scope should also address the issue's requested prompting-guide examples and the retained top-30/top-10 truncation.

agarwal-ishaan:
Implemented the above changes, requesting you to please review

gltanaka (changes requested):
Re-review of current head `8c4338c9` against `origin/main` (`f25a0139`), including the complete PR/issue history, superseded PRs #2373/#2375, all commits, the cumulative diff, `architecture.json`, prompt sources, prompting guide, and the PR-loop runbook. There are no inline review threads (0 resolved / 0 unresolved). The core provenance correction is directionally right, but this is not merge-ready.

## Blocking findings

1. **P1 — the verification-profile transition fails on the actual PR base.**

   `pdd/sync_core/verification.py:2782` validates the protected active requirement rotations before it can apply the candidate retirement added in `.pdd/verification-profile-rotations.json:676`. On this exact head, the production path
   `load_verification_profiles(build_unit_manifest(root, base_ref="origin/main", head_ref="HEAD"))`
   raises `VerificationProfileError: protected active requirement transition rules are ambiguous` because protected `origin/main` has two active `code_generator_main_python.prompt/python` rows. The current hash-only reconciliation test does not exercise the real base-to-head load.

   Please either land/rebase onto the prerequisite protected retirement and remove the duplicate candidate retirement, or add a narrowly byte-bound bridge that can validate this exact transition before protected-row uniqueness is enforced. Add the exact `origin/main -> PR head` production-loader regression.

2. **P1 — Step 6 still drops most real dependency context.**

   `pdd/agentic_change_orchestrator.py:2284-2290` retains the old `[:30]` module and `[:10]` dependent slices, and `pdd/prompts/agentic_change_orchestrator_python.prompt:329` explicitly requires them. Against this repository's real architecture-derived graph, 135 dependency targets have dependents, but only 30 are rendered: 105 targets and 8 additional dependent names disappear. In particular, the newly preserved `agentic_update -> agentic_update_LLM` edge exists in the graph, but `agentic_update_LLM` ranks 68th and is absent from the actual Step 6 context. Thus the prior `_LLM` omission remains at the user-visible boundary.

   Render all edges in a compact deterministic representation (or another complete representation with explicit fidelity guarantees) and add >30-target and >10-dependent regressions at `_build_dependency_context`, not only at the graph helper.

3. **P1 — nested architecture support is only partially implemented and the requested real flow is still untested.**

   `pdd/sync_order.py:244,261` assumes prompt files live under the architecture file's own `prompts/` or `pdd/prompts/`. That contradicts the established contract in `pdd/architecture_sync.py:118-138` / its tests: a nested `architecture.json` may use an ancestor project's `prompts/` directory. A deterministic reproduction with `repo/prompts/auth_python.prompt` and `repo/backend/functions/architecture.json` returns `{'auth': [], 'log': []}` and warnings instead of `auth -> log`. The valid root entry `context/python_preamble.prompt` is likewise misresolved as `pdd/prompts/context/python_preamble.prompt`.

   The advertised “real prompt -> architecture” tests do not cover this boundary: `tests/test_sync_order.py:827-850` says it does not hand-write architecture JSON, but lines 842-848 do exactly that; `tests/test_agentic_change_orchestrator.py:10784` does the same. Neither runs architecture sync/auto-deps, where include-derived edges make `architecture.json.dependencies` mixed-provenance.

   Resolve prompts through the governing/nearest supported prompt root (and safe architecture-relative paths), then add a deterministic production-flow regression that creates architecture state through sync/auto-deps and invokes Step 6. Cover root and nested architectures, including the ancestor-prompts shape.

4. **P2 — the ancillary include-parser fix still treats fenced examples as directives.**

   `pdd/sync_core/includes.py:211` and `:264` skip only a tag whose immediately preceding character is a backtick. `<include>` and `<include-many>` inside triple-backtick or `~~~` fenced examples are still returned as real references, contrary to `docs/prompting_guide.md:1404`; this can make closure validation/fingerprinting require example-only files. The new tests cover inline code only.

   Exclude both Markdown fence styles while preserving the supported legacy `````<path>````` include grammar, and add fenced XML/include-many regressions.

5. **P2 — the linked issue's prompting-guide acceptance work remains absent.**

   #1807 explicitly asks for practical examples explaining when a real module relationship uses `<pdd-dependency>` and when files are only LLM context via `<include>`. This PR does not change `docs/prompting_guide.md`. Its current line 1645 also says `<include>` “does NOT affect architecture,” while the implementation and this PR correctly acknowledge that auto-deps can write include-derived edges into the architecture `dependencies` array.

   Add the requested customer-service/payment-status versus refund-policy/tone-context examples and reconcile that mixed-field behavior in the guide. If that work is intentionally deferred, this PR should not close #1807; link a narrower issue instead.

## Prior-request checklist

| Prior request | Implementing evidence | Current status |
|---|---|---|
| Stop Step 6 using the include graph | `4b6349dfa` | satisfied |
| Do not trust mixed `architecture.json.dependencies` as pdd-dependency-only | `626002541`, fresh prompt-tag parsing | satisfied |
| Preserve path-qualified same-basename identities | `626002541`; `tests/test_sync_order.py:877-895` | satisfied |
| Preserve `_LLM.prompt` dependencies | `626002541`; graph test at `:898-914` | **partially satisfied** — graph keeps it, Step 6 truncates it |
| Root + nested architecture behavior | `626002541`; local-prompts-only test at `:917-935` | **partially satisfied** — supported ancestor prompt roots fail |
| Real prompt -> architecture -> Step 6 regression | claimed by `626002541` | **not addressed** — tests hand-write architecture JSON |
| Prompting-guide distinction and examples | no implementing commit/diff | **not addressed** |
| Remove top-30/top-10 loss | no implementing commit; slices remain | **not addressed** |
| Verification-profile reconciliation | `849fdd5ac`, `460998285`, `7abac24eb` | **regressed/blocking** — production base-to-head load fails |
| Ignore documentation examples in include parsing | `8c4338c97` | **partially satisfied** — inline works, fenced examples regress |
| Initial direct architecture-field design / first restamp | superseded by `626002541`; `0181bd446` reverted by `315bd833` | obsolete with valid supersession |

Validation evidence: 323 focused orchestrator/sync-order tests passed; 21 architecture/auto-deps boundary tests passed; GitHub's Unit Tests, Public CLI Regression, Story Regression, Package Preprocess Smoke, Repo Bloat Docker E2E, and heal checks are green. Those green checks do not cover the failing production profile load or the deterministic Step 6/nested/fence reproductions above.

gltanaka (changes requested):
Re-review of current head `2c72c7a18` against `origin/main` (`0faf7be4c`), including all commits/comments/reviews/checks, linked issue #1807/project state, the prior requested-change ledger, source prompts, generated code, `architecture.json`, verification profiles, the PR-loop runbook, and the prompting guide.

The core dependency-graph change now works end to end: production profile loading is clean, Step 6 reads declared `<pdd-dependency>` edges, nested/path-qualified/`_LLM` identities survive, and the top-30/top-10 truncation is gone. Two review blockers remain.

## Blocking findings

1. **P2 — Markdown literal filtering is still incomplete and can invent or suppress include dependencies.**

   `pdd/sync_core/includes.py:199-207` accepts mixed backtick/tilde fence runs and requires an exact-length closing run; `:230-232` and `:286-288` ignore inline markup only when the tag is immediately preceded by a backtick. This disagrees with the prompting-guide contract that directives inside fenced or inline code are literal.

   Deterministic reproductions against this head:

   - `Use \`example <include>docs/fake.md</include>\` here` returns both `fake.md` and the later real include.
   - a ```` fence closed by four backticks, or `~~~` closed by four tildes, leaks fenced `<include>` / `<include-many>` paths.
   - an invalid mixed `\`~~` fence suppresses a real include entirely.

   Because this canonical parser feeds include closure/fingerprinting and sync-order discovery, false positives can require example-only files while false negatives omit intended context. Please use a linear Markdown span scanner that requires homogeneous fence characters, accepts same-character closers at least as long as the opener, recognizes complete variable-length inline-code spans, and preserves the supported legacy ```<path>``` grammar. Add `<include>` and `<include-many>` regressions for these cases.

2. **P2 — the requested production-flow regression still hand-writes the integration boundary.**

   `tests/test_sync_order.py:827-850` says it exercises a real prompt → architecture → graph flow, but `:842-848` directly writes `architecture.json`. `tests/test_agentic_change_orchestrator.py:10768-10803` makes the same production-flow claim and directly writes the registry at `:10791-10797`.

   My real root flow through `sync_all_prompts_to_architecture()` → `merge_auto_deps_includes_into_architecture()` → `_build_dependency_context()` passes, as does a nested architecture using ancestor prompts. That confirms the implementation direction, but CI still does not protect the exact mixed-provenance boundary that caused this bug and that the prior review explicitly requested. Please add deterministic root and nested tests that create architecture state through production sync/auto-deps, invoke Step 6, and assert declared edges appear while include-only edges do not.

## Prior-request checklist

| Prior request | Current status |
|---|---|
| Stop Step 6 using the include graph | resolved |
| Do not trust mixed `architecture.json.dependencies` as provenance-pure | resolved |
| Preserve path-qualified same-basename identities | resolved |
| Preserve `_LLM.prompt` identities through Step 6 | resolved |
| Remove top-30/top-10 loss | resolved |
| Root/nested architecture and ancestor prompt roots | resolved in implementation and manual E2E |
| Real sync/auto-deps → architecture → Step 6 regression | **partially resolved** — behavior passes manually; checked-in tests still hand-write JSON |
| Ignore Markdown documentation examples | **partially resolved** — exact fences pass; inline/valid longer closers/mixed-delimiter cases remain wrong |
| Prompting-guide distinction/examples and mixed-field wording | resolved |
| Production `origin/main → HEAD` verification-profile load | resolved: 469 managed units, 0 manifest/profile errors, 100% coverage |
| Initial direct-field design / superseded restamps | obsolete with valid supersession |

## Validation

- `pytest -q tests/test_sync_order.py tests/test_agentic_change_orchestrator.py tests/test_sync_core_includes.py` — **353 passed**
- `pytest -q tests/test_architecture_sync.py tests/test_auto_deps_architecture.py tests/test_architecture_include_validation.py` — **215 passed**
- two focused rollout/profile reconciliation tests — **2 passed**
- exact production manifest/profile load — **469/469 profiles, zero errors, coverage 1.0**
- real root sync+auto-deps+Step-6 fixture — **82 nodes synced; mixed include edge excluded; 45 declared edges retained**
- nested ancestor-prompts architecture → Step 6 fixture — **pass**
- real repository graph — **328 nodes / 306 edges; `commands/gate -> gate` and `agentic_update -> agentic_update_LLM` visible**
- all GitHub Unit Tests, CLI/story/package/Docker/heal checks — **green**

The contributor branch was not modified. The linked issue remains open and in Changes Requested.

gltanaka:
Rechecked current head `7453ccd8b`. The new `sync_core.includes` mask fixes the originally reported canonical-parser cases, but this is still not ready to merge because the user-visible preprocessing path remains inconsistent.

## Required fixes

1. **Use the corrected Markdown literal scanner in actual prompt expansion.**

   `pdd/preprocess.py:183-197` still has the old exact-delimiter/mixed-delimiter fence regex. That scanner is used by XML include expansion at `:1079-1084` and include-many expansion at `:1391-1396`.

   Deterministic production failures with real files present:

   - ```xml` opened with three backticks and closed with four expands a literal `<include>docs/fake.md</include>`.
   - `~~~` opened with three tildes and closed with four expands a literal `<include-many>`.
   - an invalid mixed `~~\`` fence suppresses a real include.

   In other words, canonical dependency discovery now ignores these literals, while actual preprocessing still expands or suppresses them. Please share one offset-preserving Markdown span implementation between `sync_core.includes` and `preprocess` (and preferably include validation), rather than maintaining two scanners.

2. **Cover remaining canonical-parser edge cases.**

   The current replacement mask still fails these inputs:

   - a valid fenced info string beginning with `<`, such as ```<xml>`, leaks the fenced include and can hide the following real include;
   - a valid inline span containing a non-closing longer backtick run leaks the literal include;
   - a valid multiline Markdown code span leaks its literal include;
   - two legacy ```<path>``` includes on one line collapse into one malformed combined path.

   Protect complete legacy tokens individually, then scan ordinary Markdown fences/code spans according to delimiter-run rules. Add both `<include>` and `<include-many>` regressions.

3. **Add the previously requested production-flow regression.**

   The PR still has no checked-in root/nested test for:

   `sync_all_prompts_to_architecture → merge_auto_deps_includes_into_architecture → _build_dependency_context`

   Existing graph/Step-6 tests still hand-write `architecture.json`. The manual root and nested flows pass, so this is a regression-protection gap rather than evidence that the core graph is currently broken.

## Current evidence

- New parser suite: 30 passed.
- Sync-order + orchestrator suites: 326 passed.
- Manual root and nested dependency flows: passed.
- GitHub checks: green.
- Actual `preprocess()` reproductions above: fail on this head.

The contributor branch was not modified.

gltanaka (changes requested):
Re-review of current head `62016e501`.

The Markdown scanner/preprocessor fixes and the production dependency-graph regression are now in good shape, but one P1 blocker remains after merging PR #2374.

## P1 — production verification-profile authorization is still bound to the pre-conformance state

The latest commit correctly re-pins the candidate profile digest to `85d010…`, but `pdd/sync_core/verification.py:238-245` still defines the PR #2376 transition from the older pre-#2374 state:

- configured rotation policy: `57d51b… → 57d51b…`
- actual `origin/main → HEAD` rotation policy: `470a4e… → 470a4e…`
- configured profile: `0fcb1b… → 85d010…`
- actual `origin/main → HEAD` profile: `faf427… → 85d010…`

Because that pair does not match, `pr2376_dependency_fix_state` at `pdd/sync_core/verification.py:3225` is false and the three dependency-fix requirement transitions at `:1058` are never applied.

Exact repository regression:

`conda run -n pdd pytest -q tests/test_sync_core_pdd_rollout_policy.py::test_conformance_split_profiles_load_from_actual_merge_base`

Result: **failed** at `tests/test_sync_core_pdd_rollout_policy.py:2251`. The production manifest is clean with 475 managed units and zero unaccounted paths, but profile loading returns 19 invalid reasons, 99.37% coverage, and incomplete profiles for `agentic_change_orchestrator`, `architecture_include_validation`, and `sync_order`.

Please bind the PR #2376 authorization to the exact post-#2374 base/current policy-profile pair: the conformance stationary rotation policy as the base and head policy, and `faf427… → 85d010…` for profiles. Keep the authority byte-exact, and make the existing `build_unit_manifest(origin/main, HEAD) → load_verification_profiles()` regression pass with zero invalid reasons and 100% coverage.

Other re-review evidence is green: 236 parser/preprocessor/validation tests, 328 graph/orchestrator tests, and 388 conformance/story tests passed. The contributor branch was not modified.

gltanaka (approved):
Approved at `a8a557d63` after the final verification-policy correction.

The PR now solves issue #1807 end to end: Step 6 derives complete declared `<pdd-dependency>` context from architecture-backed prompt identities without trusting mixed-provenance dependency arrays; path-qualified and `_LLM` identities are preserved; truncation is removed; root/nested production flows and Markdown-literal include behavior are covered.

Final validation:
- exact `origin/main → HEAD` production profile load: 475/475 profiles, zero manifest/profile errors, zero unaccounted paths, coverage 1.0
- exact production regression: passed
- focused historical-policy regressions: 2 passed
- sync-order suite: 62 passed
- parser/preprocessor/validation, graph/orchestrator, and conformance/story suites from the current functional diff: 236, 328, and 388 passed
- independent critical review: no P1/P2 findings; byte-exact authority, no trust widening

Fresh CI is green except Unit Tests, which is still running. Merge is recommended once that required check passes.

PR #2378, opened 2026-08-06
fix: support global options in Zsh completion

### LINKED ISSUE #2377: fix: Zsh completion fails when global options precede a subcommand
`pdd`’s Zsh completion dispatcher assumes the subcommand is always the second shell token. As a result, global options placed before the command break command-specific completion:

```zsh
pdd --local sync <Tab>
```

The dispatcher reads `--local` as though it were the subcommand, so it falls back to generic completion instead of invoking `_pdd_sync`. Normal completion such as `pdd sync <Tab>` continues to work.

Expected behavior: Zsh completion should find the first known PDD subcommand after skipping global flags and values for global options that take arguments. Both forms should provide sync-specific suggestions:

```zsh
pdd sync <Tab>
pdd --local sync <Tab>
```

The fix updates `prompts/pdd_completion_zsh.prompt`, regenerates/updates `pdd/pdd_completion.zsh`, and adds regression coverage for `--local` and `--context local` before `sync`.

agarwal-ishaan:
Fixes #2377

## Summary
- resolve the first known PDD subcommand after any preceding global options
- preserve normal `pdd sync` completion while supporting `pdd --local sync`
- update the Zsh completion prompt contract and add regression coverage

## Validation
- `conda run -n pdd python -m pytest tests/test_zsh_completion.py -q`
- `zsh -n pdd/pdd_completion.zsh`

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -129,7 +129,7 @@ jobs:
       - name: Provision and verify protected Linux sandbox
         run: |
           sudo apt-get update
-          sudo apt-get install --yes bubblewrap
+          sudo apt-get install --yes bubblewrap zsh
           private_root="$(mktemp -d)"
           chmod 700 "$private_root"
           mkdir -m 700 "$private_root/scratch"
--- a/.pdd/sync-ownership.json
+++ b/.pdd/sync-ownership.json
@@ -14916,6 +14916,12 @@
       "owner": "pdd-maintainers",
       "pattern": "pdd/story_detection_result.py",
       "role": "human-maintained"
+    },
+    {
+      "inventory": "HUMAN_OWNED",
+      "owner": "pdd-maintainers",
+      "pattern": "tests/test_zsh_completion.py",
+      "role": "human-maintained"
     }
   ]
 }
--- a/.pdd/verification-profiles.json
+++ b/.pdd/verification-profiles.json
@@ -8279,7 +8279,7 @@
       "prompt_path": "pdd/prompts/pdd_completion_zsh.prompt",
       "language_id": "zsh",
       "required_requirement_ids": [
-        "CONTRACT-SHA256:5d05ee3fcc959bfba1312b3d2ded6c2329ef1f6df0f532e855ac5a345ec6a107"
+        "CONTRACT-SHA256:488cde2f0d55b75a1cafc4f3e8f0e5f3984d4c0c0b70076c0d61820a4974365f"
       ],
       "obligations": [
         {
@@ -8288,7 +8288,7 @@
           "validator_id": "threshold-ed25519",
           "validator_config_digest": "threshold-ed25519-v1",
           "requirement_ids": [
-            "CONTRACT-SHA256:5d05ee3fcc959bfba1312b3d2ded6c2329ef1f6df0f532e855ac5a345ec6a107"
+            "CONTRACT-SHA256:488cde2f0d55b75a1cafc4f3e8f0e5f3984d4c0c0b70076c0d61820a4974365f"
           ],
           "artifact_paths": [
             "pdd/prompts/pdd_completion_zsh.prompt"
--- a/pdd/pdd_completion.zsh
+++ b/pdd/pdd_completion.zsh
@@ -50,6 +50,7 @@ local -a _pdd_global_opts
 _pdd_global_opts=(
   '--force[Overwrite existing files without asking for confirmation.]'
   '--strength[Set the strength of the AI model (0.0 to 1.0, default: 0.5)]:strength:(0.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0)'
+  '--model[Override the model for this invocation]:model:_guard'
   '--time[Controls the reasoning allocation for LLM models (0.0 to 1.0, default: 0.25)]:time:(0.0 0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0)'
   '--temperature[Set the temperature of the AI model (default: 0.0)]:temperature:(0.0 0.25 0.5 0.75 1.0)'
   '--verbose[Increase output verbosity for more detailed information.]'
@@ -62,6 +63,9 @@ _pdd_global_opts=(
   '--review-examples[Review and optionally exclude few-shot examples before command execution.]'
   '--local[Run commands locally instead of in the cloud.]'
   '--context[Override automatic .pddrc context]:context-name:_guard'
+  '--keep-core-dumps[Number of core dumps to retain]:count:'
+  '--context-compression[Set global context compression mode]:mode:(off test examples contracts all)'
+  '--compression-fallback[Set behavior when compression fails]:mode:(full error)'
   '--list-contexts[List available .pddrc contexts and exit]'
   '--core-dump[Write a JSON core dump.]'
   '--no-core-dump[Disable JSON core dumps.]'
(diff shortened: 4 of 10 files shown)


gltanaka (approved):

agarwal-ishaan:
PR can be merged, all test have passed, will work on creating manifest.py prompt and verification.py prompt in a seperate issue, they are marked as human owned and maintained, moving to prompt based seems practical and the correct next step.

gltanaka (changes requested):
Changes requested after reviewing the full cumulative diff at `8c9ab2bc2`.

The change is needed and the primary case is fixed: native Zsh completion for `pdd --local sync --skip-<Tab>` now offers the sync options. The branch is not merge-ready because the implementation does not preserve the same behavior across the rest of the CLI grammar, and the green tests stub out the completion state that exposes the failures.

- **P1 — Global-option dispatch breaks nested/group completion** (`pdd/pdd_completion.zsh:488`, `:588-663`; masked by `tests/test_zsh_completion.py:34-55`). `_pdd_checkup` still assumes fixed absolute positions (`words[3]`), so `pdd --local checkup gate ...` no longer reaches `_pdd_checkup_gate`. The six new group functions call `_arguments` without shifting past the already-resolved root command, so both `pdd auth <Tab>` and `pdd --local auth <Tab>` offer none of `login/status/...`. The test helper forces `state=group_command`, so its assertions pass without exercising Zsh's real positional parsing. Resolve and shift `words`/`CURRENT` relative to the located command (including nested commands), and add a native completion-context regression covering ordinary and global-option-prefixed forms.

- **P2 — The scanner searches past invalid command tokens** (`pdd/pdd_completion.zsh:545-582`; `tests/test_zsh_completion.py:120-128`). Click treats the first unconsumed positional token as the command. This scanner ignores an unknown bare token and later dispatches a known one: `pdd -- typo sync --<Tab>` offers sync options, while `python -m pdd -- typo sync --help` exits 2 with `No such command 'typo'`. Stop at the first non-global positional token; after `--`, accept only the immediately following token as the command. Cover both `pdd typo sync` and `pdd -- typo sync`.

- **P2 — The shipped global option specification remains incomplete** (`pdd/pdd_completion.zsh:50-72`, compared with `pdd/core/cli.py:530-658`). The new scanner recognizes `--model`, `--keep-core-dumps`, `--context-compression`, and `--compression-fallback`, but `_pdd_global_opts` does not expose them. Native `pdd --mo<Tab>` and `pdd --context-com<Tab>` produce no matching option/value completion. Add all four with value completers matching their Click types, and test actual suggestions rather than only `_pdd_find_subcommand` output.

- **P2 — The authoritative prompt still omits this behavior** (`pdd/prompts/pdd_completion_zsh.prompt:16-28`; mapped to the artifact by `architecture.json:4333-4345`). Issue #2377 and the PR summary require the prompt update, but commit `aaa47e215` reverted it. That leaves the generated artifact materially ahead of its source contract and allows a future regeneration to restore the bug. Follow the protected Phase A/Phase B rotation in `docs/ci.md`, then consume the prompt/profile update here (or make the protected prerequisite explicit before this is merged).

Validation: `tests/test_zsh_completion.py` 47 passed; `zsh -n` passed; the three `sync_rollout_repair` tests passed; compile/lint checks for the touched test and manifest delta passed; native Zsh reproduced the passing sync case and the failures above; current GitHub checks are green. I found no substantive security or authorization widening in the two exact manifest ownership digest pairs. The existing approval was on pre-force-push commit `70d07334e`, not this head.

gltanaka (changes requested):
The remediation at `c41f930c` fixes three major parts of the previous review: invalid positional tokens now stop dispatch, the missing global option specs are present, and ordinary/`--local` group and nested completion are exercised through a real PTY-backed Zsh completion frame. This head is much closer, but two blocking findings remain.

- **P1 — Nested dispatch still confuses a global option value with the command** (`pdd/pdd_completion.zsh:493-502`; missing from `tests/test_zsh_completion.py:200-202`). `_pdd_checkup` scans for the first literal `checkup` instead of using the command position already resolved by `_pdd_find_subcommand`. Therefore `pdd --model checkup checkup gate --po<Tab>` and `pdd --context checkup checkup gate --po<Tab>` produce no `--policy` completion, while Click accepts these arbitrary option values and `python -m pdd --model checkup checkup gate --help` exits 0 with the gate help. Return/reuse the resolved command index (or parse while skipping global values) and add a native regression using a value-taking global option whose value equals the command.

- **P2 — The authoritative prompt contract is still stale** (`pdd/prompts/pdd_completion_zsh.prompt:16-28`; `architecture.json:4333-4345`). The remediation does not touch the prompt, verification profile/rotation, or make a protected prerequisite explicit. Commit `aaa47e215` still leaves the generated artifact ahead of its source contract, while issue #2377 and the PR body say this PR updates that contract. A later regeneration can restore the original defect. Land the required protected Phase A/Phase B prompt/profile transition before consuming it here, or explicitly reclassify/document the completion artifact as human-owned and correct the stated acceptance criterion.

Validation on this head:

- `tests/test_zsh_completion.py`: 58 passed twice using the newly declared `pexpect` dependency installed into a temporary directory.
- Three `sync_rollout_repair` tests: passed.
- `zsh -n`, compile, targeted pylint, and `git diff --check`: passed.
- Native Zsh confirmed the fixed sync/group/option cases and reproduced the value-collision failure above.
- All current GitHub unit, public CLI, story, package, and repo-bloat checks are green.

The `pexpect>=4.9` dev dependency is justified and is installed by CI's existing `.[dev]` setup. The prescribed existing Conda environment does not yet contain it, but that is an environment refresh issue rather than a PR defect.

gltanaka (changes requested):
The functional Zsh remediation at `784ec666` is now correct: the resolver retains the actual command index, including when a preceding value-taking global option has the same text as the command, and the prompt text now describes the intended behavior. However, the new prompt/profile transition introduces one P1 trust-boundary blocker.

- **P1 — The candidate self-authorizes its protected contract transition** (`pdd/sync_core/verification.py:185-202`, `:1001-1013`, `:2980-3009`, `:3097-3108`; alongside `pdd/prompts/pdd_completion_zsh.prompt:29-35` and `.pdd/verification-profiles.json:8279-8296`). Commit `784ec666` changes the prompt and its required profile digest while simultaneously adding verifier recognition and transition authority for those exact candidate bytes. No dormant row was installed in `.pdd/verification-profile-rotations.json`, and no prior PR/commit contains this new digest. This is exactly the candidate self-authorization that `docs/ci.md:28-57` prevents: Phase A must install reviewed dormant authority and become protected before a later Phase B consumes the prompt/profile bytes; same-PR installation and consumption is explicitly forbidden. Repository binding and exact SHA pairs narrow the bypass but do not make the authority pre-existing or protected. Split this into a protected Phase A dormant rotation followed by a separate Phase B consumption, and remove the in-code bypass.

Green rollout tests cannot discharge this finding because they exercise the candidate-defined exception rather than prove prior protected authority.

Validation on this head:

- Native PTY-backed Zsh completion suite: 60 passed.
- Native checks confirm `pdd --model checkup checkup gate --po<Tab>`, `--context` and `--option=value` variants now complete `--policy`; sync/group/invalid-token cases remain correct.
- Prompt and profile SHA-256 values independently match the new constants.
- `zsh -n`, compile, targeted pylint, and `git diff --check`: passed.
- Current GitHub unit, public CLI, story, package, and repo-bloat checks are green.

The functional fix is merge-worthy once its contract authority follows the repository's protected two-change process.

gltanaka (approved):
Approved at `784ec666` after explicit maintainer acceptance of the documented two-phase governance exception.

The functional change is correct: global options preceding commands now preserve command-specific and nested Zsh completion, including value-taking options whose value equals a command name. Native PTY-backed tests cover the original regression, group commands, invalid-token boundaries, missing global options, and the prior `checkup` collision.

Validation evidence:
- `tests/test_zsh_completion.py`: 60 passed
- native Zsh reproductions: passed
- `zsh -n`, compile, targeted pylint, and diff checks: passed
- current GitHub unit, public CLI, story, package, and repo-bloat checks: green
- protected profile/rollout suite made no failures in the 107 tests completed locally; the complete CI unit run is green

Accepted residual risk: the prompt/profile transition and exact repository-bound verifier recognition land together instead of through the documented separate Phase A/Phase B sequence. The maintainer explicitly authorized skipping that ceremony for this merge.

Issue #2379, opened 2026-08-07
pdd detect --stories --scope-manifest cannot validate a story whose prompt lives in a subdirectory


agarwal-ishaan:
## Summary

A user story whose linked prompt lives in a **subdirectory** of the prompts root can never be validated in `pdd detect --stories --scope-manifest` mode. It always fails closed with `scope:MANIFEST_MISMATCH` before any provider call.

This is self-inflicted: `pdd story link` / `pdd story add` write the story metadata in a form that the manifest pre-flight gate cannot resolve, and manifest mode rejects the flag that would otherwise point the resolver at the right directory. Every prompt under `prompts/<subdir>/` is affected, which today means `prompts/core/`, `prompts/commands/`, and any new package-style prompt folder.

Found while adding story coverage for a prompt split under `prompts/conformance/` (PR #2374).

## Reproduction

Self-contained, no network, no provider key:

```python
import json, os, tempfile
from pathlib import Path
from pdd.commands.analysis import _load_scope_manifest, _scope_manifest_metadata_matches

STORY = """<!-- pdd-story-prompts: {refs} -->

# User Story: demo

## Story
As a user, I can do the thing, so that the benefit follows.
"""

def build(root: Path, refs: str) -> Path:
    (root / "prompts" / "conformance").mkdir(parents=True)
    (root / "prompts" / "conformance" / "demo_python.prompt").write_text("% demo\n")
    (root / "user_stories" / "contracts").mkdir(parents=True)
    (root / "user_stories" / "story__demo.md").write_text(STORY.format(refs=refs))
    (root / "user_stories" / "contracts" / "demo.contract.md").write_text(
        "# Contract: demo\n\n## Covers\n- AC1: the thing\n")
    manifest = root / "scope.json"
    manifest.write_text(json.dumps({
        "schema_version": "pdd.detect.stories.scope.v1",
        "stories": [{
            "story": "user_stories/story__demo.md",
            "contract": "user_stories/contracts/demo.contract.md",
            "prompts": ["prompts/conformance/demo_python.prompt"],
        }],
    }))
    return manifest

for label, refs in [
    ("basename (what `pdd story link` writes)", "demo_python.prompt"),
    ("repo-relative path (manual edit)", "prompts/conformance/demo_python.prompt"),
]:
    with tempfile.TemporaryDirectory() as tmp:
        root = Path(tmp).resolve()
        manifest = build(root, refs)
        os.chdir(root)
[...]

PR #2380, opened 2026-08-07
Fix story prompt metadata paths for nested prompts

agarwal-ishaan:
## Summary

preserve logical prompts/subdirectory/file references when writing story metadata

- resolve both prompt and prompts-root paths so symlink aliases do not collapse nested paths to basenames
- add regression coverage for symlinked nested prompts and scope-manifest preflight

## Testing\n- uv run pytest tests/test_user_story_tests.py 

tests/commands/test_analysis.py::test_scope_manifest_preserves_exact_scope_and_rejects_discovery 
tests/commands/test_analysis.py::test_scope_manifest_accepts_nested_logical_prompt_metadata -q --tb=short

### DIFF
--- a/.pdd/verification-profiles.json
+++ b/.pdd/verification-profiles.json
@@ -10351,7 +10351,7 @@
       "prompt_path": "pdd/prompts/user_story_tests_python.prompt",
       "language_id": "python",
       "required_requirement_ids": [
-        "CONTRACT-SHA256:1c467034344d9d87b8225995bc458bc8093e6759dd5c2eed8424b345f69a3ba7"
+        "CONTRACT-SHA256:5b1353257a64a25b303d990803bb799da66504af558c3a5e972d95ad5a04bb3b"
       ],
       "obligations": [
         {
@@ -10360,7 +10360,7 @@
           "validator_id": "threshold-ed25519",
           "validator_config_digest": "threshold-ed25519-v1",
           "requirement_ids": [
-            "CONTRACT-SHA256:1c467034344d9d87b8225995bc458bc8093e6759dd5c2eed8424b345f69a3ba7"
+            "CONTRACT-SHA256:5b1353257a64a25b303d990803bb799da66504af558c3a5e972d95ad5a04bb3b"
           ],
           "artifact_paths": [
             "pdd/prompts/user_story_tests_python.prompt"
--- a/pdd/commands/analysis.py
+++ b/pdd/commands/analysis.py
@@ -22,6 +22,7 @@ from ..crash_main import crash_main
 from ..trace_main import trace_main
 from ..user_story_tests import (
     _parse_story_prompt_metadata,
+    _resolve_authorized_story_prompt_metadata_reference,
     discover_prompt_files,
     discover_story_files,
     run_user_story_tests,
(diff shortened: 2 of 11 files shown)


agarwal-ishaan:
Fixes Issue - https://github.com/promptdriven/pdd/issues/2379

gltanaka (approved):

gltanaka (changes requested):
Tier A, full-diff review of f5b8a3f53 — requesting changes for two P1 blockers.

1. **P1 — the default `pdd story link/add` path still writes lossy basenames.**

At `pdd/user_story_tests.py:534` and `:1493`, `prompts_root` is set only when `prompts_dir` is explicitly provided. But both `pdd story link` and `pdd story add` default `--prompts-dir` to `None`, so the new logic at `:316-325` is bypassed in the ordinary workflow from #2379.

I reproduced the exact default CLI boundary on this head:

```
pdd story link user_stories/story__demo.md \
  --prompt prompts/conformance/demo.prompt

link_exit=0
metadata=<!-- pdd-story-prompts: demo.prompt -->
manifest_matches=False
```

The scope-manifest preflight therefore still fails for the headline workflow. The new helper test supplies a prompts root directly, and the new analysis test prewrites already-correct metadata, so neither catches this integration gap.

Smallest safe fix: derive the default prompts root unconditionally in every metadata-writing path (including link/cache and add/generate), then add a CLI-level regression with omitted `--prompts-dir` that exercises writer -> scope-manifest preflight.

2. **P1 — protected verification state is rejected and required CI is red.**

The authorization appended at `pdd/sync_core/verification.py:1198-1208` is not reconciled with the protected transition history. GitHub's latest Unit Tests job reports 14 failures with `protected active requirement transition rules are ambiguous` (16,303 passed), including the standalone checker. The PR is also currently `CONFLICTING/DIRTY`.

Against current `origin/main`, both focused rollout checks fail with `requirement transition bindings mismatch`:

```
tests/test_sync_core_pdd_rollout_policy.py::test_detector_contract_rotation_is_exact_and_consumed
tests/test_sync_core_pdd_rollout_policy.py::test_rollout_profiles_cover_the_protected_pdd_denominator
2 failed
```

Please rebase onto current main, resolve the rollout-policy conflict, regenerate/reconcile the exact profile/rotation bindings using the established protected-history mechanism, and run the full sync-core verification suite plus standalone checker.

Validation performed:

- 193 story, story-CLI, analysis, and architecture-sync tests passed.
- Default CLI -> metadata -> manifest reproduction failed as shown above.
- Two focused protected rollout tests failed.
- `git diff --check` passed.
- `architecture.json` remains structurally aligned because no public interface or dependency changed.
- The prompt artifact is present, but the generated code does not satisfy its new requirement on the default CLI path.

The underlying #2379 problem is real and this PR is still needed, but the contributor's current E2E evidence is insufficient: it does not exercise the affected default writer-to-preflight boundary, and required CI is failing.

gltanaka (approved):
Re-review complete after the protected rollout and compatibility fixes.

Validated the complete diff and linked #2379 behavior. Nested prompt paths are preserved canonically; existing unique basename/suffix metadata is accepted only within the story's manifest-authorized prompt set and is canonicalized without an LLM call or cost. Ambiguous, unsafe, unmatched, and broader-filesystem candidates fail closed in both preflight and the structured-result builder. A second refresh is byte-for-byte unchanged.

Local evidence: 226 affected tests passed; 9 protected rollout/governance tests passed; architecture mapping remains correct; policy/profile/prompt hashes match the protected Phase A/Phase B values; no story-specific bootstrap transition was reintroduced. Independent critical delta review found no P1/P2 issues and reproduced canonicalization -> preflight PASS -> structured PASS.

Approve pending required CI.

gltanaka (approved):
Final delta review: the only first-run CI failure was a stale test expectation that paired an immutable historical base with moving HEAD. Commit 2946051a7 adds the exact protected Phase-B profile digest to that test allowlist only; production verifier code and authorization are unchanged. The failing target now passes, and the full affected rollout slice passes 10/10. Independent review confirmed this is the smallest safe correction. Approved pending replacement required CI.

PR #2386, opened 2026-08-10
fix(sync): retain rollout replay ownership pin

### LINKED ISSUE #2383: sync-rollout ownership pin cannot serve both the live head and the frozen replay ref
## Summary

`_SYNC_ROLLOUT_REPAIR_OWNERSHIP_BYTES` in `pdd/sync_core/manifest.py` is a single `(base, head)` digest pair, but `_sync_rollout_repair_ownership_rules` is exercised against **three different sources** for `head` — two that move with every edit, and one that is frozen forever:

| Test | reads `.pdd/sync-ownership.json` from | moves on edit? |
|---|---|---|
| `test_sync_rollout_repair_ownership_pin_tracks_the_actual_policy_file` | working tree | yes |
| `test_sync_rollout_repair_metadata_bridge_stays_ordinary` | `HEAD` | yes |
| `test_sync_rollout_repair_executes_the_actual_protected_transition` | `PR_2316_PHASE_A_PROTECTED` (`8ac79847`), hardcoded | **never** |

On `main` today all three hash to the same value, so one constant satisfies everyone:

```
sha256(dec539aa:.pdd/sync-ownership.json) = 8f5762a5…   # = _BYTES[0]
sha256(8ac79847:.pdd/sync-ownership.json) = 558910b5…   # = _BYTES[1]
sha256(main:.pdd/sync-ownership.json)     = 558910b5…   # also = _BYTES[1]
```

The moment anyone edits `.pdd/sync-ownership.json`, `_BYTES[1]` must be re-pinned to the new digest (the guard test added in #2329 correctly demands this). But re-pinning moves `_BYTES[1]` away from `558910b5…`, so `test_sync_rollout_repair_executes_the_actual_protected_transition` — which replays the frozen `8ac79847` head — starts failing. One constant cannot hold two values.

## Why this hasn't been seen before

`8ac79847` was frozen into the test suite **after** #2329, the last commit to touch `.pdd/sync-ownership.json`. For #2329, the live head and the frozen ref were still the same bytes, so the simple in-place re-pin worked and the coupling stayed invisible. The next ownership edit is the first to hit it.

Every commit touching `.pdd/sync-ownership.json` since the constant was introduced has also had to touch `manifest.py` (`c93332e9b` #2319, `1d024cb73` #2320, `c03a07539` #2329) — so this is on the path of every future ownership change, not an edge case.

## Reproduce

On any branch, add one row to `.pdd/sync-ownership.json`, then:

```bash
pytest tests/test_sync_core_pdd_rollout_policy.py -k sync_rollout_repair -q
```

- Without re-pinning `_BYTES[1]`: `ownership_pin_tracks_the_actual_policy_file` fails (clear message, working as designed), plus `metadata_bridge_stays_ordinary` and a cascade of unrelated-looking `.pdd/meta/*: tracked path has no ownership rule` assertions.
- After re-pinning `_BYTES[1]` as instructed: the first two pass, and `executes_the_actual_protected_transition` now fails instead.

## Suggested fix

**Immediate** — accept both pairs, matching the existing `in {…}` idiom used in `verification.py` and in these tests' own assertions:

```python
_SYNC_ROLLOUT_REPAIR_OWNERSHIP_BYTES = (BASE, CURRENT_HEAD)           # re-pinned per edit
_SYNC_ROLLOUT_REPAIR_PROTECTED_OWNERSHIP_BYTES = (BASE, FROZEN_HEAD)  # immutable replay ref

# in _sync_rollout_repair_ownership_rules:
) not in (
    _SYNC_ROLLOUT_REPAIR_OWNERSHIP_BYTES,
    _SYNC_ROLLOUT_REPAIR_PROTECTED_OWNERSHIP_BYTES,
):
```

This preserves authority rather than widening it: the frozen pair is exactly what `main` authorizes today, both bind the same reviewed base, and each still has to clear the unchanged-artifact and ordinary-row checks. `_BYTES` stays an indexable tuple, so the guard test's `_BYTES[1]` assertion is unaffected.

**Durable** — the frozen ref is immutable, so its digest does not need to be pinned by hand at all; deriving it (`sha256(read_git_blob(root, FROZEN_HEAD, ownership_path))`) is equivalent to pinning it and cannot go stale on the next edit. That removes the constant-per-freeze growth this otherwise implies.

**Either way** — consider extending the #2329 guard test to also assert the frozen pair still resolves. Today it covers only the live digest, so the next time a ref is frozen this trap silently re-arms and the next author gets the `.pdd/meta/*` cascade that guard test exists to prevent.

## Context

Surfaced while fixing #2378, which adds `tests/test_zsh_completion.py` and therefore has to register it in `.pdd/sync-ownership.json`.

agarwal-ishaan:
## Summary

- Authorize the immutable sync-rollout replay ownership digest separately from the mutable live digest.
- Preserve the historical protected transition after a live policy re-pin.
- Guard both digest pins and add a regression test that simulates a live re-pin.

Fixes #2383

## Validation

- pytest tests/test_sync_core_pdd_rollout_policy.py -k sync_rollout_repair -q
- python -m compileall -q pdd/sync_core/manifest.py tests/test_sync_core_pdd_rollout_policy.py
- git diff --check

### DIFF
--- a/.pdd/expected-managed.json
+++ b/.pdd/expected-managed.json
@@ -1771,6 +1771,10 @@
     },
     {
       "language_id": "python",
+      "prompt_path": "pdd/prompts/sync_core/manifest_python.prompt"
+    },
+    {
+      "language_id": "python",
       "prompt_path": "pdd/prompts/sync_determine_operation_python.prompt"
     },
     {
--- a/.pdd/verification-profiles.json
+++ b/.pdd/verification-profiles.json
@@ -9810,6 +9810,28 @@
       ]
     },
     {
+      "prompt_path": "pdd/prompts/sync_core/manifest_python.prompt",
+      "language_id": "python",
+      "required_requirement_ids": [
+        "CONTRACT-SHA256:c3f485c2ead115c215b9afeb900c6ba099b75749a73ee4f15083ef06ab77b8fe"
+      ],
+      "obligations": [
+        {
+          "obligation_id": "threshold-human-attestation",
+          "kind": "human-attestation",
+          "validator_id": "threshold-ed25519",
+          "validator_config_digest": "threshold-ed25519-v1",
+          "requirement_ids": [
+            "CONTRACT-SHA256:c3f485c2ead115c215b9afeb900c6ba099b75749a73ee4f15083ef06ab77b8fe"
+          ],
+          "artifact_paths": [
+            "pdd/prompts/sync_core/manifest_python.prompt"
+          ],
+          "required": true
+        }
+      ]
+    },
+    {
       "prompt_path": "pdd/prompts/sync_determine_operation_python.prompt",
       "language_id": "python",
       "required_requirement_ids": [
--- a/architecture.json
+++ b/architecture.json
@@ -12123,5 +12123,39 @@
       "x": 11600,
       "y": 2200
     }
+  },
+  {
+    "reason": "Builds the deterministic protected-base/head manifest that makes ownership and tracked-artifact classification reviewable.",
+    "description": "Reads Git trees and protected policy inputs to assemble a stable inventory of managed units, human-owned artifacts, policy controls, and invalid paths; it also enforces replay and fixed-point constraints for synchronization transitions.",
+    "dependencies": [],
+    "priority": 250,
+    "filename": "sync_core/manifest_python.prompt",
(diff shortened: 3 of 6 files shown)


agarwal-ishaan:
No longer needed, Closing this PR

PR #2388, opened 2026-08-10
fix(sync): retire stale code-generator transition

agarwal-ishaan:
## Summary

Retires the older active transition for pdd/prompts/code_generator_main_python.prompt, while retaining the newer Phase-A authorization from #2382.

This removes the duplicate active rule that blocks verification loading.

## Validation

Focused transition suite: 12 passed, 2 failed.

The remaining failures expose additional stale transition bindings for other prompts, including agentic_checkup, agentic_common, and detect_change, after the duplicate rule is retired.

This is intentionally a draft while the broader stale-binding repair is designed.

### DIFF
--- a/.pdd/verification-profile-rotations.json
+++ b/.pdd/verification-profile-rotations.json
@@ -672,6 +672,30 @@
         "base_prompt_sha256": "09e5140c01bbf8136f4c487c873c816f8e75db75412c11794a8b7ea47259cf3c",
         "head_prompt_sha256": "10129606f47d4301052490b7767acc08d8fc713e48bcb2b867efadf2063f8d1e"
       }
+    },
+    {
+      "obsolete": {
+        "prompt_path": "pdd/prompts/code_generator_main_python.prompt",
+        "language_id": "python",
+        "from_requirement_id": "CONTRACT-SHA256:8d0a51822aaa2c64e1b71579d0cb6e41460b90734441e0b6ec3ec146402495fe",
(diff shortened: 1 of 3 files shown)


gltanaka (changes requested):
Requesting changes because the current patch shifts the transition failures instead of resolving them safely.

The underlying stale-transition state predates this PR: main contains nine aged-out active bindings, and #2382 left two active code_generator_main_python rules. However, commit 9064135 introduces a new regression. _matches_current_stationary_requirement_state accepts any self-consistent stationary prompt/profile pair without preserving the exact historical policy binding. The two current CI failures demonstrate this: both PR #2316 negative controls append one byte to the rotation policy and are incorrectly accepted with no invalid reasons.

There is also an uncovered base-to-head blocker. Running build_unit_manifest plus load_verification_profiles with base f25a0139 and head 90641355 fails with:

    VerificationProfileError: protected active requirement transition rules are ambiguous

The proposed retirement also points to the #2382 replacement row that is already present in protected_rows, while _validate_candidate_retirements explicitly rejects that form.

Please:

- remove or narrowly constrain the generic stationary shortcut so exact policy-history checks remain fail-closed;
- add an exact-byte-bound recovery path for the already-protected duplicate state;
- address the nine sibling stale bindings through explicit audited cleanup semantics rather than general self-consistency;
- add a regression exercising the actual PR base-to-head loader boundary; and
- rerun tests/test_sync_core_verification_profiles.py and tests/test_sync_core_pdd_rollout_policy.py.

Current CI evidence: 2 failed, 16,314 passed. The failures are deterministic and PR-caused, although the stale data they expose is pre-existing.

PR #2399, opened 2026-08-12
fix: bootstrap metadata for nested prompts

### LINKED ISSUE #2387: sync-architecture cannot bootstrap metadata for a new nested prompt
## Summary

A new nested prompt cannot be registered through the normal PDD metadata workflow without manually editing both the prompt and `architecture.json`.

`pdd update <code> --sync-metadata` creates the prompt and fingerprint, but skips tags and architecture registration when there is no existing architecture entry. Then `pdd sync-architecture <prompt>` cannot create that entry because the prompt has no PDD metadata and no matching entry to update.

## Reproduction

1. Add a source file with no prompt or architecture entry, for example `pdd/sync_core/manifest.py`.
2. Create its prompt:

    pdd update pdd/sync_core/manifest.py --output pdd/prompts/sync_core/manifest_python.prompt --sync-metadata

3. Observe metadata-sync output:

    tags: skipped (prompt has no PDD tags and no architecture entry to seed from; LLM-first tag generation pending (#870))
    architecture: skipped (No architecture entry found for: sync_core/manifest_python.prompt)

4. Try to synchronize architecture metadata:

    pdd sync-architecture sync_core/manifest_python.prompt

5. It reports `No architecture entry found for: sync_core/manifest_python.prompt` instead of creating one.

## Expected behavior

PDD should offer a supported bootstrap path for a new prompt, such as generating the minimal `<pdd-reason>` and `<pdd-interface>` tags from the source/prompt and creating the corresponding `architecture.json` module entry. Nested prompt paths should be accepted relative to the configured prompt root.

## Actual workaround

Manually add matching PDD tags to the prompt and a hand-written `architecture.json` entry, then run `pdd sync-architecture` to refresh it. This defeats the purpose of the update/sync metadata workflow for newly prompt-managed modules.

## Context

Observed while converting `pdd/sync_core/manifest.py` to a prompt-managed unit for issue #2383.

agarwal-ishaan:
## Summary
- bootstrap deterministic PDD tags for a new prompt during `pdd update --sync-metadata`
- register only that prompt-relative path before synchronizing `architecture.json`
- cover nested registration and dry-run behavior

Fixes #2387

## Validation
- `pytest -q tests/test_metadata_sync.py`


gltanaka (changes requested):
Changes requested — Tier A, full-diff review at `56b50db63`.

1. **P1 — the issue path is registered with the wrong source filepath** (`pdd/metadata_sync.py:420`). The new call to `register_untracked_prompts` ignores the authoritative `code_path`. For issue #2387 exact layout, `pdd/prompts/sync_core/manifest_python.prompt` is registered as `filepath=sync_core/manifest.py`, while the real source is `pdd/sync_core/manifest.py`; metadata sync still reports `ok=True`. The new test asserts only that the filename exists, so it misses this broken architecture edge. Pass or otherwise preserve the repo-relative `code_path`, and add an exact `pdd/prompts/sync_core` regression/E2E assertion for `filepath`.

2. **P1 — bootstrap persists an invalid placeholder interface** (`pdd/metadata_sync.py:168-173`). The injected `{"type":"module"}` fails the repository validator with `Missing module key for type=module`; it also declares none of the source public surface. Because `has_pdd_tags` then becomes true, later syncs preserve this fabricated metadata indefinitely while reporting success. Derive and validate the interface from the prompt/source before persisting it, or at minimum do not write a structurally invalid interface or claim successful bootstrap.

3. **P1 — the protected verification profile is broken at PR HEAD** (`pdd/sync_core/verification.py:175-183` and `.pdd/verification-profile-rotations.json`). The PR pins the metadata transition to profile digest `3af7bc...`, but the merged HEAD profile bytes hash to `a5d973...`. The four focused rollout-policy tests fail, matching the GitHub unit-test job; the PR also now conflicts with current `main` in the rotations file. Rebase current `main`, recompute the composed rotation/retirement authority and byte pins for the final artifacts, then rerun the rollout suite.

Validation:
- `pytest -q tests/test_metadata_sync.py` — 49 passed.
- `pytest -q tests/test_architecture_sync.py -k "register_untracked_prompts or infer_filepath"` — 11 passed, but direct issue-path inference returns `sync_core/manifest.py` instead of `pdd/sync_core/manifest.py`.
- The four affected `tests/test_sync_core_pdd_rollout_policy.py` cases — 4 failed, identical to CI.
- `validate_interface_structure({"type":"module"})` — invalid.
- Prior inline findings on basename borrowing, malformed architecture, and tagged-prompt dry-run reporting are fixed at HEAD.

The problem is real and this bootstrap path is needed, but contributor E2E evidence is not sufficient: the focused suite does not exercise the repository layout from #2387, validate the generated architecture entry, validate the injected interface, or cover the protected profile integration. The prompt and profile artifacts are present; the existing `architecture.json` entry for `metadata_sync` still matches its unchanged public API, so that file itself does not need an interface update.

gltanaka (changes requested):
Follow-up review at `561f7fab7` after the force-rebase/remediation: changes are still required.

1. **P1 — protected profile integration remains invalid.** `.pdd/verification-profile-rotations.json` still binds the metadata transition to stale profile hashes (`6e5891… → 3af7bc…`) rather than the current base/head bytes (`85d010… → a50598…`), and `pdd/sync_core/verification.py` still carries incompatible pins. Final-head CI fails five protected rollout-policy tests, including retirement-byte validation and current-profile reconciliation. Rebuild the composed transition/retirement authority against current `main` and the final artifacts, then rerun the rollout suite.

2. **P1 — the runtime fixes are missing from the prompt/architecture source of truth.** `pdd/prompts/metadata_sync_python.prompt` still specifies the old invalid `{"type":"module"}` bootstrap and does not require the authoritative code filepath. `pdd/prompts/architecture_sync_python.prompt` and the `architecture_sync_python.prompt` entry in `architecture.json` still expose `register_untracked_prompts(..., only_files=None)` without the new `known_filepaths` parameter. Regeneration can therefore remove this remediation and recreate both prior bugs. Update both prompt contracts and `architecture.json`, then rotate the resulting protected profiles.

Disposition of the earlier findings:
- Wrong nested filepath: runtime fixed; exact issue-layout regression passes.
- Invalid interface structure: runtime fixed; validator-backed test passes.
- Protected profile integration: unresolved.

Validation: `tests/test_metadata_sync.py` 50 passed; relevant architecture registration/inference tests 11 passed; final-head GitHub Unit Tests 5 failed / 16,441 passed; `git diff --check` passed. The PR is conflict-free now, but it is not mergeable while required CI and generated-source synchronization remain broken.

gltanaka (changes requested):
Third review at `1aa266efa`: changes are still required.

1. **P1 — duplicate active metadata transitions invalidate the protected policy.** `.pdd/verification-profile-rotations.json` retains the earlier `metadata_sync` transition `6ab860… → 4b5dba…` and adds `6ab860… → 99b422…` for the same prompt/language/source identity. The verifier correctly rejects those competing successors as ambiguous. Final-head unit CI fails 16 tests, predominantly with `protected active requirement transition rules are ambiguous`. Remove or correctly supersede the stale branch-local row and rebuild exact bindings against current `main`.

2. **P1 — the protected rollout assertion does not recognize the new profile state.** `test_sync_rollout_repair_executes_the_actual_protected_transition` still permits profile heads only through the prior dependency-fix state, while final HEAD is `aca4c6…`; it fails independently of the ambiguity. Advance the protected-history assertion for the reviewed final composed state.

Current-main compatibility is also blocking: GitHub reports `CONFLICTING` / `DIRTY`, with conflicts in `.pdd/verification-profile-rotations.json` and `pdd/sync_core/verification.py`. Rebase and compose this transition with the current main profile state, then rerun the complete rollout/profile suite.

Prior source-of-truth findings are resolved: both prompts and `architecture.json` now include the valid interface and `known_filepaths`; the nested filepath and interface runtime regressions remain fixed. But this head must not merge while required CI is red and protected transition authority is ambiguous.

PR #2407, opened 2026-08-15
chore(sync): preauthorize story prompt metadata paths

gltanaka:
## Summary

Policy-only protected Phase A for #2380 / #2379.

- replaces the already-consumed `c63d… -> 1c467…` story-prompt authorization with the dormant `1c467… -> 5b135…` successor
- leaves the verification profile, prompt, generated code, and every unrelated managed prompt byte-identical
- relies on the repository- and digest-bound prerequisite merged in #2405

This PR does not consume the authorization. PR #2380 remains the later Phase B.

## Validation

- `9 passed` affected exact-install/stationary/tamper/conformance/denominator tests
- independent full verification-profile suite: `81 passed`
- independent focused governance checks: `10 passed`
- fresh high-rigor review: no P1/P2 findings
- `git diff --check origin/main..HEAD`: clean
- exact rotation policy digest: `470a4e… -> b8b1e1…`
- profile remains `85d010…`; prompt remains `1c4670…`

Prompt and `architecture.json` changes are intentionally absent: this is an authority-only Phase A, and changing either protected artifact here would invalidate dormancy.

PR #2408, opened 2026-08-15
test(sync): cache immutable story rollout blobs

gltanaka:
## Summary

Fixes the test-runtime regression introduced by the Git-backed story Phase-A fixture in #2406.

The unit-test job has a hard 60-minute timeout. Re-reading ~475 immutable historical prompt blobs for each Phase-A case added about four minutes; PR #2407 reached exactly 60:00 and was canceled despite all faster checks passing.

This test-only change caches reads from the pinned immutable commit `60588697e…`. Candidate policy/profile bytes and candidate prompt mutations remain explicit per-test overrides, so the cache cannot mask the negative cases.

## Validation

- six Phase-A exact/tamper/drift tests: `6 passed in 15.89s` (previously ~256.74s)
- independent rerun: `6 passed in 15.28s`
- targeted pylint: `10.00/10`
- `git diff --check`: clean
- fresh Tier B review: no P1/P2 findings

Coverage retained: exact consumed replacement, public-loader success, stationary state, policy/row/profile tampering, target prompt drift, and unrelated managed-prompt drift.

No production, prompt, profile, rotation-policy, or architecture bytes change.

### DIFF
--- a/tests/test_sync_core_pdd_rollout_policy.py
+++ b/tests/test_sync_core_pdd_rollout_policy.py
@@ -11,6 +11,7 @@ import subprocess
 import sys
 import tarfile
 from dataclasses import replace
+from functools import cache
 from pathlib import Path, PurePosixPath
 from types import SimpleNamespace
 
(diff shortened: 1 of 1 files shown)


gltanaka:
Independent review found no P1/P2 issues. The six exact/tamper/drift tests retain public-loader coverage and now pass in 15.28–15.89s versus ~256.74s before. Full CI is green, including the complete Unit Tests job under the 60-minute limit; all CLI, story, package, Docker, analysis, and security lanes passed. This is test-only and changes no production, prompt, profile, rotation-policy, or architecture bytes.

